Issue #10, opened 2019-11-09
coroutines are required for some of the example


dietmarkuehl:
I have tried to build the library with gcc which, unfortunately, doesn't support coroutines, yet. However, the build unconditionally builds the examples and most of them depend on headers including `<experimental/coroutines>`. It would be helpful to have a `cmake` option to avoid building the corresponding examples.

lewissbaker:
With #12 it should now auto-detect whether coroutine support is present and avoid including/defining the parts that depend on coroutines.

You can override this auto-detection rules by adding `-DUNIFEX_HAVE_COROUTINES=0` to the `CMAKE_CXX_FLAGS` variable on the command-line.

lewissbaker:
This has now been merged.
Define `-DUNIFEX_NO_COROUTINES=1` to disable coroutines (it should auto-detect by default but you can override by defining this macro).

ericniebler:
@dietmarkuehl I've beefed up the cmake scripts to do a better job of auto-detecting what features are available, including both coroutines and `<memory_resource>`. Give it a spin.

dietmarkuehl:
The code certainly builds with  gcc. I haven't managed to build a recent LLVM but I'd believe it is working there, too. There were also some other changes you made such that everything is building with gcc!

Thank you very much!

PR #12, opened 2019-11-10
Allow compilation without coroutine support

lewissbaker:
Conditionally disable facilities that depend on coroutine support when `UNIFEX_HAVE_COROUTINES` is false.

Also try to give a meaningful error message if you try to include coroutine-specific unifex headers without coroutine support enabled.

This should address #10.

### DIFF
--- a/examples/async_trace_test.cpp
+++ b/examples/async_trace_test.cpp
@@ -26,7 +26,7 @@
 #include <unifex/typed_via.hpp>
 #include <unifex/when_all.hpp>
 
-#if UNIFEX_HAVE_COROUTINES
+#if !UNIFEX_NO_COROUTINES
 #include <unifex/awaitable_sender.hpp>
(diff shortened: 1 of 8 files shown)


examples/async_trace_test.cpp
```diff
@@ -14,17 +14,24 @@
...
-#include <unifex/task.hpp>
+
+#include <unifex/config.hpp>
+#include <unifex/just.hpp>
 #include <unifex/timed_single_thread_context.hpp>
 #include <unifex/transform.hpp>
 #include <unifex/typed_via.hpp>
 #include <unifex/when_all.hpp>
 
+#if UNIFEX_HAVE_COROUTINES
```
ericniebler: My preference would be to flip the sense of the macro here. `UNIFEX_NO_COROUTINES` has the nice property that in the long run, when all compilers support coroutines, we don't have to keep defining a macro to keep things working.

Issue #21, opened 2019-11-13
Add ability to reschedule onto the current scheduler


lewissbaker:
Add overloads of the `schedule()`, `schedule_after()` algorithms that don't take a scheduler and that use whatever the current scheduler from the receiver is (ie. using `get_scheduler()` - see #20).

This would simplify writing some kinds of code, like a `timeout()` algorithm.
```c++
template<typename Op, typename Duration>
auto timeout(Op op, Duration duration) {
  return take_until(std::move(op), schedule_after(duration));
}
```

Whereas currently we'd need to do something like:
```
template<typename Op, typename Scheduler, typename Duration>
auto timeout(Op op, Scheduler sched, Duration duration) {
  return take_until(std::move(op), schedule_after(sched, duration));
}
```

Q. Should `schedule()` be named `reschedule()`?

Issue #20, opened 2019-11-13
Add a get_scheduler() CPO for querying the current scheduler from the receiver


lewissbaker:
This would allow algorithms to be written that can be written in such a way that they just use the ambient current scheduler rather than having to explicitly pass the scheduler into the algorithm.

This should be done in conjunction with having the `on()` algorithm inject the current scheduler into the receiver passed to the successor. This part depends on #15.

PR #27, opened 2019-11-13
Add support for passing the associated scheduler into child operations

lewissbaker:
Adds the following:
* `unifex::cpo::get_scheduler(receiver)` - Get the current scheduler.
* `unifex::cpo::schedule()` - Returns a sender that will schedule onto the current scheduler.

Also contains some fixes:
* Adds missing docs for `get_stop_token()`, `get_allocator()`, `with_query_value()` and `with_allocator()`
* Fixes the `reduce_stream` cleanup receivers to forward through query CPOs
* Const-qualify the `schedule()`, `schedule_at()` and `schedule_after()` methods on scheduler implementations.

Addresses #21, #20.

### DIFF
--- a/doc/api_reference.md
+++ b/doc/api_reference.md
@@ -1,7 +1,9 @@
 # Index
 
 * Receiver Queries
+  * `get_stop_token()`
   * `get_scheduler()`
+  * `get_allocator()`
 * Sender Algorithms
   * `transform()`
   * `via()`
@@ -10,6 +12,8 @@
   * `let()`
   * `sync_wait()`
   * `when_all()`
+  * `with_query_value()`
+  * `with_allocator()`
 * Sender Types
   * `async_trace_sender`
 * Sender Queries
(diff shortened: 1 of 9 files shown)


LeeHowes:
Would it be worth adding a with_scheduler algorithm to make use of this functionality?

Such that:
with_scheduler(S<T>) -> S<T, Sch>

Thereby giving, for example:
auto v = sync_wait(just(3) | with_scheduler() | transform([](auto v, auto s) { return v; }) | on(s));

ericniebler:
> ```
> auto v = sync_wait(just(3) | with_scheduler() | transform([](auto v, auto s) { return v; }) | on(s));
> ```

I don't know how to read this. What is `s` in `on(s)` ? It's not in the `transform`.

LeeHowes:
Oh, the use of s there is confusing.
auto sch = make_a_scheduler();
auto v = sync_wait(just(3) | with_scheduler() | transform([](auto v, auto s) { return v; }) | on(sch));

Does that help? on is providing a receiver that can return sch, transform can see that, with_scheduler can see it too if transform propagates it. The with_scheduler algorithm is zipping up the scheduler from get_scheruler with the current propagating value.

lewissbaker:
I can certainly imagine having some kind of sender that will immediately complete with the result of some query on the receiver.  This could then be passed to `when_all()` to obtain both.
```
on(
  transform(
    when_all(just(3), query_context(get_scheduler)),
    [](int x, auto scheduler) { ... }),
  someScheduler);
```

But, yes, you could alternatively define some special operator that just appended an extra argument to the set_value() call.
```
on(
  transform(
    query_context(just(3, 5), get_scheduler),
    [](int x, int y, auto scheduler) { ... }),
  someScheduler);
```

Then the former can be defined as `query_context(queryCpo) -> query_context(just(), queryCpo)`.

PR #29, opened 2019-11-13
fixed private access from hidden friends

dietmarkuehl:
The code in `libunifex` uses in multiple places `friend` functions as customization points. In several of these places the `friend` function is declared in a nested class and accesses a `private` member of an outer class.  Using this approach fails to compile with [gcc](http://gcc.gnu.org/), Intel's compiler, and [EDG](http://edg.com/): A `friend` of a `friend` isn't a `friend`. It seems [clang](http://clang.llvm.org/) does not flag an error for the code.

Here is a minimal example showing the problem ([live](https://godbolt.org/z/EWCrCJ)):

    class outer {
        class inner {
            friend int access(inner i) { return i.o->value; }
            outer* o;
        public:
            inner(outer* o): o(o){}
        };
        int value{};

    public:
        inner get() { return inner{this}; }
    };

    int main() { return access(outer().get()); }

A local fix to the problem is to have the `friend` of the nested class delegate to a [`private`] member of the nested class ([live](https://godbolt.org/z/3j-nq_)]:

    class outer {
        class inner {
            int get_value() const { return o->value; }
            friend int access(inner i) { return i.get_value(); }
            outer* o;
        public:
            inner(outer* o): o(o){}
        };
        int value{};

    public:
        inner get() { return inner{this}; }
    };

    int main() {
        return access(outer().get());
    }

This pull request fixes these problems. It may not be the ideal change as the auxilary methods are probably not named quite the right way and placed a bit random. I'd be happy to adjust these issues based on corresponding comments.

### DIFF
--- a/include/unifex/thread_unsafe_event_loop.hpp
+++ b/include/unifex/thread_unsafe_event_loop.hpp
@@ -274,13 +274,13 @@ class thread_unsafe_event_loop {
       friend const StopToken& tag_invoke(
           tag_t<get_stop_token>,
           const receiver& r) noexcept {
-        return r.get_stopToken();
+        return r.get_stop_token();
       }
 
      private:
       friend sync_wait_promise;
 
-      StopToken& get_stopToken() const {
+      StopToken& get_stop_token() const {
         return promise_.stopToken_;
       }
 
--- a/include/unifex/thread_unsafe_event_loop.hpp
+++ b/include/unifex/thread_unsafe_event_loop.hpp
@@ -274,11 +274,15 @@ class thread_unsafe_event_loop {
       friend const StopToken& tag_invoke(
           tag_t<get_stop_token>,
           const receiver& r) noexcept {
-        return r.promise_.stopToken_;
+        return r.get_stopToken();
       }
 
      private:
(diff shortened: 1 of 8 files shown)


include/unifex/thread_unsafe_event_loop.hpp
```diff
@@ -274,12 +274,16 @@ class thread_unsafe_event_loop {
...
           tag_t<get_stop_token>,
           const receiver& r) noexcept {
-        return r.promise_.stopToken_;
+        return r.get_stopToken();
       }
 
      private:
       friend sync_wait_promise;
 
+      StopToken& get_stopToken() const {
```
lewissbaker: Similarly, can this please be renamed to `get_stop_token()`?

ericniebler:
I've taken the liberty of doing the renames.

dietmarkuehl:
@ericniebler: thank you! I can’t make the changes from work, i.e., it would still be sone time before I could do them.

There are a few other issues which prevent building the code with gcc but I haven’t tracked them down to their root cause, yet: my objective is to make the library usable with other compilers.

Issue #17, opened 2019-11-13
Add static_thread_pool execution context


lewissbaker:
Add a simple thread-pool scheduler to allow building examples that incorporate parallelism.

PR #30, opened 2019-11-14
Add a simple static_thread_pool type.

lewissbaker:
Adds a simple static_thread_pool class.

Its implementation makes use of relatively heavy-weight condition_variable/mutex synchronisation to synchronise access to the single queue that is shared across all threads. More sophisticated data-structures are possible (mostly-lock-free, work-stealing) with various fairness/performance tradeoffs.

Addresses #17.

### DIFF
--- a/source/static_thread_pool.cpp
+++ b/source/static_thread_pool.cpp
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ #include <unifex/static_thread_pool.hpp>
+
+namespace unifex
+{
+
+static_thread_pool::static_thread_pool()
+: static_thread_pool(std::thread::hardware_concurrency())
+{}
+
+static_thread_pool::static_thread_pool(std::uint32_t threadCount) {
+    assert(threadCount > 0);
+
+    threads_.reserve(threadCount);
+    try {
+        threads_.emplace_back([this] { this->run(); });
+    } catch (...) {
+        close();
+    }
+}
+
+static_thread_pool::~static_thread_pool() {
+    close();
+}
+
+void static_thread_pool::run() noexcept {
+    std::unique_lock lk{mut_};
+    while (true) {
+        while (!queue_.empty()) {
+            auto* task = queue_.pop_front();
+            lk.unlock();
+            task->execute(task);
+            lk.lock();
+        }
+
+        if (stop_) {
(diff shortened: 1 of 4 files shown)


source/static_thread_pool.cpp
```diff
@@ -0,0 +1,75 @@
...
+{}
+
+static_thread_pool::static_thread_pool(std::uint32_t threadCount) {
+    assert(threadCount > 0);
+
+    threads_.reserve(threadCount);
+    try {
+        threads_.emplace_back([this] { this->run(); });
+    } catch (...) {
+        close();
```
lewissbaker: Need to rethrow the exception here.

ericniebler:
After hearing Sean Parent talk about how egregiously slow it is for a thread pool to have a single queue, I'm inclined to not put this one in. Better to have N queues for N threads. When enqueuing, use a round-robin approach, spilling over to the next queue when one is full, and only growing when all are full. On dequeue, each thread takes from its queue and steals from other threads only when its queue is empty. Only block after trying to pull from each queue. This, Sean says, has *MUCH* better performance.

source/static_thread_pool.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+        if (stop_) {
+            return;
+        }
+
+        cv_.wait(lk);
+    }
+}
+
+void static_thread_pool::close() noexcept {
```
kirkshoop: I'd rather use `stop_token` and `sender join()` for this than `close()` and `stop_`

source/static_thread_pool.cpp
```diff
@@ -0,0 +1,75 @@
...
+    close();
+}
+
+void static_thread_pool::run() noexcept {
+    std::unique_lock lk{mut_};
+    while (true) {
+        while (!queue_.empty()) {
+            auto* task = queue_.pop_front();
+            lk.unlock();
+            task->execute(task);
```
kirkshoop: exceptions? I would love to have a scoped reverse lock that would release the lock in the constructor and take it in the destructor. much safer to use than this manual locking

lewissbaker: The execute() function is noexcept so should be safe to manually unlock/lock here.
I agree it would be nice to have some RAII unlocker, though.

source/static_thread_pool.cpp
lewissbaker: Ok, so to add a `request_stop()` method to the thread-pool?

Adding an asynchronous `.join()` operation on the thread-pool could prove difficult given the current APIs for joining threads are all blocking in nature. I guess we could ask all threads to shut-down and when the last thread is about to exit it's `run()` function it invokes the receiver passed to the sender.

The destructor of the thread-pool would still need to call `.join()` on the threads, however most of the threads should already have terminated by this point.

lewissbaker: I've added a `request_stop()` method now to at least partially support the asynchronous shutdown use-case.

source/static_thread_pool.cpp
```diff
@@ -0,0 +1,151 @@
...
+      }
+
+      task->execute(task);
+    }
+  }
+
+  void static_thread_pool::join() noexcept {
+    for (auto& t : threads_) {
+      t.join();
+    }
```
ericniebler: If you used `jthread` this loop would be unnecessary. 🤣

lewissbaker: Indeed :)
I've heard that libstdc++ will have jthread/stop_token soon.
I haven't heard about libc++/msvc yet.

PR #34, opened 2019-11-15
CMake configuration of coroutines and pmr/memory_resource

ericniebler:
Update cmake to do the following:

* Propagate the `cxx_std_20` feature to any cmake target that depends on unifex
* Test for coroutines support and set compiler flags appropriately
* Test for PMR support and set compiler flags appropriately
* Adds a `UNIFEX_STDLIB_PATH` cmake cache entry for specifying where to find -lc++experimental


source/CMakeLists.txt
```diff
@@ -4,18 +4,18 @@
...
     manual_event_loop.cpp
     trampoline_scheduler.cpp
     thread_unsafe_event_loop.cpp
     timed_single_thread_context.cpp)
+target_link_libraries(unifex PUBLIC unifex-interface)
 
 if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
   target_sources(unifex
-    PRIVATE
+    INTERFACE
```
cor3ntin: I can't think of a reason for this change ?

ericniebler: Probably a holdover from a time when I naively thought simply making that target INTERFACE would cause features to propagate to dependencies. I can revert that.

PR #35, opened 2019-11-16
Allow schedule_after(d) to use implicit scheduler.

lewissbaker:
You can now specify the `schedule_after(d)` CPO without needing to explicitly specify a scheduler.

This will pick up the implicit current scheduler from the receiver that is eventually attached to the `schedule_after()` operation.

### DIFF
--- a/include/unifex/scheduler_concepts.hpp
+++ b/include/unifex/scheduler_concepts.hpp
@@ -96,6 +96,43 @@ inline constexpr struct schedule_after_cpo {
       -> decltype(tag_invoke(*this, (TimeScheduler &&) s, (Duration &&) d)) {
     return tag_invoke(*this, (TimeScheduler &&) s, (Duration &&) d);
   }
+
+  template<typename Duration>
+  class schedule_after_sender {
+  public:
+    template<template<typename...> class Variant, template<typename...> class Tuple>
+    using value_types = Variant<Tuple<>>;
+
+    template<template<typename...> class Variant>
+    using error_types = Variant<std::exception_ptr>;
+
+    explicit schedule_after_sender(Duration d)
+    : duration_(d)
+    {}
+
+  private:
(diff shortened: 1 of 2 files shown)


lewissbaker:
This should address the remaining piece of #21

include/unifex/scheduler_concepts.hpp
```diff
@@ -97,6 +97,43 @@ inline constexpr struct schedule_after_cpo {
...
+
+    template<typename Receiver>
+    friend auto tag_invoke(tag_t<connect>, const schedule_after_sender& s, Receiver&& r)
+    -> std::invoke_result_t<
+          decltype(connect),
+          std::invoke_result_t<
+              schedule_after_cpo,
+              std::invoke_result_t<decltype(get_scheduler), const Receiver &>,
+              const Duration&>,
+          Receiver> {
```
ericniebler: Style nit: this return type is hella long. Consider making an alias template for it.

include/unifex/scheduler_concepts.hpp
```diff
@@ -133,5 +170,12 @@ inline constexpr struct now_cpo {
   }
 } now;
 
+inline constexpr struct bulk_schedule_cpo {
+  template<typename BulkScheduler, typename Range>
+  auto operator()(BulkScheduler&& scheduler, Range&& range) const {
+
```
ericniebler: Uh, this line intentionally left blank? :-D I'm guessing this got included by accident.

lewissbaker: This whole `bulk_schedule` CPO shouldn't be in this PR.
Will remove.

include/unifex/scheduler_concepts.hpp
lewissbaker: This should be improved now.

PR #36, opened 2019-11-16
Add new allocate() algorithm.

lewissbaker:
This algorithm adapts a sender, causing it to heap-allocate the operation-state of the wrapped sender instead of embedding it inside the operation-state of the caller.

This can be useful in cases where a child operation has a very large operation state that you don't necessarily want to be inlined into the parent's operation-state.

Note that this does not type-erase the operation, only heap-allocate it.


include/unifex/heap_allocate.hpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  ~heap_allocated_operation() {
+    op_->~Operation();
+    std::allocator_traits<allocator_t>::deallocate(allocator_, op_, 1);
+  }
+
+  void start() noexcept {
+      cpo::start(*op_);
+  }
```
ericniebler: Should this be `friend auto tag_invoke(tag_t<start>, heap_allocated_operation...` ?

include/unifex/heap_allocate.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+  template <typename Receiver>
+  friend auto tag_invoke(
+      tag_t<cpo::connect>, const heap_allocate_sender& s, Receiver&& r)
+      -> heap_allocated_operation<operation_t<const Sender&, Receiver>,
+                                  std::remove_cvref_t<get_allocator_t<Receiver>>> {
+    return heap_allocated_operation<operation_t<const Sender&, Receiver>,
+                                  std::remove_cvref_t<get_allocator_t<Receiver>>>{
+        std::as_const(s.sender_), (Receiver&&) r};
```
ericniebler: `as_const` is not needed here, is it?

include/unifex/heap_allocate.hpp
lewissbaker: That would be the other way to implement the start operation.

At the moment, you can still either define a `tag_invoke()` overload or can define a member-function.

include/unifex/heap_allocate.hpp
lewissbaker: Yes, it's not strictly needed as long as you assume that the `Sender` template is not a non-const lvalue reference.

kirkshoop:
`heap_allocate` is not accurate when the allocator is not a heap allocator..

lewissbaker:
> heap_allocate is not accurate when the allocator is not a heap allocator..

@kirkshoop I've renamed this algorithm to just be `allocate()` now, rather than  `heap_allocate()`

PR #41, opened 2019-11-19
Remove unifex::cpo namespace

lewissbaker:
Move all of the CPOs that were previously defined in the `unifex::cpo` namespace to now be defined in the `unifex` namespace.

Since we want to eventually make most algorithms/functions CPOs it doesn't really make sense to organise them into a namespace based on whether they are CPOs or not. We shouldn't need to update call-sites that invoke algorithms when we convert an algorithm from an ordinary function to be a CPO.

Also, modify the receiver CPOs to look for `.set_value()`/`.set_error()`/`.set_done()` member functions instead of `.value()`/`.error()`/`.done()` so that we are more consistent with the design in [P0443R11](https://wg21.link/P0443R11).

### DIFF
--- a/include/unifex/adapt_stream.hpp
+++ b/include/unifex/adapt_stream.hpp
@@ -23,57 +23,61 @@
 namespace unifex {
 
 template <
-    typename StreamSender,
+    typename Stream,
     typename NextAdaptFunc,
     typename CleanupAdaptFunc>
 struct adapted_stream {
-  StreamSender innerStream_;
+  Stream innerStream_;
   NextAdaptFunc nextAdapter_;
   CleanupAdaptFunc cleanupAdapter_;
 
-  auto next() {
-    return std::invoke(nextAdapter_, cpo::next(innerStream_));
+  friend auto tag_invoke(tag_t<next>, adapted_stream& s)
+    -> std::invoke_result_t<NextAdaptFunc&, next_sender_t<Stream>> {
+    return std::invoke(s.nextAdapter_, next(s.innerStream_));
   }
 
-  auto cleanup() {
-    return std::invoke(cleanupAdapter_, cpo::cleanup(innerStream_));
+  friend auto tag_invoke(tag_t<cleanup>, adapted_stream& s)
+    -> std::invoke_result_t<CleanupAdaptFunc&, cleanup_sender_t<Stream>> {
+    return std::invoke(s.cleanupAdapter_, cleanup(s.innerStream_));
   }
 };
 
-template <typename StreamSender, typename AdaptFunc>
+template <typename Stream, typename AdaptFunc>
 struct both_adapted_stream {
-  StreamSender innerStream_;
+  Stream innerStream_;
   AdaptFunc adapter_;
 
-  auto next() {
-    return std::invoke(adapter_, cpo::next(innerStream_));
+  friend auto tag_invoke(tag_t<next>, both_adapted_stream& s)
+    -> std::invoke_result_t<AdaptFunc&, next_sender_t<Stream>> {
+    return std::invoke(s.adapter_, next(s.innerStream_));
   }
 
-  auto cleanup() {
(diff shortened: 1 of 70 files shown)


include/unifex/stop_immediately.hpp
```diff
@@ -86,7 +86,7 @@ struct stop_immediately_stream {
           stream_.stopSource_.request_stop();
           auto receiver = std::exchange(stream_.nextReceiver_, nullptr);
           assert(receiver != nullptr);
-          std::move(*receiver).done();
+          std::move(*receiver).set_done();
```
kirkshoop: could use the cpo's for calls to `std::move(*receiver)`, yes?

include/unifex/type_erased_stream.hpp
```diff
@@ -131,35 +131,35 @@ struct type_erased_stream {
...
-      void value(Values&&... values) && noexcept {
+      void set_value(Values&&... values) && noexcept {
         try {
           // Take a copy of the values before destroying the next operation
           // state in case the values are references to objects stored in
           // the operation object.
           [&](Values... values) {
             stream_.next_.destruct();
-            receiver_.value((Values &&) values...);
+            receiver_.set_value((Values &&) values...);
```
kirkshoop: use cpo's?

include/unifex/stream_concepts.hpp
```diff
@@ -19,15 +19,8 @@
...
 namespace unifex {
-namespace cpo {
 
 inline constexpr struct next_cpo {
-  template <typename Stream>
-  friend auto tag_invoke(next_cpo, Stream& s) noexcept(noexcept(s.next()))
-      -> decltype(s.next()) {
-    return s.next();
-  }
-
```
kirkshoop: what is the reason for removing the member function impls for next and cleanup?

include/unifex/stop_immediately.hpp
lewissbaker: This particular receiver is a pointer to a type-erased base-class `next_receiver_base` for which we know the interface.

Also, this type, while it has the same interface as a receiver is never actually used as a receiver (it's never passed to `connect()`).

So I didn't think it necessary to use the CPOs here.

include/unifex/stream_concepts.hpp
lewissbaker: There were two reasons for this:
1. The `tag_invoke` paper talked about the potential for conflict when using ADL names or method names are used as customisation points and so I was trying to minimise the number of member function names that CPOs dispatch to.
2. I initially encountered naming conflicts when converting adapter code to call the CPO without the `cpo::` namespace qualification (the member name now shadows the unqualified CPO name) and so I started changing some of the member functions to tag_invoke() overloads to avoid this naming conflict. However, this could also have been addressed by prefixing the call with `unifex::` to disambiguate (I ended up doing this in some of the other CPOs in other commits).

lewissbaker: Ok, I've updated the CPOs to restore the behaviour of `next()` and `cleanup()` dispatching to member functions.

include/unifex/adapt_stream.hpp
```diff
@@ -23,57 +23,61 @@
...
     typename CleanupAdaptFunc>
 struct adapted_stream {
-  StreamSender innerStream_;
+  Stream innerStream_;
   NextAdaptFunc nextAdapter_;
   CleanupAdaptFunc cleanupAdapter_;
 
-  auto next() {
-    return std::invoke(nextAdapter_, cpo::next(innerStream_));
+  friend auto tag_invoke(tag_t<next>, adapted_stream& s)
```
kirkshoop: thoughts for the future - perhaps if there was a way to get clang format to always add a <CR> before the first arg in a tag_invoke decl/def?

ericniebler: To always add a what?

Issue #39, opened 2019-11-18
Adopt a unit-testing framework


lewissbaker:
The tests we have at the moment are mostly smoke-tests/examples that check that things compile and at least pass some basic runtime checks.

If we want to write more comprehensive tests then we should consider making use of a unit-test framework (eg. googletest, catch2, etc.) for writing the tests.

ericniebler:
I vote against catch2. I have bad experiences with catch2 greatly increasing compile times. (Ironic for me to be complaining about that, no?)

ericniebler:
I'm remembering that I did work to make pushmi use GTest and wrote some cmake to make it happen: https://github.com/facebookresearch/pushmi/blob/master/cmake/CMakeLists.txt.in

We should steal that.

PR #43, opened 2019-11-19
Use GoogleTest test harness

ericniebler:
* add cmake magic to download and build gtest
* port any_unique_test to use gtest
* refs #39


ericniebler:
Obviously, all the other tests need to be ported also to GTest.

ericniebler:
We should separate examples from tests.

examples/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ foreach(file-path ${test-sources})
     string( REPLACE ".cpp" "" file-path-without-ext ${file-path} )
     get_filename_component(file-name ${file-path-without-ext} NAME)
     add_executable( ${file-name} ${file-path})
-    target_link_libraries( ${file-name} PUBLIC unifex)
+    target_link_libraries(${file-name} PUBLIC unifex gtest_main)
```
lewissbaker: I assume linking in gtest_main treat the `main()` it provides as a weak symbol and still prefer the `main()` from the sources?

ericniebler: I _think_ so.

PR #44, opened 2019-11-21
Add sequence() algorithm and refactor on() to be in terms of this

lewissbaker:
- Adds a new `sequence(senders...)` algorithm that supports chaining multiple senders together in sequence.
  This is a variadic version with the semantics of the previous `on()` algorithm.
- Modified `on()` to now take a sender and a scheduler instead of two senders.
  This now also injects the scheduler as the current scheduler into the child operations by
  customising the `get_scheduler` CPO on the receiver.

### DIFF
--- a/include/unifex/sequence.hpp
+++ b/include/unifex/sequence.hpp
@@ -1,0 +1,493 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


include/unifex/sequence.hpp
```diff
@@ -0,0 +1,493 @@
...
+
+    friend blocking_kind
+    tag_invoke(tag_t<blocking>, const sequence_sender& sender) {
+      const auto predBlocking = blocking(sender.predecessor_);
+      const auto succBlocking = blocking(sender.successor_);
+      if (predBlocking == blocking_kind::never) {
+        return blocking_kind::never;
+      } else if (
+          predBlocking == blocking_kind::always_inline &&
+          predBlocking == blocking_kind::always_inline) {
```
kirkshoop: change second `predBlocking` to `succBlocking`?

include/unifex/sequence.hpp
```diff
@@ -0,0 +1,493 @@
...
+                    Predecessor,
+                    Successor,
+                    std::remove_cvref_t<Receiver>>> &&
+                is_connectable_v<
+                    Successor,
+                    detail::sequence_successor_receiver<
+                        Predecessor,
+                        Successor,
+                        Receiver>>,
+            int> = 0>
```
ericniebler: Say what you will, but IMO the concepts emulation macros from the range library would be an improvement here.

PR #46, opened 2019-12-03
Add a finally() algorithm for senders

lewissbaker:
This adds a new `finally(source, completion) -> Sender` algorithm that produces a composed operation that starts `source` and then upon completion saves the result of `source` and launches `completion`. When `completion` completes (with a `void` value) then the composed operation completes on that context with the result from `source`. Otherwise, the composed operation completes with the 'done' or 'error' result from `completion`.

This is basically a refactoring of the existing `typed_via` algorithm with some name changes and a change in code-structure that was needed to be able to constrain `connect()` on the resulting sender (which was not possible when the inner receivers were defined as nested members of the operation-state class which might not be valid to instantiate).

NOTE: The `finally()` algorithm currently does not pass through the consumer's stop-token to the completion operation. The thinking here is that the completion sender may be executing some cleanup code that we don't want to be cancelled. However, I'm not sure whether this is the right choice or not.

### DIFF
--- a/doc/api_reference.md
+++ b/doc/api_reference.md
@@ -89,13 +89,21 @@ Returns a sender that transforms the value of the `predecessor` by calling
 Returns a sender that will first launch `source` and upon completion of
 `source` will launch the `completion` sender.
 
-If `completion` completes with `set_value()` (which must ) then the
-composed operation completes with the result of `source`.
-Otherwise, if `completion` sender completes with 'done' or 'error'
+If `completion` completes with `set_value()` (which must complete with an
+empty value pack) then the composed operation completes with the result of
+`source`.
+Otherwise, if `completion` sender completes with `set_done` or `set_error`
 then the composed operation completes with the result of `completion`.
 
-Note that `completion` sender must complete with an empty value pack.
-ie. be a `void`-value sender.
+The composed finally-operation will complete inline on the execution context
+that the `completion` sender completes on, except in the case that the call
+to `connect()` on the completion-sender exits with an exception, in which case
+the operation will complete with `set_error()` inline on whatever execution
+context the `source` sender completed on.
+
+Note that `completion` sender must complete with an empty value pack
+if it completes with `set_value`.
+ie. it must be a `void`-value sender.
 
 ### `via(Sender successor, Sender predecessor) -> Sender`
 
--- a/doc/api_reference.md
+++ b/doc/api_reference.md
@@ -6,6 +6,7 @@
   * `get_allocator()`
 * Sender Algorithms
   * `transform()`
+  * `finally()`
   * `via()`
   * `typed_via()`
   * `on()`
@@ -82,6 +83,19 @@ See the [Cancellation](cancellation.md) section for more details on cancellation
 
 Returns a sender that transforms the value of the `predecessor` by calling
 `func(value)`.
+
+### `finally(Sender source, Sender completion) -> Sender`
+
+Returns a sender that will first launch `source` and upon completion of
+`source` will launch the `completion` sender.
+
+If `completion` completes with `set_value()` (which must ) then the
+composed operation completes with the result of `source`.
+Otherwise, if `completion` sender completes with 'done' or 'error'
+then the composed operation completes with the result of `completion`.
+
+Note that `completion` sender must complete with an empty value pack.
+ie. be a `void`-value sender.
 
 ### `via(Sender successor, Sender predecessor) -> Sender`
 
--- a/include/unifex/finally.hpp
+++ b/include/unifex/finally.hpp
@@ -16,14 +16,12 @@
 #pragma once
 
 #include <unifex/async_trace.hpp>
-#include <unifex/get_stop_token.hpp>
 #include <unifex/manual_lifetime.hpp>
 #include <unifex/manual_lifetime_union.hpp>
 #include <unifex/receiver_concepts.hpp>
 #include <unifex/scope_guard.hpp>
 #include <unifex/sender_concepts.hpp>
 #include <unifex/type_traits.hpp>
-#include <unifex/unstoppable_token.hpp>
 
 #include <cassert>
 #include <exception>
@@ -149,11 +147,6 @@ namespace unifex
         std::invoke(func, r.get_receiver());
       }
 
-      friend unstoppable_token tag_invoke(
-          tag_t<get_stop_token>, const finally_value_receiver&) noexcept {
-        return {};
-      }
-
     private:
       const Receiver& get_receiver() const noexcept {
         return op_->receiver_;
@@ -257,11 +250,6 @@ namespace unifex
         std::invoke(func, r.get_receiver());
       }
 
-      friend unstoppable_token tag_invoke(
-          tag_t<get_stop_token>, const finally_error_receiver&) noexcept {
-        return {};
-      }
-
     private:
       const Receiver& get_receiver() const noexcept {
         return op_->receiver_;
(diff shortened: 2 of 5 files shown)


doc/api_reference.md
```diff
@@ -83,6 +84,19 @@ See the [Cancellation](cancellation.md) section for more details on cancellation
 Returns a sender that transforms the value of the `predecessor` by calling
 `func(value)`.
 
+### `finally(Sender source, Sender completion) -> Sender`
+
+Returns a sender that will first launch `source` and upon completion of
+`source` will launch the `completion` sender.
+
+If `completion` completes with `set_value()` (which must ) then the
```
ericniebler: "which must..." ???

include/unifex/finally.hpp
```diff
@@ -0,0 +1,679 @@
...
+      void set_value() && noexcept {
+        auto* op = op_;
+        op->completionDoneOp_.destruct();
+        unifex::set_done(static_cast<Receiver&&>(op->receiver_));
+      }
+
+      template <
+          typename Error,
+          std::enable_if_t<
+              std::is_invocable_v<decltype(unifex::set_error), Receiver, Error>,
```
ericniebler: I've been thinking that a formulation of `is_tag_invocable_v` such as the following would be handy:

```c++
template<auto& Tag, class...As>
inline constexpr bool is_tag_invocable_v =
  std::is_invocable_v<decltype(Tag), As...>;
```

Would change the above to `is_tag_invocable_v<unifex::set_error, Receiver, Error>`.

include/unifex/finally.hpp
```diff
@@ -0,0 +1,679 @@
...
+          tag_t<visit_continuations>,
+          const finally_done_receiver& r,
+          Func&& func) {
+        std::invoke(func, static_cast<const Receiver&>(r.op_->receiver_));
+      }
+
+      friend unstoppable_token
+      tag_invoke(tag_t<get_stop_token>, const finally_done_receiver&) noexcept {
+        return {};
+      }
```
ericniebler: Why not just give `get_stop_token` a default that returns `unstoppable_token`?

include/unifex/finally.hpp
```diff
@@ -0,0 +1,679 @@
...
+
+    private:
+      operation_type* op_;
+    };
```
ericniebler: All the `finally_(value|error|done)_receiver` class templates are very, very similar. Would be nice if there was some way to avoid spamming out nearly identical code three times.

include/unifex/finally.hpp
```diff
@@ -0,0 +1,679 @@
...
+      void set_value(Values&&... values) && noexcept {
+        auto* op = op_;
+        auto& valueStorage =
+            op->value_.template get<std::tuple<std::decay_t<Values>...>>();
+        try {
+          valueStorage.construct(static_cast<Values&&>(values)...);
+        } catch (...) {
+          std::move(*this).set_error(std::current_exception());
+          return;
+        }
```
ericniebler: I don't love the contortions you have to go through to make your code exception safe when using `manual_lifetime`. Had you opted to use `std::optional` instead with the implicit-convertibility hack for non-movable types, this would be much simpler.

include/unifex/finally.hpp
lewissbaker: We already have `is_tag_invocable_v<CPO, Args...>` to be a query that checks whether you can call `tag_invoke(cpo, args...)`.

But perhaps this could be named `is_cpo_invocable_v`?

include/unifex/finally.hpp
lewissbaker: The default behaviour of `get_stop_token()` is to return `unstoppable_token`.

This overload is required here so that we don't forward through the `get_stop_token()` query to the parent receiver via the generic `tag_invoke()` overload above on line 303.

include/unifex/finally.hpp
lewissbaker: We could probably factor out some of the common bits (like forwarding through of CPOs) to a CRTP base-class to reduce some of the duplication. I think the logic on the set_value/error/done methods probably needs to stay separate though.

doc/api_reference.md
lewissbaker: I think this was supposed to be something like "which must be invoked with an empty pack of values"??

kirkshoop:
@ericniebler, In Rx `finally(func)` is a side-effecting operation, like `tap()`, that invokes func with no arguments *after* any terminal function is called on the consumer (in Rx this would be `on_error` and `on_complete`, in S/R this is `set_value`, `set_error`, `set_done`)

lewissbaker:
> In Rx `finally(func)` is a side-effecting operation, like `tap()`, that invokes func with no arguments _after_ any terminal function is called on the consumer (in Rx this would be `on_error` and `on_complete`, in S/R this is `set_value`, `set_error`, `set_done`)

That semantic feels like it would probably be the wrong behaviour for sender/receiver with connect/start.
Once we call `set_value()` (or any of the other completion-signalling operations), that code can go on to do execute an arbitrary amount of other logic, possibly transition execution contexts, etc. This then makes it difficult to reason about how the cleanup work executed by a finally function is sequenced with respect to the logic of the continuation that receives the result.

If we want to do some cleanup after having delivered the result to the final consumer then it's probably something we should integrate into the async cleanup design rather than replicate the Rx model directly. Then this algorithm could simply become one that introduced a cleanup scope that ran the cleanup of `source` before producing the result.

I chose the name `finally()` here because it seemed to represent the equivalent of a hypothetical `try/finally` structure (like C# has).

In coroutines this might look like:
```
template<typename Source, typename Completion>
task<await_result_t<Source>> finally(Source source, Completion completion) {
  try {
    co_return co_await std::move(source);
  } finally {
    co_await std::move(completion);
  }
}
```

> with no specification on what execution context will be used to complete, just like take_until(). This feels like a conflation of concerns, execution context is specified by on() and via(), this algorithm specifies an order of events across two senders.

The algorithm is specified to complete on whatever execution context the `completion` sender completes on. I agree this is not (quite) the same as `on()` or `via()`, but it is equivalent to the behaviour of something like `sequence()` or `let()`.

include/unifex/finally.hpp
lewissbaker: Actually, this breaks one of the current examples `delayed_stream_cancellation.cpp` as it prevents interrupting the `schedule_at()` operation that delays delivery of the stream results.

I'll remove this for now and just let the parent-receiver's stop_token propagate through to the completion sender.

include/unifex/finally.hpp
kirkshoop: I am used to ding this as a template type that implements the common parts and stores a function that is invoked as the body of the method being parameterized.

doc/api_reference.md
```diff
@@ -92,13 +92,21 @@ Returns a sender that transforms the value of the `predecessor` by calling
...
 then the composed operation completes with the result of `completion`.
 
-Note that `completion` sender must complete with an empty value pack.
-ie. be a `void`-value sender.
+The composed finally-operation will complete inline on the execution context
+that the `completion` sender completes on, except in the case that the call
+to `connect()` on the completion-sender exits with an exception, in which case
+the operation will complete with `set_error()` inline on whatever execution
+context the `source` sender completed on.
+
```
kirkshoop: > on whatever execution
context the `source` sender completed on

I wonder how finally and on should compose with a scheduler that guarantees completion on a defined context. It would be surprising if they were able to thwart the scheduler contract, but perhaps that is impossible to maintain..

kirkshoop:
> The algorithm is specified to complete on whatever execution context the completion sender completes on. I agree this is not (quite) the same as on() or via(), but it is equivalent to the behaviour of something like sequence() or let().

I think that we agree here. My question is whether on and via can be reasonably defined in terms of an algorithm that uses an implicit execution context.

PR #48, opened 2019-12-08
Add materialize() and dematerialize() sender algorithms

lewissbaker:
The `materialize()` algorithm takes an input sender and returns a new sender that transforms the completion signals (ie. `set_value`, `set_error` and `set_done`) into the value channel (ie. calls to `set_value`) by prepending the respective completion CPO as an additional argument to `set_value`.
ie.

* `set_value(r, values...)` -> `set_value(r, set_value, values...)`
* `set_error(r, error)` -> `set_value(r, set_error, error)`
* `set_done(r)` -> `set_value(r, set_done)`

The `dematerialize()` algorithm reverses the transformation of `materialize()`, turning calls:
* `set_value(r, set_value, values...)` -> `set_value(r, values...)`
* `set_value(r, set_error, error)` -> `set_error(r, error)`
* `set_value(r, set_done)` -> `set_done(r)`
* `set_error(r, e)` and `set_done(r)` pass through unchanged

Also adds a `type_list<Ts...>` helper for manipulating lists of types and a `concat_type_lists_unique<UniqueLists...>` metafunction that operates on `type_list` instances and can be used to deduplicate entries of a type-list more reliably compared with doing this with using whatever type the user passes in for `Variant` parameter of `value_types`/`error_types`.

### DIFF
--- a/doc/api_reference.md
+++ b/doc/api_reference.md
@@ -13,6 +13,8 @@
   * `sequence()`
   * `sync_wait()`
   * `when_all()`
+  * `materialize()`
+  * `dematerialize()`
   * `with_query_value()`
   * `with_allocator()`
 * Sender Types
@@ -208,6 +210,36 @@ to `value()`.
 If any of the input senders complete with done or error then it will request
 any senders that have not yet completed to stop and the operation as a whole
 will complete with done or error.
+
+### `materialize(Sender sender) -> Sender`
+
+Materializes the completion signal of `sender` into the value-channel by
+invoking prepending the completion arguments with the corresponding
+`set_value`, `set_error` or `set_done` CPO as an additional argument.
+
+ie. Transforms the following LHS completion signals to the RHS completion signals
+* `set_value(r, values...)` -> `set_value(r, set_value, values...)`
+* `set_error(r, e)` -> `set_value(r, set_error, e)`
+* `set_done(r)` -> `set_value(r, set_done)`
+
+This allows you to treat any result as a success and process the result as
+a value.
+
+### `dematerialize(Sender sender) -> Sender`
+
+Converts a sender of materialized signals into a sender of those signals.
+This reverses the transformation of signals performed by `materialize()`.
+
+If `sender` completes with `set_value(r, set_value, values...)` then the
+dematerialized sender will complete with `set_value(r, values...)`.
+
+Similarly if `sender` completes with `set_value(r, set_error, e)` then the
+dematerialized sender will complete with `set_error(r, e)`.
+
+And if `sender` completes with `set_value(r, set_done)` then the dematerialized
+sender will complete with `set_done(r)`.
+
+Any `set_error()` or `set_done()` signals are passed through unchanged.
 
 ### `with_query_value(Sender sender, CPO cpo, T value) -> Sender`
 
--- a/include/unifex/dematerialize.hpp
+++ b/include/unifex/dematerialize.hpp
@@ -1,0 +1,258 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


include/unifex/materialize.hpp
```diff
@@ -0,0 +1,272 @@
...
+        } else {
+          try {
+            unifex::set_value(
+                static_cast<Receiver&&>(receiver_),
+                unifex::set_error,
+                static_cast<Error&&>(error));
+          } catch (...) {
+            unifex::set_error(
+                static_cast<Receiver&&>(receiver_), std::current_exception());
+          }
```
ericniebler: Conditionally wrapping in `try`/`catch` should not be necessary. If everything in a `try` block is `noexcept`, the compiler is able to eliminate the overhead of the `try`/`catch`.

include/unifex/materialize.hpp
```diff
@@ -0,0 +1,272 @@
...
+          typename Error,
+          std::enable_if_t<
+              std::is_invocable_v<
+                  decltype(unifex::set_value),
+                  Receiver,
+                  decltype(unifex::set_error),
+                  Error>,
+              int> = 0>
+      void set_error(Error&& error) && noexcept {
+        if constexpr (std::is_nothrow_invocable_v<
```
ericniebler: Unifex needs `is_nothrow_receiver_of_v`.

include/unifex/materialize.hpp
```diff
@@ -0,0 +1,272 @@
...
+                          decltype(unifex::set_value),
+                          Receiver,
+                          decltype(unifex::set_error),
```
ericniebler: Why `decltype` here instead of `tag_t`?

doc/api_reference.md
```diff
@@ -210,6 +212,36 @@ If any of the input senders complete with done or error then it will request
...
+invoking prepending the completion arguments with the corresponding
+`set_value`, `set_error` or `set_done` CPO as an additional argument.
+
+ie. Transforms the following LHS completion signals to the RHS completion signals
+* `set_value(r, values...)` -> `set_value(r, set_value, values...)`
+* `set_error(r, e)` -> `set_value(r, set_error, e)`
+* `set_done(r)` -> `set_value(r, set_done)`
+
+This allows you to treat any result as a success and process the result as
+a value.
```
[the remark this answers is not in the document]
lewissbaker: This could be used in conjunction with something like `when_all()` to allow you to inspect each of the results without having the failure/cancellation of one operation resulting in a failure of the entire operation.

include/unifex/dematerialize.hpp
```diff
@@ -0,0 +1,258 @@
...
+          CPO cpo,
+          const dematerialize_receiver& r,
+          Args&&... args) noexcept(std::
+                                       is_nothrow_invocable_v<
+                                           CPO,
+                                           const Receiver&,
+                                           Args...>)
+          -> std::invoke_result_t<CPO, const Receiver&, Args...> {
+        return static_cast<CPO&&>(cpo)(
+            std::as_const(r.receiver_), static_cast<Args&&>(args)...);
```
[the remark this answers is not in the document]
lewissbaker: This was mainly guarding against `Receiver` being an lvalue reference.
This may be overly defensive though.

include/unifex/materialize.hpp
lewissbaker: `tag_t` returns the decayed type.
Whereas when we invoke the CPO we are invoking it with the CPO objects themselves which will be const-lvalue-ref.

include/unifex/materialize.hpp
lewissbaker: This is trying to avoid instantiating the `set_error()` path if it will never be called.

include/unifex/materialize.hpp
lewissbaker: I think unifex could use an `is_nothrow_value_receiver_v<R, Values...>` (and possibly also an `is_value_receiver_v<R, Values...>`?)

Using `is_nothrow_receiver_of_v` here (from the paper) would be overconstraining as that also checks for `set_error(r, exception_ptr{})` which would not be required in this case and counterproductively would put us on the error-handling path in the case where `set_error()` was not invocable with an `exception_ptr`.

include/unifex/materialize.hpp
lewissbaker: Maybe this should be `decltype((unifex::set_value))` instead to actually have it decay to the type of the expression rather than the type of the declaration?

PR #50, opened 2020-01-02
An implementation of indexed_for as described in P1897R2

LeeHowes:
`indexed_for` applies an operation to data accessed by reference, across a provided indexed space and following the rules in a provided policy.

### DIFF
--- a/examples/p1897.cpp
+++ b/examples/p1897.cpp
@@ -53,30 +53,30 @@ struct int_iterator {
     return *this;
   }
 
+  bool operator!=(const int_iterator& rhs) {
+    return base_ != rhs.base_;
+  }
+
   int base_;
 };
 
 struct iota_view {
   int size_;
+
+  int_iterator begin() {
+    return int_iterator{0};
+  }
+
+  int_iterator end() {
+    return int_iterator{size_};
+  }
 };
-
-template<class T>
-int_iterator begin(const iota_view& r) {
-  return int_iterator{0};
-}
-
-template<class T>
-int_iterator end(const iota_view& r) {
-  return int_iterator{r.size_};
-}
-}
+} // namespace ranges
 
 int main() {
-// TODO: This is still acting as transform
-// TODO: This does not use the range yet or iterate
   auto result = sync_wait(indexed_for(
       just(42),
-      ranges::iota_view{3},
+      ranges::iota_view{10},
       execution::seq,
       [](int& x) {
         x = x + 3;
--- a/examples/p1897.cpp
+++ b/examples/p1897.cpp
@@ -52,14 +52,15 @@ struct int_iterator {
     return base_;
   }
 
-  int_iterator& operator++() {
+  int_iterator operator++() {
     ++base_;
     return *this;
   }
 
   int_iterator operator++(int) {
+    auto cur = *this;
     ++base_;
-    return *this;
+    return cur;
   }
 
   bool operator!=(const int_iterator& rhs) const {
--- a/examples/p1897.cpp
+++ b/examples/p1897.cpp
@@ -80,8 +80,8 @@ struct iota_view {
 int main() {
   auto result = sync_wait(indexed_for(
       just(42),
-      ranges::iota_view{10},
       execution::seq,
+      ranges::iota_view{10},
       [](int idx, int& x) {
         x = x + idx;
       }));
@@ -96,8 +96,8 @@ int main() {
   auto indexed_for_sender =
     indexed_for(
       std::move(just_sender),
-      ranges::iota_view{3},
       execution::seq,
+      ranges::iota_view{3},
       [](int idx, std::vector<int>& vec, const int& i){
         vec[idx] = vec[idx] + i + idx;
       });
--- a/examples/p1897.cpp
+++ b/examples/p1897.cpp
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <unifex/just.hpp>
+#include <unifex/let.hpp>
+#include <unifex/scheduler_concepts.hpp>
+#include <unifex/sync_wait.hpp>
+#include <unifex/timed_single_thread_context.hpp>
+#include <unifex/indexed_for.hpp>
+#include <unifex/when_all.hpp>
+
+#include <chrono>
+#include <iostream>
+
+using namespace unifex;
+using namespace std::chrono;
+using namespace std::chrono_literals;
+
+namespace execution {
+class sequenced_policy {};
+inline constexpr sequenced_policy seq{};
+}
+
+namespace ranges {
+struct int_iterator {
+  const int operator[](size_t offset) const {
+    return base_+offset;
+  }
+
+  const int operator*() const {
+    return base_;
+  }
+
+  int_iterator& operator++() {
+    ++base_;
+    return *this;
+  }
+
+  int_iterator& operator++(int) {
+    ++base_;
+    return *this;
+  }
+
+  int base_;
+};
+
+struct iota_view {
+  int size_;
+};
+
+template<class T>
+int_iterator begin(const iota_view& r) {
+  return int_iterator{0};
+}
+
+template<class T>
+int_iterator end(const iota_view& r) {
+  return int_iterator{r.size_};
+}
+}
+
+int main() {
+// TODO: This is still acting as transform
+// TODO: This does not use the range yet or iterate
+  auto result = sync_wait(indexed_for(
+      just(42),
+      ranges::iota_view{3},
+      execution::seq,
+      [](const int& x) {
+        return x+1;
+      }));
+
+  std::cout << "all done " << *result << "\n";
+
+  return 0;
+}
--- a/examples/p1897.cpp
+++ b/examples/p1897.cpp
@@ -78,8 +78,8 @@ int main() {
       just(42),
       ranges::iota_view{3},
       execution::seq,
-      [](const int& x) {
-        return x+1;
+      [](int& x) {
+        x = x + 3;
       }));
 
   std::cout << "all done " << *result << "\n";
--- a/examples/p1897.cpp
+++ b/examples/p1897.cpp
@@ -39,6 +39,10 @@ inline constexpr parallel_policy par{};
 namespace ranges {
 struct int_iterator {
   using value_type = int;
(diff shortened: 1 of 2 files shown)


include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+      const indexed_for_sender& sender) {
+    return blocking(sender.pred_);
+  }
+
+  template <typename Receiver>
+  struct indexed_for_receiver {
+    UNIFEX_NO_UNIQUE_ADDRESS Func func_;
+    UNIFEX_NO_UNIQUE_ADDRESS Policy policy_;
+    UNIFEX_NO_UNIQUE_ADDRESS Range range_;
+    UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
```
lewissbaker: I have been generally trying to avoid bundling up the state into the receiver and instead having it stored in the operation-state.

This reduces the number of times the state is copied/moved around as the receivers are wrapped and unwrapped, but also helps to make it easier to write exception-safe code that honors the sender protocol without having to be careful about making sure you don't leave a receiver in a partially moved-from state and thus leave you unable to deliver the result.

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+      for(auto idx = 0; idx < range.size(); ++idx) {
+        std::invoke((Func &&) func, start[idx], values...);
+      }
+    }
+
+    template <typename... Values>
+    void set_value(Values&&... values) && noexcept {
+      if constexpr (noexcept(std::invoke(
+                        (Func &&) func_, std::declval<typename Range::iterator::value_type>(), values...))) {
+        apply_func_with_policy(policy_, (Range&&) range_, (Func &&) func_, values...);
```
lewissbaker: This implementation is passing through the `values` passed to `set_value()` directly to the function.
This is fine because the invocation of `func_` is blocking and is guaranteed to complete before `set_value()` returns.

But what if some implementation wants to execute these concurrently on multiple threads?
We may not want to block the call to `set_value()` on the completion of tasks on other threads and so would presumably need to copy the `values...` to a stable storage location (in the operation-state?) first before launching the tasks on other threads.

Those specialisations may have to alter the `value_types` to apply `remove_cvref_t<Values...>` as they will be forwarding through the copies when it completes, rather than the original values passed to `set_value()`.

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+    friend void tag_invoke(
+        tag_t<visit_continuations>,
+        const indexed_for_receiver& r,
+        Visit&& visit) {
+      std::invoke(visit, r.receiver_);
+    }
+  };
+
+  template <typename Receiver>
+  auto connect(Receiver&& receiver) && {
```
lewissbaker: This method should probably be constrained to only allow receivers that are able to receive the appropriate result types.

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+    void apply_func_with_policy(const execution::parallel_policy& policy, Range&& range, Func&& func, Values&... values)
+        noexcept(noexcept(std::invoke((Func &&) func_, std::declval<typename Range::iterator::value_type>(), values...))) {
+      auto start = range.begin();
+      for(auto idx = 0; idx < range.size(); ++idx) {
+        std::invoke((Func &&) func, start[idx], values...);
+      }
+    }
+
+    template <typename... Values>
+    void set_value(Values&&... values) && noexcept {
```
lewissbaker: This should probably be constrained to only allow value types that can both be passed to func_ and that can be passed to receiver_.set_value().

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+  template <typename... Args>
+  using is_overload_noexcept = std::bool_constant<noexcept(
+      std::invoke(std::declval<Func>(), std::declval<Args>()...))>;
+
+  template <template <typename...> class Variant>
+  struct calculate_errors {
+   public:
+    template <typename... Errors>
+    using apply = std::conditional_t<
```
lewissbaker: I think in general we need to always add std::exception_ptr to the list of error_types here since we don't yet know whether or not the output-receiver will have a noexcept `set_value()` or not.

This is a side-effect of allowing `set_value()` to be potentially-throwing.

Once I make the change to allow also putting the `value_types` and `error_types` onto the operation-state, then we can more accurately deduce whether or not something is non-throwing or not.
The operation_state::error_types might be able to report a reduced subset of the error types reported by the sender::error_types member.

So, for now, we can probably simplify this to remove the conditional_t and just always append exception_ptr.

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+  template <typename Receiver>
+  struct indexed_for_receiver {
+    UNIFEX_NO_UNIQUE_ADDRESS Func func_;
+    UNIFEX_NO_UNIQUE_ADDRESS Policy policy_;
+    UNIFEX_NO_UNIQUE_ADDRESS Range range_;
+    UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+
+    // sequenced_policy version supports forward range
+    template<typename... Values>
+    void apply_func_with_policy(const execution::sequenced_policy& policy, Range&& range, Func&& func, Values&... values)
```
lewissbaker: Could be declared `static`.

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+        noexcept(noexcept(std::invoke((Func &&) func_, std::declval<typename Range::iterator::value_type>(), values...))) {
+      for(auto idx : range) {
+        std::invoke((Func &&) func, idx, values...);
+      }
+    }
+
+    // parallel_policy version requires random access range
+    template<typename... Values>
+    void apply_func_with_policy(const execution::parallel_policy& policy, Range&& range, Func&& func, Values&... values)
```
lewissbaker: Could be declared `static`.

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+        noexcept(noexcept(std::invoke((Func &&) func_, std::declval<typename Range::iterator::value_type>(), values...))) {
```
lewissbaker: Should be `func` rather than `func_`?

I think you should be invoking the lvalue-overload of `Func::operator()` since you will be calling it multiple times.

I think you should be using the `iterator::reference` type rather than the `iterator::value_type` here.
Also, prefer to use `std::iterator_traits` to find the members rather than looking for nested type-names.
The nested type-names won't work if the iterator type is a pointer (like it might be for `std::vector`/`std::array`).

You could use `std::is_nothrow_invocable_v<Func&, typename std::iterator_traits<typename Range::iterator>::reference, Values&...>`.

Technically you probably also need to check for noexceptness of:
* `range.size()`
* `iter::operator[]`, `iter::operator++()`

Sometimes I really wish we had a `noexcept(auto)` 😜

include/unifex/indexed_for.hpp
lewissbaker: What should the behaviour be here if you pass a policy of `execution::par` but pass a range that is an input or forward range? Should it fall back to the sequential implementation?

The existing parallel algorithms often only require forward-iterators rather than random-access iterators.

Should this also be tag-dispatching based on the `iterator_category`?

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+  template <template <typename...> class Variant>
+  struct calculate_errors {
+   public:
+    template <typename... Errors>
+    using apply = std::conditional_t<
+        Predecessor::
+            template value_types<std::conjunction, is_overload_noexcept>::value,
+        Variant<Errors...>,
+        deduplicate_t<Variant<Errors..., std::exception_ptr>>>;
```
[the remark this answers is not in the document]
LeeHowes: That code is copy/paste from transform, I believe.

include/unifex/indexed_for.hpp
LeeHowes: Makes sense. Copy/paste from transform though so this is a wider change that should be made later.

include/unifex/indexed_for.hpp
LeeHowes: The next PR actually has to do that anyway. Sensible backport.

LeeHowes: Also simplifies the noexcept specification.

include/unifex/indexed_for.hpp
LeeHowes: It should. Making it static forced me to correct that downstream.

Yes, it took me a while to get that noexcept(noexcept( nonsense compiling at all. PR 51 had a rewrite anyway because invoking as above needs this, because of func_ but more generally was messy so I cleaned it up to use the trait.

include/unifex/indexed_for.hpp
LeeHowes: I think it should fail. We can't fully sanity check applications of func to see if they are par-safe, but requiring random access is a minimal check on the iterator. I was planning to change the text in P1897 to just require random_access for non-seq policies, forward for seq.

I moved the policy before the range because I think it is a declaration of the correctness of access to both the range and the function. If you do that and pass a forward range then that is immediately undefined behaviour by the terms of the declaration the programmer is making, but it's one we can check.

lewissbaker: Could you relax it further to 'input range' for seq policy, at least for single-pass algorithms like this?
That would allow using things like generators as the range.

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,163 @@
...
+        noexcept(std::is_nothrow_invocable_v<Func, typename std::iterator_traits<typename Range::iterator>::reference, Values...>) {
+      auto start = range.begin();
+      for(auto idx = 0; idx < range.size(); ++idx) {
+        std::invoke(func, start[idx], values...);
+      }
+    }
+
+    template <typename... Values>
+    void set_value(Values&&... values) && noexcept {
+      if constexpr (std::is_nothrow_invocable_v<Func, typename std::iterator_traits<typename Range::iterator>::reference, Values...>) {
```
lewissbaker: Should be `std::is_nothrow_invocable_v<Func&, ...>` (note the lvalue-reference).

examples/p1897.cpp
```diff
@@ -53,12 +57,12 @@ struct int_iterator {
     return *this;
   }
 
-  int_iterator& operator++(int) {
+  int_iterator operator++(int) {
     ++base_;
     return *this;
```
lewissbaker: You need to take a copy of *this before doing the increment.

include/unifex/indexed_for.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+  template <template <typename...> class Tuple>
+  struct indexed_for_result {
+   public:
+    template <typename... Args>
+    using apply = Tuple<Args...>;
+  };
+
+  template<typename... Args>
+  using is_overload_noexcept = std::is_nothrow_invocable<Func, Args...>;
```
lewissbaker: This appears to be unused now.

include/unifex/indexed_for.hpp
LeeHowes: Yes, I think so.

LeeHowes:
@lewissbaker Fixes made.

PR #51, opened 2020-01-02
indexed_for customisation

LeeHowes:
Builds on https://github.com/facebookexperimental/libunifex/pull/50 by adding a customisation of indexed_for using tag_invoke.

This customisation specialises the indexed_for algorithm when the predecessor was indexed_for, to not use connect/start between the two. This is a simple simulation of how we might customise more specifically, such that an implementation of indexed_for might trigger off of the type returned by applying via to a known executor type, for example.

### DIFF
--- a/include/unifex/indexed_for.hpp
+++ b/include/unifex/indexed_for.hpp
@@ -37,7 +37,11 @@ template <typename Predecessor, typename Policy, typename Range, typename Func>
 struct indexed_for_sender {
   UNIFEX_NO_UNIQUE_ADDRESS Predecessor pred_;
   UNIFEX_NO_UNIQUE_ADDRESS Policy policy_;
+<<<<<<< HEAD
   UNIFEX_NO_UNIQUE_ADDRESS Range range_;
+=======
+  UNIFEX_NO_UNIQUE_ADDRESS RangeOrRangeSelector rangeOrSelector_;
+>>>>>>> Majority of requested changes
   UNIFEX_NO_UNIQUE_ADDRESS Func func_;
 
   template <template <typename...> class Tuple>
@@ -74,7 +78,7 @@ struct indexed_for_sender {
   // sequenced_policy version supports forward range
   template<typename Range, typename... Values>
   static void apply_func_with_policy(const execution::sequenced_policy& policy, Range&& range, Func&& func, Values&... values)
-      noexcept(std::is_nothrow_invocable_v<Func, typename Range::iterator::value_type, Values...>) {
+      noexcept(std::is_nothrow_invocable_v<Func&, typename Range::iterator::value_type, Values...>) {
     for(auto idx : range) {
       std::invoke(func, idx, values...);
     }
@@ -83,7 +87,7 @@ struct indexed_for_sender {
   // parallel_policy version requires random access range
   template<typename Range, typename... Values>
   static void apply_func_with_policy(const execution::parallel_policy& policy, Range&& range, Func&& func, Values&... values)
-      noexcept(std::is_nothrow_invocable_v<Func, typename Range::iterator::value_type, Values...>) {
+      noexcept(std::is_nothrow_invocable_v<Func&, typename Range::iterator::value_type, Values...>) {
     auto start = range.begin();
     for(auto idx = 0; idx < range.size(); ++idx) {
       std::invoke(func, start[idx], values...);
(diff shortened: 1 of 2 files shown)


include/unifex/indexed_for.hpp
```diff
@@ -74,39 +74,183 @@ struct indexed_for_sender {
...
     UNIFEX_NO_UNIQUE_ADDRESS Range range_;
     UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
 
-    // sequenced_policy version supports forward range
-    template<typename... Values>
-    static void apply_func_with_policy(const execution::sequenced_policy& policy, Range&& range, Func&& func, Values&... values)
-        noexcept(std::is_nothrow_invocable_v<Func, typename std::iterator_traits<typename Range::iterator>::reference, Values...>) {
-      for(auto idx : range) {
-        std::invoke(func, idx, values...);
+    template <typename... Values>
```
LeeHowes: No wonder the rebase messed up. Clearly this diff algorithm got very confused with this similar-but-different code.

LeeHowes:
After a lot of pain (git merges and rebases are very confusing)  I have put this on top of the range factory PR.

There is a linear history, for now. If we can merge the earlier two with minimal changes that'd be helpful.

include/unifex/indexed_for.hpp
```diff
@@ -151,29 +318,58 @@ struct indexed_for_sender {
...
-auto indexed_for(Sender&& predecessor, Policy&& policy, RangeOrRangeSelector&& range, Func&& func) {
-  return indexed_for_sender<
-      std::remove_cvref_t<Sender>, std::decay_t<Policy>, std::decay_t<RangeOrRangeSelector>, std::decay_t<Func>>{
-      (Sender &&) predecessor, (Policy&&) policy, (RangeOrRangeSelector&& ) range, (Func &&) func};
+// Customisation for when two indexed_for_senders are chained
+// This simulates how we might customise an arbitrary set of types for
+// optimal interaction
+template <typename InnerPredecessor, typename InnerPolicy, typename InnerRange, typename InnerFunc, typename Policy, typename Range, typename Func>
+auto
+tag_invoke(
```
lewissbaker: We should try to find a way to avoid putting this definition at namespace scope where it could be added to the overload-set for every invocation of every unifex CPO.

One idea is to define the `indexed_for_cpo` type in a nested cpo-specific namespace and then put the overload in there. Other ways are possible by declaring things in a different order.

include/unifex/indexed_for.hpp
```diff
@@ -151,29 +318,58 @@ struct indexed_for_sender {
...
+      (Policy&&) policy,
+      (Range&& ) range,
+      (Func &&) func};
+}
+
+template <typename Sender, typename Policy, typename Range, typename Func>
+auto indexed_for(Sender&& predecessor, Policy&& policy, Range&& range, Func&& func) {
+  return indexed_for_customization(
+      (Sender &&) predecessor, (Policy&&) policy, (Range&& ) range, (Func &&) func);
 }
```
lewissbaker: I don't think we need this function. We can just rename the `indexed_for_customization` CPO to `indexed_for`.

include/unifex/indexed_for.hpp
```diff
@@ -86,42 +86,209 @@ struct indexed_for_sender {
...
-    template<typename Range, typename... Values>
-    static void apply_func_with_policy(const execution::sequenced_policy& policy, Range&& range, Func&& func, Values&... values)
-        noexcept(std::is_nothrow_invocable_v<Func, typename std::iterator_traits<typename Range::iterator>::reference, Values...>) {
-
-      for(auto idx : range) {
-        std::invoke(func, idx, values...);
+    template <typename... Values>
+    void set_value(Values&&... values) && noexcept {
+      using Range = std::invoke_result_t<indexed_for_detail::extract_range, RangeOrRangeSelector&&, Values&...>;
+      auto range = indexed_for_detail::extract_range{}(std::move(range_or_selector_), values...);
```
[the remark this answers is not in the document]
LeeHowes: Possibly. decltype(auto) is one of those unnecessarily subtle features of the language that I can never remember the point of.

include/unifex/indexed_for.hpp
LeeHowes: I put all of this in a detail namespace as I was tidying the CPO anyway.

Maybe we should define a namespace for each algorithm with a known name, and that's where you are supposed to put tag_invoke customisations?

LeeHowes:
@lewissbaker I've rebased this off of the factory change. That will be easy to reimplement on top if necessary and from the discussion it sounds like we'll be changing it.

LeeHowes:
By off, I mean I have moved it away from, rather than rebasing it on the factory diff. Not sure quite how "rebase off" would come across ;)

LeeHowes:
Yes, this was a proof of concept. We've moved on from this now.

PR #54, opened 2020-01-06
Use factory form of indexed_for to generate a range at async runtime

LeeHowes:
Implementation of the factory-form of indexed_for where a factory converts the inputs to a range to execute over, allowing adaptation at runtime to the arguments flowing through the pipeline.


AnujYamdagni (changes requested):

LeeHowes:
Hmm, I missed this one. Closing.

PR #55, opened 2020-01-15
fixes for build issues on linux

### LINKED ISSUE #45: Libunifex does not compile on Linux
It appears that there is a namespace called `linux`. For some reason, Clang 9 #define's linux to be 1 on my platform (Ubuntu 18.04). I'm using the clang-9 from apt.llvm.org. Error below:

/usr/bin/clang++-9   -I../source/../include -Iinclude -stdlib=libc++ -std=gnu++2a -MD -MT source/CMakeFiles/unifex.dir/linux/monotonic_clock.cpp.o -MF source/CMakeFiles/unifex.dir/linux/monotonic_clock.cpp.o.d -o source/CMakeFiles/unifex.dir/linux/monotonic_clock.cpp.o -c ../source/linux/monotonic_clock.cpp
In file included from ../source/linux/monotonic_clock.cpp:16:
../source/../include/unifex/linux/monotonic_clock.hpp:24:11: error: expected identifier or '{'
namespace linux {
          ^
<built-in>:392:15: note: expanded from here
#define linux 1
              ^

kirkshoop:
this should fix #45 in addition to the issues found once that was fixed.

### DIFF
--- a/cmake/unifex_flags.cmake
+++ b/cmake/unifex_flags.cmake
@@ -29,7 +29,7 @@ else()
 endif()
 
 # Probe for libUring support
-find_package(LibUring COMPONENTS Experimental Final)
+find_package(LibUring COMPONENTS)
 # Set some variables to be used by configure_file.
 if(LIBURING_FOUND)
   set(UNIFEX_NO_LIBURING FALSE)
--- a/cmake/unifex_flags.cmake
+++ b/cmake/unifex_flags.cmake
@@ -27,3 +27,15 @@ if(CXX_MEMORY_RESOURCE_HAVE_PMR)
 else()
   set(UNIFEX_NO_MEMORY_RESOURCE TRUE)
 endif()
+
+# Probe for libUring support
+find_package(LibUring COMPONENTS Experimental Final)
+# Set some variables to be used by configure_file.
+if(LIBURING_FOUND)
+  set(UNIFEX_NO_LIBURING FALSE)
+  set(UNIFEX_URING_HEADER ${LIBURING_LIBRARIES})
+  set(UNIFEX_URING_LIBRARY ${LIBURING_INCLUDE_DIRS})
+else()
+  set(UNIFEX_NO_LIBURING TRUE)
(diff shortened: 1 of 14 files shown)


include/unifex/detail/intrusive_heap.hpp
```diff
@@ -32,7 +32,7 @@ class intrusive_heap {
       assert(item->*Prev == nullptr);
     }
     while (item != nullptr) {
-      std::printf("- %p\n", item);
+      printf("- %p\n", item);
```
lewissbaker: I think this was missing `#include <cstdio>` rather than being incorrectly scoped.
Although I'd like to replace this with some kind of `TRACE()` macro at some point that allows it to be compiled out if required.

include/unifex/linux/io_uring_context.hpp
```diff
@@ -42,7 +44,7 @@
 #include <sys/uio.h>
 
 namespace unifex {
-namespace linux {
+namespace linuxos {
```
lewissbaker: I was debating what to call this namespace.
I think linuxos is a reasonable tradeoff. 👍 

Do we change the directory name to match, though?

cmake/unifex_flags.cmake
```diff
@@ -27,3 +27,15 @@ if(CXX_MEMORY_RESOURCE_HAVE_PMR)
 else()
   set(UNIFEX_NO_MEMORY_RESOURCE TRUE)
 endif()
+
+# Probe for libUring support
+find_package(LibUring COMPONENTS Experimental Final)
```
lewissbaker: Do we need the Experimental/Final parameters here?

These are added to the find_package() calls above to indicate that they should search for both experimental/final versions of the standard-library features. I don't think that really applies to liburing.

include/unifex/linux/io_uring_context.hpp
kirkshoop: I don't think so - at least for this diff.

PR #57, opened 2020-01-17
add io_epoll_context - scheduler-only, socket io next

kirkshoop:
Copied the io_uring_context and then converted it to use epoll.

epoll does not work with file io and we have not done socket io as yet. 

Socket io is on the docket.


include/unifex/linux/io_epoll_context.hpp
```diff
@@ -0,0 +1,520 @@
...
+  std::optional<time_point> currentDueTime_;
+
+  bool remoteQueueReadSubmitted_ = false;
+  bool timersAreDirty_ = false;
+
+  std::uint32_t activeTimerCount_ = 0;
+
+  static const std::uint32_t maxCount_ = 256;
+  epoll_event completed_[maxCount_];
+  std::uint32_t count_;
```
lewissbaker: Can we remove these data-members and just make them local variables inside `acquire_completion_queue_items()`?

include/unifex/linux/io_epoll_context.hpp
```diff
@@ -0,0 +1,520 @@
...
+
+  // Execute all ready-to-run items on the local queue.
+  // Will not run other items that were enqueued during the execution of the
+  // items that were already enqueued.
+  // This bounds the amount of work to a finite amount.
+  void execute_pending_local() noexcept;
+
+  // Check if any completion queue items are available and if so add them
+  // to the local queue.
+  void acquire_completion_queue_items() noexcept;
```
lewissbaker: I think this method should probably be modified to incorporate the call to `epoll_wait()` and have it allocate the buffer of `epoll_event` objects on the stack instead of as data-members.

Also, consider renaming this method to something more epoll-specific (the "completion queue" terminology comes from io_uring).

source/linux/io_epoll_context.cpp
```diff
@@ -0,0 +1,485 @@
...
+        timersAreDirty_ = false;
+      }
+    }
+  } else {
+    time_point now = monotonic_clock::now();
+    const auto earliestDueTime = timers_.top()->dueTime_;
+    LOGX(
+      "next timer in %i ms\n",
+      (int)std::chrono::duration_cast<std::chrono::milliseconds>(
+          earliestDueTime - now)
```
lewissbaker: Consider inlining the call to `monotonic_clock::now()` here to avoid the extra call in the case that logging is disabled.
The `now` variable is not used outside of the logging statement.

source/linux/io_epoll_context.cpp
```diff
@@ -0,0 +1,485 @@
...
+    if (currentDueTime_) {
+      constexpr auto threshold = std::chrono::microseconds(1);
+      if (earliestDueTime < (*currentDueTime_ - threshold)) {
+        LOG("active timer, need to cancel and submit an earlier one");
+
+        // An earlier time has been scheduled.
+        // Cancel the old timer before submitting a new one.
+        if (try_submit_timer_io_cancel()) {
+          currentDueTime_.reset();
+          if (try_submit_timer_io(earliestDueTime)) {
```
lewissbaker: As we are using a single timerfd there is no need to make two calls to `timerfd_settime()` here - one to cancel and another to set a new due time. We can just set a new time.

include/unifex/linux/io_epoll_context.hpp
```diff
@@ -0,0 +1,520 @@
...
+ public:
+  class schedule_sender;
+  class schedule_at_sender;
+  template <typename Duration>
+  class schedule_after_sender;
+  class read_sender;
+  class write_sender;
+  class async_read_only_file;
+  class async_read_write_file;
+  class async_write_only_file;
```
lewissbaker: These forward declarations have no implementation.
Remove them for now?

source/linux/io_epoll_context.cpp
```diff
@@ -0,0 +1,485 @@
...
+      count_ = result;
+    }
+  }
+}
+
+bool io_epoll_context::is_running_on_io_thread() const noexcept {
+  return this == currentThreadContext;
+}
+
+void io_epoll_context::schedule_impl(operation_base* op) {
```
lewissbaker: This method is not marked `noexcept` but all of the methods it's calling are marked `noexcept`.
Can we mark this one noexcept as well?
Or should we remove the noexcept declaration on schedule_remote()?

source/linux/io_epoll_context.cpp
```diff
@@ -0,0 +1,485 @@
...
+}
+
+void io_epoll_context::schedule_remote(operation_base* op) noexcept {
+  LOG("schedule_remote");
+  bool ioThreadWasInactive = remoteQueue_.enqueue(op);
+  if (ioThreadWasInactive) {
+    // We were the first to queue an item and the I/O thread is not
+    // going to check the queue until we signal it that new items
+    // have been enqueued remotely by writing to the eventfd.
+    signal_remote_queue();
```
lewissbaker: This method is potentially-throwing (if signalling the eventfd fails for some reason).
Should we remove the noexcept declaration here?

include/unifex/linux/io_epoll_context.hpp
```diff
@@ -0,0 +1,520 @@
...
+
+#include <atomic>
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <system_error>
+#include <utility>
+
+#include <sys/epoll.h>
+#include <sys/timerfd.h>
```
lewissbaker: Can we move this `#include` to the .cpp file?
I don't think we use anything from it in the header.

include/unifex/linux/io_epoll_context.hpp
```diff
@@ -0,0 +1,520 @@
...
+#include <unifex/linux/safe_file_descriptor.hpp>
+
+#include <atomic>
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <system_error>
+#include <utility>
+
+#include <sys/epoll.h>
```
lewissbaker: If we remove the epoll_event data-members from the class then we may be able to move this `#include` to the .cpp.

source/linux/io_epoll_context.cpp
```diff
@@ -0,0 +1,485 @@
...
+    }
+  }
+
+  LOG("io_epoll_context construction done");
+}
+
+io_epoll_context::~io_epoll_context() {
+  epoll_event event;
+  (void)epoll_ctl(epollFd_.get(), EPOLL_CTL_DEL, remoteQueueEventFd_.get(), &event);
+  (void)epoll_ctl(epollFd_.get(), EPOLL_CTL_DEL, timerFd_.get(), &event);
```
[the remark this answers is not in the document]
kirkshoop: the docs say that closing an fd will remove it from all epoll interest lists. Thus this is not strictly necessary. However, I think that it makes human readers worry more if they do not see the resource cleaned. This was intended to satiate humans. Another way to do that would be an epoll_item type that did the ADD/DEL for one epoll item and add a member for each of these.

source/linux/io_epoll_context.cpp
kirkshoop: I replaced the throw with an explicit terminate.

kirkshoop:
rebased on master.

include/unifex/linux/io_epoll_context.hpp
```diff
@@ -0,0 +1,495 @@
...
+  // to the remoteQueue and being told that the I/O thread is
+  // inactive.
+  void signal_remote_queue();
+
+  void remove_timer(schedule_at_operation* op) noexcept;
+  void update_timers() noexcept;
+  bool try_submit_timer_io(const time_point& dueTime) noexcept;
+
+  void* timer_user_data() const {
+    return const_cast<void*>(reinterpret_cast<const void*>(&timers_));
```
lewissbaker: The reinterpret_cast can be changed to a static_cast

include/unifex/linux/io_epoll_context.hpp
```diff
@@ -0,0 +1,495 @@
...
+  timer_heap timers_;
+
+  // The time that the current timer operation submitted to the kernel
+  // is due to elapse.
+  std::optional<time_point> currentDueTime_;
+
+  bool remoteQueueReadSubmitted_ = false;
+  bool timersAreDirty_ = false;
+
+  static const std::uint32_t maxCount_ = 256;
```
lewissbaker: This constant can be moved to the .cpp file.

include/unifex/linux/io_epoll_context.hpp
```diff
@@ -0,0 +1,495 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <unifex/config.hpp>
+#if !UNIFEX_NO_EPOLL
```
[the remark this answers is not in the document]
kirkshoop: I do not know why they encourage double negatives. I only care about consistency

PR #71, opened 2020-02-04
Add new_thread_context scheduler type

lewissbaker:
Adds `unifex::new_thread_context` which implements the `schedule()` interface by creating a new `std::thread` for each schedule operation.

When a thread finishes it installs itself as the 'thread-to-be-joined' in the context. If there was a previous thread that had not yet been joined then the newly finished thread takes responsibility for joining that thread before returning. The destructor of the new_thread_contex then only has to wait until all threads have been retired and then just join the last thread in the chain.

This ensures that we are able to join all of the threads and do so in a relatively timely manner without needing to store a collection of threads that could grow without bound.

Thanks to @kirkshoop for the idea for the implementation strategy!

### DIFF
--- a/include/unifex/new_thread_context.hpp
+++ b/include/unifex/new_thread_context.hpp
@@ -48,6 +48,13 @@ private:
       }
 
       try {
+        // Acquire the lock before launching the thread.
+        // This prevents the run() method from trying to read the thread_ variable
+        // until after we have finished assigning it.
+        //
+        // Note that this thread_ variable is private to this particular operation
+        // state and so will only be accessed by this start() method and the run()
+        // method.
         std::lock_guard opLock{mut_};
         thread_ = std::thread([this]() noexcept { this->run(); });
       } catch (...) {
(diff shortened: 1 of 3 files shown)


include/unifex/new_thread_context.hpp
```diff
@@ -0,0 +1,165 @@
...
+        unifex::set_error(std::move(receiver_), std::current_exception());
+      }
+    }
+
+  private:
+    void run() noexcept {
+      std::thread thisThread;
+      {
+        std::lock_guard opLock{mut_};
+        thisThread = std::move(thread_);
```
ericniebler: The assumption here is that `thisThread` ends up referring to the thread created on line 52, but I don't think that's necessarily the case. It's possible this code executed *before* the thread created on line 52 is assigned to `thread_`. Also, isn't it possible for another thread object to get written to `thread_` in a racy way?

lewissbaker: The `thread_` variable is a member of the operation-state class and so should have a 1:1 correspondence with a particular thread-creation. This variable is assigned only by the `start()` method that created the thread and is only read by the newly created thread.

We create the `std::thread()` while holding the `mut_` lock, so the new thread will not be able to acquire the lock until the thread that created it releases the lock, which only happens after we have completed assigning to the `thread_` variable.

lewissbaker: @kirkshoop also had the same concern on first reading.
I'll add a comment here to explain as I'm sure others will probably have similar concerns.

ericniebler: Ah. I got mixed up and thought this was the std::thread in the context.

ericniebler:
Fiendishly clever. I like it.

PR #75, opened 2020-02-27
Make libunifex play nicer with the wider C++ ecosystem, plus add in LLFIO

ned14:
Changes:

- Port third party dependency management over to cmake hunter, and fetch gtest from there from now on.
- Use cmake toolchain files for alternative compilers and STL implementations.
- Add in LLFIO and liburing as an external project.
- Make libunifex installable, and a good player for third party cmake inclusion.
- Add Ninja build support for Windows (untested, but should work)
- Coroutines detection no longer forces STL to libc++ on Linux, and now supports future GCC coroutines flags.


ned14:
FYI I depart tomorrow for my father-in-law's funeral, then will be onsite in MayStreet all next week, so I probably can't do more work on this until the week after. Comments shall be answered when I get the chance.

ericniebler:
Hi @ned14. Firstly, thank you for this, especially the cmake-y and ninja-y bits. If you couldn't tell, none of us are cmake experts. I had never heard of cmake hunter before. I'm happy (joyful, actually) to defer to your expertise here. I notice, however, that clang build is failing after your changes. Most of the failures are actually clang warnings, and we can clean those up.

However, I see that the tests are not linking with gtest anymore. If I had to guess, it's probably because gtest is getting compiled against libstdc++ and the libunifex tests are compiled against libc++. Could you look into this when you get a chance?

Also, the gcc build is failing on MacOS with a cmake error I don't understand.

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
libcxx_cxxexperimental
```

Thoughts?

ned14:
The link issue for clang on Linux smells to me like an ABI problem with std::string, that gift that keeps on giving for libstdc++. It should be easily fixable by asking cmake hunter to build gtest with a different ABI. I'll fix it.

For MacOS, it would seem that LLFIO no longer builds on MacOS. I had been told that had occurred, but I had very little incentive to fix it until now. My wife has a MacBook, so I'll borrow it next week when I get home and get that fixed.

Thanks for the feedback and patience.

ned14:
FYI I've not forgotten you, I received a slew of packaging bugs against Outcome from the OS distro packaging teams intending to ship Outcome with the major OSs. As they're higher priority, I'm slowly plugging through resolving those. ETA on returning to this work is next week. Sorry.

ned14:
I've reworked your libc++ selection to not be anti-social (#86), and to use cmake toolchain files instead, which is how you should be doing it. I've updated the Readme to explain how to use. The link errors for libc++ on Linux are now fixed.

Next week I'll borrow the wife's MacBook and fix the Mac problems. I'm also going to bring in liburing as a thirdparty git submodule, it'll superbuild it if cmake `find_package()` doesn't find a system install.

ned14:
Update: LLFIO now passes its test suite on Mac OS. It currently requires Mac OS 10.15 for its `<filesystem>` which now ships in XCode 11. Do you need libunifex to keep working on older Mac OS versions going into the future?

ned14:
My branch of libunifex now builds fine on osx with homebrew. The failing Travis tests are due to brew timing out during its very, very lengthy install.

This morning LLFIO went all-clean on C++ 20 with libc++. So by end of this week I'd say this PR will be ready, if you guys still wish to accept it.

ericniebler:
Thanks for all the CMake improvements, Niall. It looks good so far. The CI failures look odd, though. I see

```
-- Check for working CXX compiler: /usr/local/bin/g++-9 -- broken
CMake Error at /usr/local/Cellar/cmake/3.16.5/share/cmake/Modules/CMakeTestCXXCompiler.cmake:53 (message):
  The C++ compiler
    "/usr/local/bin/g++-9"
  is not able to compile a simple test program.
  It fails with the following output:
    Change Dir: /Users/travis/build/facebookexperimental/libunifex/build/CMakeFiles/CMakeTmp
    
    Run Build Command(s):/usr/local/bin/ninja cmTC_b4601 && [1/2] Building CXX object CMakeFiles/cmTC_b4601.dir/testCXXCompiler.cxx.o
    [2/2] Linking CXX executable cmTC_b4601
    FAILED: cmTC_b4601 
    : && /usr/local/bin/g++-9  -isysroot /Applications/Xcode-10.3.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names /usr/local/opt/llvm/lib CMakeFiles/cmTC_b4601.dir/testCXXCompiler.cxx.o  -o cmTC_b4601   && :
    ld: can't map file, errno=22 file '/usr/local/opt/llvm/lib' for architecture x86_64
    collect2: error: ld returned 1 exit status
    ninja: build stopped: subcommand failed.
```

Somehow Travis is installing a broken gcc-9. I've never seen that before. Where do you see a brew timeout?

ned14:
Woohoo, it's all finally working! Only the io_uring test fails, and that's because Travis has banned it. I'll add some code to not fatally exit the test if io_uring is not available, and we'll be good to go.

ericniebler:
Hi Niall. What do you make of the clang/xcode travis failures? And is the liburing problem due to Travis running older Linux kernels?

ned14:
> What do you make of the clang/xcode travis failures?

Yes the liburing problem is due to the kernel 5.0 used by Travis which doesn't have i/o uring kernel support. I've modified the test to fail gracefully if and only if i/o uring support is not there. You might want to consider improving libunifex's handling of that situation e.g. add a factory function which returns the "best available" kernel async i/o implementation, as most users will just want it to work.

The osx stuff is rather more curious. My wife's MacBook is on latest Catelina, whose xcode is sufficient to build LLFIO entirely without any requirement for brew LLVM. However travis is on osx 10.14, and for some reason I'll have to go pester Louis about, libc++'s <filesystem> refuses to run on anything before 10.15. You can see examples of this at https://travis-ci.org/github/ned14/libunifex/jobs/668033556. This seems to include recent editions of brew llvm's libc++ for 10.14, which is just "yay wonderful" I'm sure you'll agree.

I have tried xcode10.1 with osx 10.13 in the hope that its brew llvm contains a working <filesystem>.  Another alternative is to revert to llvm 8, but it'll have a much worse coroutines QOI. Worst comes to worst, we simply await travis shipping osx 10.15 Let's see how it goes.

ned14:
FYI brew llvm@8 appears to not ship the `<filesystem>` runtime, just the header. Fabulous. See https://travis-ci.org/github/ned14/libunifex/jobs/668210869

Back when LLFIO last worked on OS X, I had to hand compile my own copy of libc++ to get `<filesystem>`. That was several years ago. I find it amazing that things did not improve between then, and until just there very recently.

Do you want to hold off merging this PR until Travis ships an osx 10.15 image?

ericniebler:
@ned14, I have tried merging this with trunk. We've dropped travis in favor of GitHub action. Things mostly work, but configure on Win32 is failing with this:

```
[hunter ** FATAL ERROR **] Incorrect MSVC setup:
[hunter ** FATAL ERROR **]   At least one of the following should be an absolute path
[hunter ** FATAL ERROR **]   CMAKE_VS_DEVENV_COMMAND:()
[hunter ** FATAL ERROR **]   CMAKE_VS_MSBUILD_COMMAND:()
[hunter ** FATAL ERROR **] [Directory:D:/a/libunifex/libunifex]

------------------------------ ERROR -----------------------------
    https://hunter.readthedocs.io/en/latest/reference/errors/error.vs.devenv.html
------------------------------------------------------------------

CMake Error at C:/.hunter/_Base/Download/Hunter/0.23.251/5659b15/Unpacked/cmake/modules/hunter_error_page.cmake:12 (message):
Call Stack (most recent call first):
  C:/.hunter/_Base/Download/Hunter/0.23.251/5659b15/Unpacked/cmake/modules/hunter_fatal_error.cmake:20 (hunter_error_page)
  C:/.hunter/_Base/Download/Hunter/0.23.251/5659b15/Unpacked/cmake/modules/hunter_setup_msvc.cmake:148 (hunter_fatal_error)
  C:/.hunter/_Base/Download/Hunter/0.23.251/5659b15/Unpacked/cmake/modules/hunter_finalize.cmake:117 (hunter_setup_msvc)
  C:/.hunter/_Base/Download/Hunter/0.23.251/5659b15/Unpacked/cmake/modules/hunter_add_package.cmake:23 (hunter_finalize)
  CMakeLists.txt:75 (hunter_add_package)


-- Configuring incomplete, errors occurred!
See also "D:/a/libunifex/libunifex/build/CMakeFiles/CMakeOutput.log".
See also "D:/a/libunifex/libunifex/build/CMakeFiles/CMakeError.log".
CMake Error at D:/a/_temp/d9027134-1e5e-4f20-8f82-8cc412d7099a:36 (message):
  Bad exit status
```

https://github.com/facebookexperimental/libunifex/runs/644333728

ned14:
Thanks for attempting a rebase of this PR. I'll take a look next few days.

I've been making good progress with the abstracted third-party i/o multiplexer support for LLFIO. I'm currently writing a well-documented "null i/o multiplexer" to serve firstly for optimisation during a few rounds of statistical profiling, but secondly as documentation for you guys so you can plug libunifex into LLFIO and it'll all "just work".

I'll also write a test io_uring multiplexer as my Linux kernel is finally v5.3, just to ensure that my abstraction fully maxes out io_uring. It'll also test the new registered i/o buffers implementation, in theory all pipe i/o becomes invariant to i/o performed, because it would all simply become TLB shootdowns to remap pages between userspace and the Linux kernel, without ever copying a single byte. If it works well, it should provide very convincing benchmarks for presentation to WG21, doing no work is always faster than doing work!

ned14:
Just some quick benchmarks from the LLFIO i/o multiplexer work, which might be useful for you to compare against what you're doing. These are for MSVC in Release build, unity build, and the numbers are in *nanoseconds* from the beginning of the read, including the write, and receiving the completion of the write. The time between the completion of the write and release of the state is not included. Note that the high resolution timer granularity on this platform is 100 nanoseconds.

```c++
Benchmarking Null i/o multiplexer unsynchronised with 1 handles ...
   per-handle create 7e-06 cancel 0 destroy 6e-06
   total i/o min 100 max 3.49408e+08 mean 152.97 stddev 66587.2
             @ 50% 100 @ 95% 200 @ 99% 300 @ 99.9% 1500 @ 99.99% 3800
   total results collected = 47732735

Benchmarking Null i/o multiplexer synchronised with 1 handles ...
   per-handle create 1.54e-05 cancel 1e-07 destroy 4.2e-06
   total i/o min 100 max 2.66671e+08 mean 217.766 stddev 62937.9
             @ 50% 200 @ 95% 300 @ 99% 400 @ 99.9% 1600 @ 99.99% 4100
   total results collected = 27370495
```
The synchronised results are where locks are taken and released in case multiple threads might be awaiting i/o completions. Only one thread does so in this test. The synchronised results are about 20% slower.

No i/o is actually done, we just iterate through a null multiplexer which goes through the motions as if i/o were being done. This shows the minimum possible overheads for the LLIO multiplexer abstraction.

I'm going to claim that the 99.99% numbers are actually page fault allocated memory caused by the storage for the results. The 1500 nanosecond overhead is probably a reasonable worst case for this multiplexer abstraction.

Unfortunately, Linux can complete a synchronous i/o in 900 nanoseconds. That means for the 50% latency distribution case, i/o multiplexing would add ~22% overhead, and potentially more than double the latency in the worst case.

It's entirely possible that I simply wrote a crap design, and you guys can do much better. However almost all of the overhead goes on the i/o state lifecycle management, and I don't personally think it can be improved upon.

In case you're wondering, this newly refactored i/o multiplexer design (it is not what I presented at Prague) blows ASIO out of the water. If I deliberately neuter the test IOCP multiplexer so it behaves like-for-like with ASIO, it beats ASIO by 35%. If I let it go full belt, it beats it by over 50%. I would expect the numbers to be much better again on POSIX, but that is as yet untested.

I'll do a bit of optimisation work on this tomorrow morning, and then next week start into an io_uring test i/o multiplexer to make sure the abstraction suits io_uring correctly.

I'll try to have a look into github actions sometime soon. I had a play with them last night, there is a learning curve there, it didn't "just work" for my test github actions pipeline.

lewissbaker:
Closing this PR as it hasn't had any activity in quite a while and I think there's been a lot of cmake improvements to libunifex in the meantime.

ned14:
Yeah, so covid happened, my non-work free time plummeted, and then a new baby has since turned up, so that kinda ended this initiative for me. Sorry.

Wider picture, all of anything new which is bigger than trivial is obviously now C++ 26 facing, as whatever F2F time remains next year will clearly go on clearing whatever is towards the end of the C++ 23 pipeline. That has significantly changed my priority queue, as I'm sure it has for all of you and everybody else. For me anyway, I intend to focus on filesystem path view which is 23 IS track, then on status code which could still make the 23 IS track. Everything else is to be shunted into 2023 or later for the 26 IS.

I guess that means we shall be parting ways for a few years at least. But thanks for all your cooperation and welcome, and I hope to work together again with any of you in the future.

PR #76, opened 2020-02-27
Fixes for building with io_uring on Linux

lewissbaker:
- CMake fixes for liburing selection
- Work around removal of raw io_uring syscalls from liburing.
  Now define the syscall wrappers ouselves.
  This means we don’t need to link against liburing any more.
- Fix bug in io_uring example that failed to compile when `sequence()` was made more strict.


lewissbaker:
> I would prefer to have linux publish the uring syscalls in a header. Is this planned?

It seems linux doesn't expose functions for many syscalls but instead relies on them being exposed via other libraries. eg. posix functions implemented by glibc.

PR #78, opened 2020-03-03
Add stop_when() sender algorithm.

lewissbaker:
Allows you to request one operation to stop when a trigger operation completes.

This can be used to implement timeouts (e.g. by passing `schedule_after(100ms)` as the trigger) as well as other cancellation behaviours.

### DIFF
--- a/include/unifex/stop_when.hpp
+++ b/include/unifex/stop_when.hpp
@@ -28,6 +28,7 @@
 #include <variant>
 #include <optional>
 #include <atomic>
+#include <functional>
 
 namespace unifex
 {
--- a/include/unifex/stop_when.hpp
+++ b/include/unifex/stop_when.hpp
@@ -1,0 +1,339 @@
+/*
+ * Copyright 2020-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


ericniebler:
This needs the ADL isolation treatment to be consistent with the rest of libunifex.

lewissbaker:
> This needs the ADL isolation treatment to be consistent with the rest of libunifex.
@ericniebler This has now been updated.

include/unifex/stop_when.hpp
```diff
@@ -0,0 +1,429 @@
...
+                  std::apply(
+                      [&](auto set_xxx, auto&&... args) {
+                        set_xxx(
+                            std::move(receiver_),
+                            static_cast<decltype(args)>(args)...);
+                      },
+                      static_cast<decltype(tuple)>(tuple));
+                } else {
+                  // Should be unreachable
+                  std::terminate();
```
ericniebler, line 251: Why not make it a `static_assert` instead of an `if constexpr` if it can't happen?

lewissbaker: The variant does actually contain an empty tuple<> as the first alternative (a bit like the monostate).
So this callback will be instantiated for that alternative, we just should never be hitting that overload at runtime since this function should only be called after initialising the variant to some other alternative.

lewissbaker:
> I would like this to be renamed take_until I do not like different names for the same tool across task and stream..

I would also like them to be consistent, but I don't feel that the name take_until works very well for single-senders.

We can revisit the naming in future to try and merge the stream/many_sender/sender versions into a single algorithm once the many_sender/stream abstractions have settled.

PR #80, opened 2020-03-03
Remove stopToken parameter from sync_wait()

lewissbaker:
Now that we have `stop_when()`, and also the fallback manual injection
of a stop-token using `with_query_value()` we don't need to have `sync_wait()`
concerned with stop-tokens. If the caller wants to be able to stop an
operation then they can apply an appropriate algorithm to the sender
before passing it to `sync_wait()`.


lewissbaker:
I accidentally merged this into `stop_when` branch instead of into `master`.
I'll manually apply the changes to `master`

PR #84, opened 2020-03-06
Improve compile times of tag_invoke metafunctions

lewissbaker:
Reimplement the tag_invoke metafunctions manually rather than
basing them off the std::is_invocable/invoke_result metafunctions.

This seems to yield reasonable compile-time improvements.
The examples/let.cpp file saw a 20% reduction in template instantiation
time on clang/libc++ with this change.

Added some tag_invoke unit-tests to make sure I wasn't breaking anything.

### DIFF
--- a/include/unifex/tag_invoke.hpp
+++ b/include/unifex/tag_invoke.hpp
@@ -57,6 +57,12 @@ using namespace tag_invoke_cpo_ns;
 template <auto& CPO>
 using tag_t = std::remove_cvref_t<decltype(CPO)>;
 
+// Manually implement the traits here rather than defining them in terms of
+// the corresponding std::invoke_result/is_invocable/is_nothrow_invocable traits
+// to improve compile-times. We don't need all of the generality of the std::
+// traits and the tag_invoke traits are used heavily through libunifex so
+// optimising them for compile time makes a big difference.
+
 using tag_invoke_impl::tag_invoke_result_t;
 
 template<typename CPO, typename... Args>
--- a/include/unifex/tag_invoke.hpp
+++ b/include/unifex/tag_invoke.hpp
@@ -34,7 +34,7 @@ struct tag_invoke_cpo {
 };
 
 template<typename CPO, typename... Args>
-using tag_invoke_result_t = decltype(tag_invoke(static_cast<CPO&&(*)()>(nullptr)(), static_cast<Args&&(*)()>(nullptr)()...));
+using tag_invoke_result_t = decltype(tag_invoke(static_cast<CPO&&(*)() noexcept>(nullptr)(), static_cast<Args&&(*)() noexcept>(nullptr)()...));
 
 struct yes_type { char dummy; };
 struct no_type { char dummy[2]; };
@@ -42,10 +42,10 @@ struct no_type { char dummy[2]; };
 template<typename CPO, typename... Args>
 auto try_tag_invoke(int)
   noexcept(noexcept(tag_invoke(static_cast<CPO&&(*)() noexcept>(nullptr)(), static_cast<Args&&(*)() noexcept>(nullptr)()...)))
-  -> decltype(tag_invoke(static_cast<CPO&&(*)() noexcept>(nullptr)(), static_cast<Args&&(*)() noexcept>(nullptr)()...), yes_type{});
+  -> decltype(static_cast<void>(tag_invoke(static_cast<CPO&&(*)() noexcept>(nullptr)(), static_cast<Args&&(*)() noexcept>(nullptr)()...)), yes_type{});
 
 template<typename CPO, typename... Args>
-no_type try_tag_invoke(...);
+no_type try_tag_invoke(...) noexcept(false);
 
 } // namespace tag_invoke_impl
 
--- a/include/unifex/tag_invoke.hpp
+++ b/include/unifex/tag_invoke.hpp
@@ -19,67 +19,84 @@
 
 #include <type_traits>
 
-namespace unifex {
-namespace tag_invoke_impl {
+namespace unifex
+{
+  namespace tag_invoke_impl
+  {
+    void tag_invoke();
 
-void tag_invoke();
+    struct tag_invoke_cpo {
+      template <typename CPO, typename... Args>
+      constexpr auto operator()(CPO cpo, Args&&... args) const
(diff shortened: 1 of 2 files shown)


include/unifex/tag_invoke.hpp
```diff
@@ -32,6 +32,30 @@ struct tag_invoke_cpo {
     return tag_invoke((CPO&&)cpo, (Args &&) args...);
   }
 };
+
+template<typename CPO, typename... Args>
+using tag_invoke_result_t = decltype(tag_invoke(std::declval<CPO>(), std::declval<Args>()...));
```
ericniebler: `declval` instantiates templates because it must handle the `void` case. `CPO` and `Args` can never be `void`. If you want to make this compile even faster, you can replace `declval<T>()` with `static_cast<T&&(*)()>(nullptr)()`.

include/unifex/tag_invoke.hpp
```diff
@@ -42,26 +66,23 @@ using namespace tag_invoke_cpo_ns;
...
+};
 
 template <typename CPO, typename... Args>
-inline constexpr bool is_tag_invocable_v =
-    std::is_invocable_v<tag_t<tag_invoke>, CPO, Args...>;
+using is_tag_invocable = typename tag_invoke_impl::is_tag_invocable<CPO, Args...>::type;
 
 template <typename CPO, typename... Args>
-using is_tag_invocable = std::is_invocable<tag_t<tag_invoke>, CPO, Args...>;
+inline constexpr bool is_tag_invocable_v = is_tag_invocable<CPO, Args...>::value;
```
ericniebler: Instantiating variable templates is *much* cheaper than instantiating class templates. The class templates should be defined in terms of the variable templates, not the other way 'round.

lewissbaker: The implementation of this has already produced the result as a `std::bool_constant`.
I would need to change the implementation of the `try_call()` mechanism in `tag_invoke_impl::is_tag_invocable` to avoid this.

Any suggestions? Maybe using `sizeof(decltype(try_call(0)) == sizeof(yes_type)`?

include/unifex/tag_invoke.hpp
```diff
@@ -32,6 +32,21 @@ struct tag_invoke_cpo {
...
+template<typename CPO, typename... Args>
+using tag_invoke_result_t = decltype(tag_invoke(static_cast<CPO&&(*)()>(nullptr)(), static_cast<Args&&(*)()>(nullptr)()...));
+
+struct yes_type { char dummy; };
+struct no_type { char dummy[2]; };
+
+template<typename CPO, typename... Args>
+auto try_tag_invoke(int)
+  noexcept(noexcept(tag_invoke(static_cast<CPO&&(*)() noexcept>(nullptr)(), static_cast<Args&&(*)() noexcept>(nullptr)()...)))
+  -> decltype(tag_invoke(static_cast<CPO&&(*)() noexcept>(nullptr)(), static_cast<Args&&(*)() noexcept>(nullptr)()...), yes_type{});
```
ericniebler: Why not reuse `tag_invoke_result_t` in the return type here? Also, guard against overloaded `operator,` by putting `void(), ` before `yes_type{}` here.

include/unifex/tag_invoke.hpp
lewissbaker: The `sizeof()` approach seems to work.

I moved the `try_call` function overloads to namespace scope (and renamed to `try_tag_invoke`).
I also combined the `noexcept` test into the same function instantiation.

include/unifex/tag_invoke.hpp
```diff
@@ -32,6 +32,21 @@ struct tag_invoke_cpo {
     return tag_invoke((CPO&&)cpo, (Args &&) args...);
   }
 };
+
+template<typename CPO, typename... Args>
+using tag_invoke_result_t = decltype(tag_invoke(static_cast<CPO&&(*)()>(nullptr)(), static_cast<Args&&(*)()>(nullptr)()...));
```
ericniebler: Maybe use `noexcept` function pointers here as you do in `try_tag_invoke` below.

include/unifex/tag_invoke.hpp
```diff
@@ -32,6 +32,21 @@ struct tag_invoke_cpo {
     return tag_invoke((CPO&&)cpo, (Args &&) args...);
   }
 };
+
```
kirkshoop: comment explaining why this is manually reimplementing invocable traits would make it less likely for someone to later change it back to invocable traits without doing compiler time tests..

include/unifex/tag_invoke.hpp
lewissbaker: I was trying to use the same `noexcept`-qualified `static_cast` expressions in both the noexcept and return-type components - thinking that the compiler might be able to reuse the overload resolution results for both expressions.

If it doesn't make a difference then, yes, I may as well use `tag_invoke_result_t`.

include/unifex/tag_invoke.hpp
```diff
@@ -42,26 +57,23 @@ using namespace tag_invoke_cpo_ns;
...
-template <typename CPO, typename... Args>
-using tag_invoke_result = std::invoke_result<tag_t<tag_invoke>, CPO, Args...>;
+using tag_invoke_impl::tag_invoke_result_t;
 
-template <typename CPO, typename... Args>
-using tag_invoke_result_t =
-    std::invoke_result_t<tag_t<tag_invoke>, CPO, Args...>;
+template<typename CPO, typename... Args>
+struct tag_invoke_result {
+    using type = tag_invoke_result_t<CPO, Args...>;
```
ericniebler: Elsewhere we use 2-space indent.

include/unifex/tag_invoke.hpp
lewissbaker: I've added a comment to that effect.

ericniebler:
The next step is to define a `callable_result_t` as:

```c++
template<typename Fn, typename... Args>
using callable_result_t = decltype(
    static_cast<Fn&&(*)()noexcept>(nullptr)()(static_cast<Args&&(*)()noexcept>(nullptr)()...));
```

... and use it everyplace instead of `std::invoke_result_t` where the `Fn` is guaranteed to not be a member pointer. That should be most place.

Ditto for replacing `std::is_invocable_v` and `std::is_nothrow_invocable_v`.

PR #90, opened 2020-03-13
Remove the use of std::invoke machinery where it is unnecessary

ericniebler:


include/unifex/type_traits.hpp
```diff
@@ -80,6 +80,41 @@ struct decayed_tuple {
...
+      noexcept(_is_throw_callable_v<Fn, As...>);
+  std::false_type _try_call(...);
+} // namespace detail
+
+template <typename Fn, typename... As>
+using is_callable =
+    decltype(detail::_try_call(static_cast<Fn(*)(As...)>(nullptr)));
+
+template <typename Fn, typename... As>
+inline constexpr bool is_callable_v = is_callable<Fn, As...>::value;
```
lewissbaker: In `is_tag_invocable_v` we used `yes_type` and `no_type` of different sizes and compared result sizes to determine the bool result rather than looking up true_type/false_type::value. Then we defined `is_tag_invocable` in terms of `is_tag_invocable_v`.

Should we do that here too? Or is it ok if we use bool_constant since it's probably going to be instantiated anyway?

ericniebler: If you use any type trait, then `std::false_type` and `std::true_type` are instantiated.

include/unifex/allocate.hpp
```diff
@@ -141,7 +141,7 @@ namespace _alloc_cpo {
   public:
     template <typename Sender>
     auto operator()(Sender&& predecessor) const
-        noexcept(std::is_nothrow_invocable_v<
+        noexcept(is_nothrow_callable_v<
           _impl<is_tag_invocable_v<_fn, Sender>>, Sender>) {
```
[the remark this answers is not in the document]
ericniebler: I've measured. It makes no difference. I'd rather leave the variable template form since it reads a little better.

PR #95, opened 2020-03-21
add connect overloads for lvalue-ref and const-lvalue-ref

kirkshoop:
This adds overloads of `connect() &` and `connect() const&` to a variety of senders.

This turns out to be a non-issue for tag_invoke because `Target&&` is a forwarding-reference while `connect() &&` does not provide access to a forwarding-reference to `this`.


include/unifex/stop_immediately.hpp
```diff
@@ -296,6 +296,10 @@ struct _stream<SourceStream, Values...>::type {
     operation<Receiver> connect(Receiver&& receiver) && {
       return operation<Receiver>{stream_, (Receiver&&)receiver};
     }
+    template<typename Receiver>
+    void connect(Receiver&& receiver) & =delete;
```
lewissbaker: Is this overload declaration necessary given that we have the `const &` overload below?

include/unifex/stop_immediately.hpp
```diff
@@ -413,6 +417,14 @@ struct _stream<SourceStream, Values...>::type {
...
       return operation<Receiver>{stream_, (Receiver &&) receiver};
     }
+    template<typename Receiver>
+    operation<Receiver> connect(Receiver&& receiver) & {
+      return operation<Receiver>{stream_, (Receiver &&) receiver};
+    }
+    template<typename Receiver>
+    operation<Receiver> connect(Receiver&& receiver) const& {
+      return operation<Receiver>{stream_, (Receiver &&) receiver};
+    }
```
lewissbaker: I'm pretty sure it doesn't make sense to allow repeatedly running the cleanup operation here.
This cleanup operation implementation shared references to the stream object and assumes that it will only be run once - the state-machine does not allow for running the cleanup operations multiple times.

In general, I think cleanup operations should generally not be repeatable. They will often have side-effects that release resources and should only be run once.

It also seems suspicious that each of these are performing the same

include/unifex/sync_wait.hpp
```diff
@@ -188,7 +188,7 @@ namespace _sync_wait_cpo {
 
         auto operation = connect(
           (Sender&&)sender,
-          _sync_wait::thread_unsafe_receiver<Result, StopToken&&>{
+          _sync_wait::thread_unsafe_receiver<Result, StopToken>{
```
lewissbaker: Why was this change necessary?

include/unifex/thread_unsafe_event_loop.hpp
```diff
@@ -149,6 +149,16 @@ namespace _thread_unsafe_event_loop {
...
+    template <typename Receiver>
+    after_operation<Duration, std::remove_cvref_t<Receiver>> connect(Receiver&& r) & {
+      return after_operation<Duration, std::remove_cvref_t<Receiver>>{
+          (Receiver &&) r, duration_, loop_};
+    }
+    template <typename Receiver>
+    after_operation<Duration, std::remove_cvref_t<Receiver>> connect(Receiver&& r) const& {
+      return after_operation<Duration, std::remove_cvref_t<Receiver>>{
+          (Receiver &&) r, duration_, loop_};
+    }
```
lewissbaker: Can we combine these three overloads into a single `const&` overload to avoid duplication here?

include/unifex/thread_unsafe_event_loop.hpp
```diff
@@ -225,6 +235,16 @@ namespace _thread_unsafe_event_loop {
...
+    template <typename Receiver>
+    at_operation<std::remove_cvref_t<Receiver>> connect(Receiver&& r) & {
+      return at_operation<std::remove_cvref_t<Receiver>>{
+          (Receiver &&) r, dueTime_, loop_};
+    }
+    template <typename Receiver>
+    at_operation<std::remove_cvref_t<Receiver>> connect(Receiver&& r) const& {
+      return at_operation<std::remove_cvref_t<Receiver>>{
+          (Receiver &&) r, dueTime_, loop_};
+    }
```
lewissbaker: Can we combine these overloads into a single `const&` overload to avoid the duplication?

include/unifex/timed_single_thread_context.hpp
```diff
@@ -134,7 +134,17 @@ namespace _timed_single_thread_context {
...
+    template <typename Receiver>
+    after_operation<Duration, Receiver> connect(Receiver&& receiver) & {
+      return after_operation<Duration, Receiver>{
+          *context_, duration_, (Receiver &&) receiver};
+    }
+    template <typename Receiver>
+    after_operation<Duration, Receiver> connect(Receiver&& receiver) const & {
       return after_operation<Duration, Receiver>{
           *context_, duration_, (Receiver &&) receiver};
     }
```
lewissbaker: Can we combine these overloads into a single `const&` overload to avoid the duplication?

include/unifex/trampoline_scheduler.hpp
```diff
@@ -123,6 +123,14 @@ class scheduler {
...
       return operation<Receiver>{(Receiver &&) receiver, maxRecursionDepth_};
     }
+    template <typename Receiver>
+    operation<Receiver> connect(Receiver&& receiver) & {
+      return operation<Receiver>{(Receiver &&) receiver, maxRecursionDepth_};
+    }
+    template <typename Receiver>
+    operation<Receiver> connect(Receiver&& receiver) const& {
+      return operation<Receiver>{(Receiver &&) receiver, maxRecursionDepth_};
+    }
```
lewissbaker: Combine into a single `const&` overload?

include/unifex/transform.hpp
```diff
@@ -170,6 +171,7 @@ struct _sender<Predecessor, Func>::type {
 
   template <typename Receiver>
   auto connect(Receiver&& r) &
+      noexcept(is_nothrow_tag_invocable_v<tag_t<unifex::connect>, Predecessor&, receiver<std::remove_cvref_t<Receiver>>>)
```
lewissbaker: Also needs to check that both `Receiver` and `Func` is nothrow-decay-copyable (as a proxy for detecting that `receiver` is nothrow constructible.

Similarly for the other overloads.

include/unifex/transform_stream.hpp
```diff
@@ -27,7 +27,7 @@ namespace _tfx_stream {
     auto operator()(StreamSender&& stream, Func&& func) const {
       return next_adapt_stream(
           (StreamSender &&) stream, [func = (Func &&) func](auto&& sender) mutable {
-            return transform((decltype(sender))sender, std::ref(func));
+            return transform((decltype(sender))sender, (Func &&)func);
```
lewissbaker: The use of `std::ref()` here was intentional to allow the use of stateful function objects.

Also, the function will be called once for each `next()` and so we shouldn't be moving the function into the transform-sender.

include/unifex/when_all.hpp
```diff
@@ -63,8 +63,8 @@ template <
     typename... Rest>
 struct _operation_tuple<Index, Receiver, First, Rest...>::type
   : operation_tuple<Index + 1, Receiver, Rest...> {
-  template <typename Parent>
-  explicit type(Parent& parent, First&& first, Rest&&... rest)
+  template <typename First2, typename Parent>
+  explicit type(Parent& parent, First2&& first, Rest&&... rest)
```
lewissbaker: Why was this change necessary?
Shouldn't the sender be perfectly forwarded through to this constructor?

include/unifex/when_all.hpp
```diff
@@ -194,9 +194,10 @@ struct _op<Receiver, Senders...>::type {
   template<std::size_t Index, typename Operation>
   friend class _element_receiver;
 
-  explicit type(Receiver&& receiver, Senders&&... senders)
+  template<typename... Senders2>
+  explicit type(Receiver&& receiver, Senders2&&... senders)
     : receiver_((Receiver &&) receiver),
-      ops_(*this, (Senders &&) senders...) {}
+      ops_(*this, (Senders2 &&) senders...) {}
```
lewissbaker: The senders should be perfectly forwarded through here so I wouldn't have thought that the Senders2 template parameter would have been necessary.
What problems were you seeing with this?

include/unifex/finally.hpp
```diff
@@ -700,39 +701,40 @@ namespace unifex
...
                 is_connectable<
                   CompletionSender,
                   done_receiver<
                       SourceSender,
                       CompletionSender,
-                      Receiver>>>,
-              int> = 0>
+                      std::remove_cvref_t<Receiver>>>,
+                std::true_type>,
+              int(*)[__LINE__]> = nullptr>
```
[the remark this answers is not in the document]
kirkshoop: yes.. though the result is a much longer symbol :(

include/unifex/when_all.hpp
kirkshoop: I reverted this and it compiles locally. Thanks!

include/unifex/when_all.hpp
kirkshoop: I reverted this and it compiles locally. Thanks!

include/unifex/transform_stream.hpp
kirkshoop: this appears to work now - I have reverted the change

include/unifex/sync_wait.hpp
kirkshoop: && did not look like it was doing anything.

include/unifex/stop_immediately.hpp
kirkshoop: Destructors should only be run once and should be run on a const object.

I do not think that const connect maps directly to repeatable.

include/unifex/stop_immediately.hpp
kirkshoop: I wanted to be explicit that this was removed on purpose. could be a comment..

include/unifex/sequence.hpp
```diff
@@ -288,7 +288,7 @@ namespace unifex
           Successor,
           Receiver>;
 
-      Successor successor_;
+      std::remove_cvref_t<Successor> successor_;
```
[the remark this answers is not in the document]
kirkshoop: this was an incomplete rollback of an intermediate state. removed this and others.

include/unifex/range_stream.hpp
```diff
@@ -63,6 +63,12 @@ struct next_sender {
   operation<Receiver> connect(Receiver&& receiver) && {
     return operation<Receiver>{stream_, (Receiver &&) receiver};
   }
+  template <typename Receiver>
+  operation<Receiver> connect(Receiver&& receiver) & {
+    return operation<Receiver>{stream_, (Receiver &&) receiver};
+  }
+  template <typename Receiver>
+  void connect(Receiver&& receiver) const& = delete;
```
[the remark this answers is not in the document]
kirkshoop: I marked the lvalue connect delete as well.

include/unifex/stop_immediately.hpp
lewissbaker: > Destructors should only be run once and should be run on a const object.

Are you equating the `cleanup()` operation to a destructor here?

I guess for ordinary types, you can call the destructor on a const object, however the object ceases to be a const object inside the destructor (I still want to be able to mutate data members inside a destructor). Also, a destructor consumes an object.

So I can perhaps see an argument for allowing the `cleanup()` method to be declared const-rvalue. However, there isn't an easy way to declare that the method was const-rvalue callable but that inside the method the data-members were all mutable without sprinkling const_cast everywhere.

I think the qualification on the `cleanup()` method is a separate concern from the qualification of the `connect()` method on the sender returned from `cleanup()`.

I don't want people to think that it's safe to call `connect()` multiple times on this sender. The current way we indicate that it is safe to call `connect()` multiple times is by adding a const-lvalue overload of `connect()`.

include/unifex/sync_wait.hpp
lewissbaker: It was avoiding a copy of the StopToken and instead capturing a reference to the stop-token passed to sync_wait.

With this change it will still capture a reference to the stop-token in the receiver if the caller passed an lvalue, but will now move-construct the stop-token into the receiver if the caller passed an rvalue.

If we think that we should be capturing a copy of the stop-token in the receiver instead of a reference then perhaps we should be decay-copying the stop-token here instead?

I guess this is all kind of moot after PR #80 is merged anyway.

examples/async_trace.cpp
```diff
@@ -74,26 +80,28 @@ int main() {
 
   auto start = steady_clock::now();
 
-  sync_wait(transform(
-      when_all(
+  sync_wait(
+    transform(
+      defer([&]{
```
lewissbaker: Why was defer() required here?
There shouldn't be a need for repeatability of the operation if just passing to sync_wait() without a repeat/retry of some sort.

include/unifex/async_trace.hpp
```diff
@@ -199,7 +199,11 @@ namespace _async_trace {
     using error_types = Variant<std::exception_ptr>;
 
     template <typename Receiver>
-    operation<Receiver> connect(Receiver&& r) {
+    operation<Receiver> connect(Receiver&& r) && {
+      return operation<Receiver>{(Receiver &&) r};
+    }
```
lewissbaker: Remove this overload? It's doing the same thing as the const& version below.

include/unifex/finally.hpp
```diff
@@ -590,16 +590,17 @@ namespace unifex
         manual_lifetime_union<std::exception_ptr, Errors...>;
 
     public:
-      template <typename CompletionSender2, typename Receiver2>
+      type() = default;
```
lewissbaker: Why was it necessary to declare a default constructor here?
I wouldn't expect it to ever be called.

include/unifex/finally.hpp
```diff
@@ -590,16 +590,17 @@ namespace unifex
...
+      template <typename SourceSender2, typename CompletionSender2, typename Receiver2>
       explicit type(
-          SourceSender&& sourceSender,
+          SourceSender2&& sourceSender,
```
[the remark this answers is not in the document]
lewissbaker: Not fixed yet?

include/unifex/finally.hpp
```diff
@@ -617,19 +618,19 @@ namespace unifex
...
 
     private:
-      UNIFEX_NO_UNIQUE_ADDRESS CompletionSender completionSender_;
+      UNIFEX_NO_UNIQUE_ADDRESS std::remove_cvref_t<CompletionSender> completionSender_;
       UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
       bool started_ = false;
 
       // Result storage.
       union {
         // Storage for error-types that might be produced by ValueSender.
```
lewissbaker: Just noticed that comment should probably be updated to reference SourceSender instead of ValueSender.

include/unifex/finally.hpp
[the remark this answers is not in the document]
lewissbaker: On the other hand, using __LINE__ means that adding a comment anywhere in the file above can change the ABI of this function.

include/unifex/finally.hpp
```diff
@@ -700,39 +701,41 @@ namespace unifex
...
+                      std::remove_cvref_t<Receiver>>>,
                 is_connectable<
                   CompletionSender,
                   done_receiver<
                       SourceSender,
                       CompletionSender,
-                      Receiver>>>,
-              int> = 0>
+                      std::remove_cvref_t<Receiver>>>,
+                std::true_type>,
```
lewissbaker: Why was it necessary to add `std::true_type` here?
I thought std::conjunction would have had the same behaviour without it?

include/unifex/finally.hpp
```diff
@@ -700,39 +701,41 @@ namespace unifex
...
       template <
           typename Receiver,
           typename CPO,
           typename S,
           std::enable_if_t<
               std::conjunction_v<
                 std::is_same<CPO, tag_t<connect>>,
-                std::is_same<S, sender>,
+                std::is_same<std::remove_cvref_t<S>, sender>,
+                std::negation<std::is_lvalue_reference<S>>,
```
lewissbaker: Was it necessary to limit to rvalue-reference only here so that the subsequent `is_connectable<SourceSender/CompletionSender` checks are using the correct value-category?

ie. they'd need to be checking `is_connectable<const SourceSender&, ...>` and also `is_copy_constructible<CompletionSender>` and `is_connectable<CompletionSender, ...>`.

Do we still need to add this other const-lvalue connect overload?

include/unifex/just.hpp
```diff
@@ -72,26 +72,38 @@ class _sender<Values...>::type {
...
 
-  template <typename Receiver>
+  template <
+    typename Value2,
+    std::enable_if_t<
+      (sizeof...(Values) == 1 &&
+      !std::is_same_v<Value2, type>), int> = 0>
+    explicit type(Value2&& value)
+    noexcept(std::is_nothrow_constructible_v<std::tuple<Values...>, Value2>)
+    : values_((Value2 &&) value) {}
```
lewissbaker: I wonder whether this could all be simplified by adding an extra std::in_place_t arg in front of the parameter pack to disambiguate the in-place constructors from the copy/move constructors?

e.g.
```
template<typename... Values2>
explicit type(std::in_place_t, Values2&&... values)
  noexcept(std::is_nothrow_constructible_v<std::tuple<Values...>, Values2...>)
  : values_((Values2&&)values...)
{}
```

Then just update the `just()` implementation to pass `std::in_place` as the first arg to the constructor.

include/unifex/manual_event_loop.hpp
```diff
@@ -93,6 +93,10 @@ class context {
       operation<Receiver> connect(Receiver&& receiver) && {
         return operation<Receiver>{(Receiver &&) receiver, loop_};
       }
```
lewissbaker: Can we remove the rvalue overload now?

include/unifex/ready_done_sender.hpp
```diff
@@ -55,7 +55,11 @@ namespace _ready_done {
     }
 
     template <typename Receiver>
-    operation<Receiver> connect(Receiver&& receiver) {
+    operation<Receiver> connect(Receiver&& receiver) && {
+      return operation<Receiver>{(Receiver &&) receiver};
+    }
```
lewissbaker: Can we remove the rvalue overload now?

include/unifex/retry_when.hpp
```diff
@@ -352,28 +352,14 @@ class _sender<Source, Func>::type {
...
     typename Receiver,
     std::enable_if_t<
         std::is_constructible_v<Source, const Source&> &&
         std::is_constructible_v<Func, const Func&> &&
         std::is_constructible_v<std::remove_cvref_t<Receiver>, Receiver> &&
         is_connectable_v<Source&, source_receiver<Source, Func, std::remove_cvref_t<Receiver>>>, int> = 0>
-  operation<Source, Func, Receiver> connect(Receiver&& r) &
+  operation<Source, Func, Receiver> connect(Receiver&& r) const&
       noexcept(std::is_nothrow_constructible_v<
           operation<Source, Func, Receiver>, const Source&, const Func&, Receiver>) {
```
lewissbaker: I wonder if the noexcept clause should be defined in terms of nothrow copyability of Source, Func, Receiver rather than of nothrow constructibility of operation to avoid needing to instantiate the operation class too early.

include/unifex/sequence.hpp
```diff
@@ -298,7 +298,7 @@ namespace unifex
       status status_;
       union {
         manual_lifetime<operation_t<
-            Predecessor,
+            std::remove_cvref_t<Predecessor>,
```
lewissbaker: This should just be `Predecessor` rather than `std::remove_cvref_t<Predecessor>` to match the type we are passing to `connect()` in the constructor.

include/unifex/stop_immediately.hpp
```diff
@@ -413,6 +415,10 @@ struct _stream<SourceStream, Values...>::type {
     operation<Receiver> connect(Receiver&& receiver) && {
       return operation<Receiver>{stream_, (Receiver &&) receiver};
     }
+    template<typename Receiver>
+    operation<Receiver> connect(Receiver&& receiver) const& {
+      return operation<Receiver>{stream_, (Receiver &&) receiver};
+    }
```
lewissbaker: The cleanup_sender should not be const-lvalue connectable.
The cleanup operation has side-effects that are not idempotent and starting it multiple times would have undefined-behaviour.

include/unifex/thread_unsafe_event_loop.hpp
lewissbaker: Can we remove the rvalue overload now?

include/unifex/thread_unsafe_event_loop.hpp
```diff
@@ -225,6 +230,11 @@ namespace _thread_unsafe_event_loop {
       return at_operation<std::remove_cvref_t<Receiver>>{
           (Receiver &&) r, dueTime_, loop_};
     }
+    template <typename Receiver>
+    at_operation<std::remove_cvref_t<Receiver>> connect(Receiver&& r) const& {
```
lewissbaker: Can we remove the rvalue connect() overload now that we have a const-lvalue overload?

include/unifex/timed_single_thread_context.hpp
```diff
@@ -134,7 +134,12 @@ namespace _timed_single_thread_context {
     using error_types = Variant<>;
 
     template <typename Receiver>
-    after_operation<Duration, Receiver> connect(Receiver&& receiver) {
+    after_operation<Duration, Receiver> connect(Receiver&& receiver) && {
+      return after_operation<Duration, Receiver>{
+          *context_, duration_, (Receiver &&) receiver};
+    }
```
lewissbaker: Can we remove the rvalue overload now?

include/unifex/trampoline_scheduler.hpp
```diff
@@ -123,6 +123,10 @@ class scheduler {
     operation<Receiver> connect(Receiver&& receiver) && {
       return operation<Receiver>{(Receiver &&) receiver, maxRecursionDepth_};
     }
```
lewissbaker: Can we remove the rvalue overload now?

include/unifex/transform.hpp
```diff
@@ -161,23 +161,23 @@ struct _sender<Predecessor, Func>::type {
 
   template <typename Receiver>
   auto connect(Receiver&& r) &&
+      noexcept(
+        std::is_nothrow_constructible_v<std::remove_cvref_t<Receiver>, Receiver> && 
+        std::is_nothrow_copy_constructible_v<Func> &&
+        is_nothrow_tag_invocable_v<tag_t<unifex::connect>, Predecessor, receiver<std::remove_cvref_t<Receiver>>>)
```
lewissbaker: Simplify using `is_nothrow_connectable_v`

include/unifex/transform.hpp
```diff
@@ -161,23 +161,23 @@ struct _sender<Predecessor, Func>::type {
 
   template <typename Receiver>
   auto connect(Receiver&& r) &&
+      noexcept(
+        std::is_nothrow_constructible_v<std::remove_cvref_t<Receiver>, Receiver> && 
+        std::is_nothrow_copy_constructible_v<Func> &&
```
lewissbaker: Should be `std::is_nothrow_move_constructible_v<Func>`

include/unifex/transform.hpp
```diff
@@ -161,23 +161,23 @@ struct _sender<Predecessor, Func>::type {
...
-        pred_,
-        receiver<std::remove_cvref_t<Receiver>>{func_, std::forward<Receiver>(r)});
-  }
-
   template <typename Receiver>
   auto connect(Receiver&& r) const &
+      noexcept(
+        std::is_nothrow_constructible_v<std::remove_cvref_t<Receiver>, Receiver> && 
+        std::is_nothrow_copy_constructible_v<Func> &&
+        is_nothrow_tag_invocable_v<tag_t<unifex::connect>, const Predecessor&, receiver<std::remove_cvref_t<Receiver>>>)
```
lewissbaker: Can be simplified using `is_nothrow_connectable_v`

include/unifex/finally.hpp
kirkshoop: at one point the variadic constructor was eating the default constructor call as well. I am using the std::in_place suggestion now.

examples/async_trace.cpp
kirkshoop: I narrowed down the scope at which to put this `defer()` to get rid of a compile error.

include/unifex/finally.hpp
kirkshoop: It allowed me to comment out full lines easily. removed.

include/unifex/finally.hpp
kirkshoop: I cannot specify an rvalue overload of tag_invoke. it always creates a forwarding overload of tag_invoke when using &&. I just decided to use the same forwarding overload to handle both rvalue and const lvalue.

lewissbaker: `std::is_lvalue_reference<S>` will still evaluate to true if `S` is a const-ref.
So this constraint would seem to only match rvalue-references as it stands.

Consider changing to `std::negation<std::is_same<S, sender&>>` if you want it to match `const sender&` and `sender&&`.
But if you do that then you also need to change the `is_connectable` check below to use the corresponding value category for the SourceSender check.

include/unifex/finally.hpp
```diff
@@ -700,39 +700,40 @@ namespace unifex
...
       template <
           typename Receiver,
           typename CPO,
           typename S,
           std::enable_if_t<
               std::conjunction_v<
                 std::is_same<CPO, tag_t<connect>>,
-                std::is_same<S, sender>,
+                std::is_same<std::remove_cvref_t<S>, sender>,
+                std::negation<std::is_same<S, sender&>>,
```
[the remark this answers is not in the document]
kirkshoop: for now I omitted the constraint

include/unifex/finally.hpp
```diff
@@ -713,7 +713,7 @@ namespace unifex
               std::conjunction_v<
                 std::is_same<CPO, tag_t<connect>>,
                 std::is_same<std::remove_cvref_t<S>, sender>,
-                std::negation<std::is_lvalue_reference<S>>,
+                std::negation<std::is_same<S, sender&>>,
                 is_connectable<
                   SourceSender,
```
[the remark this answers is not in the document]
kirkshoop: I used this to recover the value category for SourceSender
  `typename SourceSenderType = decltype(std::declval<S>().source_),`

include/unifex/retry_when.hpp
```diff
@@ -346,36 +346,26 @@ class _sender<Source, Func>::type {
...
     return operation<Source, Func, Receiver>{
         (Source&&)source_, (Func&&)func_, (Receiver&&)r};
   }
 
   template<
     typename Receiver,
-    typename Op = operation<Source, Func, Receiver>,
     std::enable_if_t<
         std::is_constructible_v<Source, Source&> &&
         std::is_constructible_v<Func, Func&> &&
```
lewissbaker, line 361: The `Source` and `Func` constraints should be using `is_copy_constructible_v`.

include/unifex/retry_when.hpp
```diff
@@ -346,36 +346,26 @@ class _sender<Source, Func>::type {
         std::is_constructible_v<std::remove_cvref_t<Receiver>, Receiver> &&
         is_connectable_v<Source&, source_receiver<Source, Func, std::remove_cvref_t<Receiver>>>, int> = 0>
   operation<Source, Func, Receiver> connect(Receiver&& r) &&
-      noexcept(std::is_nothrow_constructible_v<
-          operation<Source, Func, Receiver>, Source, Func, Receiver>) {
+      noexcept(
+        std::is_nothrow_move_constructible_v<Source> &&
+        std::is_nothrow_move_constructible_v<Func> &&
+        std::is_nothrow_constructible_v<std::remove_cvref_t<Receiver>, Receiver>) {
```
lewissbaker, line 352: Needs to also check `is_nothrow_connectable_v<Source&, source_receiver<Source, Func, std::remove_cvref_t<Receiver>>>` in the `noexcept` clause.

lewissbaker: And also in the `connect() const &`  noexcept clause below.

examples/async_trace.cpp
lewissbaker: I think I found the cause of this issue.

The transform() sender's connect() methods were not SFINAE-friendly and so when trying to call connect() on an rvalue it was still trying to instantiate the const-lvalue-ref connect() overload. Once transform() and when_all()'s connect() methods were modified to SFINAE out if input senders were not connectable then this defer() is no longer required.

See #106

PR #96, opened 2020-03-21
add repeat()

kirkshoop:
repeat() requires a void-value sender that has an lvalue & connect()
it will then repeat the sender every time it completes successfully, until stopped.


include/unifex/repeat.hpp
```diff
@@ -0,0 +1,235 @@
...
+    unifex::set_error(std::move(op_->receiver_), (Error&&)error);
+  }
+
+private:
+  template<typename CPO, typename... Args>
+  friend auto tag_invoke(CPO cpo, const repeat_source_receiver& r, Args&&... args)
+      noexcept(std::is_nothrow_invocable_v<CPO, const Receiver&, Args...>)
+      -> std::invoke_result_t<CPO, const Receiver&, Args...> {
+    return std::move(cpo)(r.get_receiver(), (Args&&)args...);
+  }
```
lewissbaker: This will probably require an `enable_if<is_callable_v<CPO, const Receiver&, Args...>, int> = 0` to be added here to avoid potential for recursive instantiation under msvc.

include/unifex/repeat.hpp
```diff
@@ -0,0 +1,235 @@
...
+  const Receiver& get_receiver() const noexcept {
+    assert(op_ != nullptr);   
+    return op_->receiver_;
+  }
+
+  operation* op_;
+};
+
+template<typename Source, typename Receiver>
+class repeat_operation {
```
lewissbaker: This will also require some adl-isolation treatment.

include/unifex/repeat.hpp
```diff
@@ -0,0 +1,235 @@
...
+#pragma once
+
+#include <unifex/config.hpp>
+#include <unifex/tag_invoke.hpp>
+#include <unifex/receiver_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
+#include <unifex/type_traits.hpp>
+#include <unifex/type_list.hpp>
+#include <unifex/manual_lifetime.hpp>
+#include <unifex/manual_lifetime_union.hpp>
```
lewissbaker: This header is unused.

include/unifex/repeat.hpp
```diff
@@ -0,0 +1,235 @@
...
+
+} // namespace detail
+
+template<typename Source>
+class repeat_sender {
+
+public:
+  template<template<typename...> class Variant,
+           template<typename...> class Tuple>
+  using value_types = typename Source::template value_types<Variant, Tuple>;
```
lewissbaker: This is incorrect - the `repeat_sender` never calls `set_value()` on the receiver.
This should be:
```
template<template<typename...> class Variant,
         template<typename...> class Tuple>
using value_types = Variant<>;
```

include/unifex/repeat.hpp
```diff
@@ -0,0 +1,235 @@
...
+
+  template<template<typename...> class Variant>
+  using error_types = typename Source::template error_types<Variant>;
```
lewissbaker: This also needs to inject the `std::exception_ptr` type to the list of error types as it might send `std::exception_ptr` if the connecting the sender with the `repeat_source_receiver` might throw.

lewissbaker:
This algorithm is the rough equivalent of:
```
task<void> repeat(sender auto src) {
  while (true) {
    co_await src;
  }
}
```

I wonder if this algorithm would be better named `repeat_forever()` to indicate this?

Is there scope for a `repeat_until(src, pred)` that executes some predicate on the value-result of the source and completes with that value if the predicate returns true?
ie. equivalent of
```
while (true) {
  auto result = co_await src;
  if (pred(result))
    co_return std::move(result);
}
```

include/unifex/repeat.hpp
```diff
@@ -162,36 +168,34 @@ class repeat_operation {
   manual_lifetime<source_op_t> sourceOp_;
 };
 
-} // namespace detail
-
 template<typename Source>
-class repeat_sender {
+class sender {
```
ericniebler: This needs ADL isolation, too.

kirkshoop:
I think that the predicate form you propose is closer to Rx [repeat_until](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#repeatUntil-io.reactivex.functions.BooleanSupplier-)

However the Rx version produces a sequence of all the results. This one requires a source that produces void and produces void.

I decided to rename to repeat_effect and add repeat_effect_until. This will keep the names open for the sequence producing one later.

doc/api_reference.md
```diff
@@ -269,6 +271,61 @@ sender will complete with `set_done(r)`.
...
+
+### `repeat_effect(Sender source) -> Sender`
+
+The `repeat_effect()` algorithm repeats the source sender until the source is 
+cancelled.
+
+The `source` sender must be lvalue connectable (ie. can be connected and started 
+multiple times).
+
+The `source` sender must be an effect. It must prooduce void.
```
lewissbaker, line 314: prooduce -> produce

doc/api_reference.md
```diff
@@ -269,6 +271,61 @@ sender will complete with `set_done(r)`.
...
 
+### `repeat_effect_until(Sender source, Invocable predicate) -> Sender`
+
+The `repeat_effect_until()` algorithm repeats the source sender for as long as the 
+predicate returns false.
+
+The `source` sender must be lvalue connectable (ie. can be connected and started 
+multiple times).
+
+The `source` sender must be an effect. It must prooduce void.
```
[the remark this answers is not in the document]
kirkshoop: I was sure that I wrote this down somewhere else. not sure where.

The form that traffics in values needs to be expressed in terms of sequences. If it does not use sequences, then it must mix in reduce() functionality. 

a version of repeat that produced a sequence of the values might be reduced back to a single result in many ways:
`just(42) | repeat() | sum()`
`just(42) | repeat() | last()`
`just(42) | repeat() | first()`

A repeat that produces a single value would be very complex if it allowed all of these to be expressed.

include/unifex/repeat_effect_until.hpp
```diff
@@ -0,0 +1,298 @@
...
+    if constexpr (std::is_nothrow_invocable_v<Predicate>) {
+      // call predicate and complete with void if it returns true
+      if(((Predicate)op->predicate_)()) {
+        return;
+      }
+    } else {
+      try {
+        // call predicate and complete with void if it returns true
+        if(((Predicate)op->predicate_)()) {
+          return;
```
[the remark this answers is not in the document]
kirkshoop: not sure what happened here, thanks!

PR #106, opened 2020-04-12
Make sender connect() implementations SFINAE-friendly.

lewissbaker:
Calling connect() on a transform/when_all sender was attempting to
instantiate both rvalue-ref and const-lvalue-ref overloads of connect().
If the predecessor was only rvalue-connectable then
this would result in a compile-error.

This change modifies the connect() methods to SFINAE out if that particular connect() overload is not supported.

This fixes compilation errors in io_uring_test that were
introduced in #95 which added some const-lvalue-ref
overloads of connect().


include/unifex/when_all.hpp
```diff
@@ -292,19 +304,39 @@ class _sender<Senders...>::type {
...
-  operation<Receiver, Senders...> connect(Receiver&& receiver) && {
+  template <
+    typename CPO,
+    typename Sender,
+    typename Receiver,
+    std::enable_if_t<
+      std::conjunction_v<
+        std::is_same<CPO, tag_t<unifex::connect>>,
+        std::is_same<Sender, type>,
+        when_all_connectable<std::remove_cvref_t<Receiver>, Senders...>>, int> = 0>
```
ericniebler, line 315: It would be nice to eliminate the `conjunction` here as well. Is short-circuiting needed here?

include/unifex/when_all.hpp
```diff
@@ -292,19 +304,39 @@ class _sender<Senders...>::type {
...
+        std::is_same<std::remove_cvref_t<Sender>, type>,
+        std::negation<std::is_same<Sender, type>>,
+        when_all_connectable<std::remove_cvref_t<Receiver>, const Senders&...>>, int> = 0>
+  friend auto tag_invoke(CPO cpo, Sender&& sender, Receiver&& receiver)
+    -> operation<Receiver, const Senders&...> {
     return std::apply([&](const Senders&... senders) {
       return operation<Receiver, const Senders&...>{
           (Receiver &&) receiver, senders...};
-    }, senders_);
+    }, sender.senders_);
```
ericniebler, line 339: Can you find a way to express this as a single `tag_invoke` overload?

include/unifex/let.hpp
```diff
@@ -317,20 +317,22 @@ class _sender<Predecessor, SuccessorFactory>::type {
...
-      -> operation<const Predecessor&, SuccessorFactory, Receiver> {
-    return operation<const Predecessor&, SuccessorFactory, Receiver>{
-      pred_, func_, (Receiver &&) receiver};
+  template <
+    typename CPO,
+    typename Sender,
+    typename Receiver,
+    std::enable_if_t<std::conjunction_v<
+      std::is_same<CPO, tag_t<unifex::connect>>,
+      std::is_same<std::remove_cvref_t<Sender>, type>>, int> = 0>
```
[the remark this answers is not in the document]
lewissbaker: Should we be adding our own `unifex::is_same_v` trait that maps to these intrinsics?
Do you know why libc++ doesn't already map to those intrinsics on compilers it knows about?

include/unifex/when_all.hpp
lewissbaker: We want to avoid evaluating the when_all_connectable<> if the first two don't match, but we can probably evaluate the first two together.

include/unifex/sequence.hpp
```diff
@@ -369,40 +369,37 @@ namespace unifex
...
+            is_connectable<const Predecessor&, predecessor_receiver<const Predecessor&, Successor, Receiver>>,
+            is_connectable<Successor, successor_receiver<const Predecessor&, Successor, Receiver>>>, int> = 0>
+      friend auto tag_invoke(tag_t<unifex::connect>, Sender&& sender, Receiver&& receiver)
           -> operation<const Predecessor&, Successor, Receiver> {
-        return operation<const Predecessor&, Successor, Receiver>{
-            predecessor_, successor_, (Receiver &&) receiver};
+        return operation<const Predecessor&, Successor,  Receiver>{
+            sender.predecessor_,
+            sender.successor_,
+            (Receiver &&) receiver};
```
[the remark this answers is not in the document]
lewissbaker: It probably could be merged.

The challenge I've had is trying to describe the types to pass to `operation<...>` and the `is_constructible` traits such that their value categories match the value category of the `Sender` type.

I could possibly use `decltype((std::declval<Sender>().predecessor_))` but I think we ran into access-control issues with private member access in template friend function default template args, although I could possibly use the workaround that uses `s.*&type::predecessor_`.

PR #107, opened 2020-04-15
fold together tag_invoke(connect) overloads; unroll std::conjunction

ericniebler:


include/unifex/linux/io_epoll_context.hpp
```diff
@@ -433,8 +433,8 @@ class io_epoll_context::schedule_at_sender {
       : context_(context), dueTime_(dueTime) {}
 
   template <typename Receiver>
-  operation<std::remove_cvref_t<Receiver>> connect(Receiver&& r) {
-    return operation<std::remove_cvref_t<Receiver>>{
+  operation<remove_cvref_t<Receiver>> connect(Receiver&& r) {
```
lewissbaker, line 436: This should be `const &` qualified.

include/unifex/linux/io_uring_context.hpp
```diff
@@ -493,8 +493,8 @@ class io_uring_context::read_sender {
       : context_(context), fd_(fd), offset_(offset), buffer_(buffer) {}
 
   template <typename Receiver>
-  operation<std::remove_cvref_t<Receiver>> connect(Receiver&& r) {
-    return operation<std::remove_cvref_t<Receiver>>{*this, (Receiver &&) r};
+  operation<remove_cvref_t<Receiver>> connect(Receiver&& r) {
```
lewissbaker, line 496: This should be `&&` qualified

include/unifex/tag_invoke.hpp
```diff
@@ -16,8 +16,7 @@
 #pragma once
 
 #include <unifex/config.hpp>
-
-#include <type_traits>
```
lewissbaker, line 20: Still using `std::bool_constant` in this file.
Should this #include remain?

include/unifex/this.hpp
```diff
@@ -16,8 +16,7 @@
 #pragma once
 
 #include <unifex/config.hpp>
-
-#include <type_traits>
```
lewissbaker, line 20: Similarly here, still using `std::` traits below.
Should this #include remain? Or are you assuming that `unifex/type_traits.hpp` will transitively include it?

include/unifex/type_traits.hpp
```diff
@@ -51,18 +82,36 @@ using instance_of = std::bool_constant<instance_of_v<T, X>>;
...
 template <typename T>
-using wrap_reference_t = std::conditional_t<
+using wrap_reference_t = conditional_t<
     std::is_reference_v<T>,
     std::reference_wrapper<std::remove_reference_t<T>>,
     T>;
 
+template <typename Class, typename Member>
+using member_t =
+    conditional_t<std::is_lvalue_reference_v<Class>, const Member&, Member>;
```
lewissbaker, line 110: Could we avoid instantiating `conditional_t` and `is_lvalue_reference_v` here through something like the following?
```
template<typename Class, typename Self, typename Member>
using member_t = decltype(((((Self(*)())nullptr)()).*((Member Class::*(*)())nullptr)()) );
```

And usage like: `member_t<type, Self, Value>`

include/unifex/when_all.hpp
```diff
@@ -272,14 +272,18 @@ struct _sender {
   class type;
 };
 template <typename... Senders>
-using sender = typename _sender<std::remove_cvref_t<Senders>...>::type;
+using sender = typename _sender<remove_cvref_t<Senders>...>::type;
 
 template<typename Receiver, typename Indices, typename... Senders>
-inline constexpr bool when_all_connectable_v = false;
+extern const bool _when_all_connectable_v;
```
lewissbaker, line 278: Huh. I didn't know you could do that!

include/unifex/tag_invoke.hpp
ericniebler: `<type_traits>` is included by `<unifex/type_traits.hpp>`. That was my thinking anyway.

include/unifex/this.hpp
ericniebler: That.

PR #109, opened 2020-04-18
Define some concept portability macros and replace enable_if with requires clauses

ericniebler:
This is a minimal change that doesn't add any additional checks. It should not effect compile times. It just uses `requires` clauses instead of `enable_if` when the compiler supports concepts.

Let's get at least this much in, and then I can proceed in future PRs to add additional checks, being careful not to regress compile times by too much.

The syntax for a constrained function template is:

```c++
template( class A, class B )
  (requires Concept<A> AND Concept<B>)
void foo( A a, B b )
{ ... }
```

In emulation mode, the clauses separated by the `AND` are placed in separate `enable_if`s, preserving short-circuiting.

### DIFF
--- a/include/unifex/dematerialize.hpp
+++ b/include/unifex/dematerialize.hpp
@@ -22,8 +22,11 @@
 #include <unifex/tag_invoke.hpp>
 #include <unifex/type_list.hpp>
 #include <unifex/type_traits.hpp>
+#include <unifex/std_concepts.hpp>
 
 #include <type_traits>
+
+#include <unifex/detail/prologue.hpp>
 
 namespace unifex {
 namespace _demat {
@@ -37,19 +40,14 @@ namespace _demat {
   template <typename Receiver>
   class _receiver<Receiver>::type {
    public:
-    template <
-      typename Receiver2,
-      std::enable_if_t<std::is_constructible_v<Receiver, Receiver2>, int> = 0>
+    template(typename Receiver2)
+      (requires constructible_from<Receiver, Receiver2>)
     explicit type(Receiver2&& receiver) noexcept(
         std::is_nothrow_constructible_v<Receiver, Receiver2>)
       : receiver_(static_cast<Receiver2&&>(receiver)) {}
 
-    template <
-        typename CPO,
-        typename... Values,
-        std::enable_if_t<
-            is_receiver_cpo_v<CPO> && is_callable_v<CPO, Receiver, Values...>,
-            int> = 0>
+    template(typename CPO, typename... Values)
+        (requires is_receiver_cpo_v<CPO> AND is_callable_v<CPO, Receiver, Values...>)
     void set_value(CPO cpo, Values&&... values) && noexcept(
         is_nothrow_callable_v<CPO, Receiver, Values...>) {
       static_cast<CPO&&>(cpo)(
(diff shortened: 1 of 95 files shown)


ericniebler:
@lewissbaker, @kirkshoop this is ready for its close-up.

include/unifex/dematerialize.hpp
```diff
@@ -198,7 +181,7 @@ namespace _demat {
 namespace _demat_cpo {
   inline constexpr struct _fn {
    private:
-    template<bool>
+    template <bool>
```
lewissbaker, line 184: Is the space here significant?

include/unifex/detail/concept_macros.hpp
```diff
@@ -0,0 +1,347 @@
...
+    (1u==sizeof(NAME ## UNIFEX_CONCEPT_FRAGMENT_( \
+      static_cast<::unifex::_concept::tag<__VA_ARGS__> *>(nullptr), nullptr)))
+
+#endif
+
+////////////////////////////////////////////////////////////////////////////////
+// UNIFEX_TEMPLATE
+// Usage:
+//   UNIFEX_TEMPLATE(typename A, typename B)
+//     (requires Concept1<A> && Concept2<B>)
```
lewissbaker, line 295: Is this supposed to be `(requires Concept1<A> UNIFEX_AND Concept2<B>)`?

include/unifex/dematerialize.hpp
ericniebler: I changed it to be consistent with the formatting elsewhere. For the purposes of reviewing the PR, you can change the settings to hide whitespace changes.

include/unifex/detail/concept_macros.hpp
ericniebler: Yes.

include/unifex/std_concepts.hpp
```diff
@@ -0,0 +1,324 @@
...
+    #define UNIFEX_DIAGNOSTIC_IGNORE_FLOAT_EQUAL UNIFEX_DIAGNOSTIC_IGNORE("-Wfloat-equal")
+    #define UNIFEX_DIAGNOSTIC_IGNORE_CPP2A_COMPAT UNIFEX_DIAGNOSTIC_IGNORE("-Wc++2a-compat")
+  #else
+    #define UNIFEX_DIAGNOSTIC_PUSH
+    #define UNIFEX_DIAGNOSTIC_POP
+    #define UNIFEX_DIAGNOSTIC_IGNORE_INIT_LIST_LIFETIME
+    #define UNIFEX_DIAGNOSTIC_IGNORE_FLOAT_EQUAL
+    #define UNIFEX_DIAGNOSTIC_IGNORE_CPP2A_COMPAT
+  #endif
+#endif // MSVC/Generic configuration switch
```
lewissbaker, line 53: This doesn't seem particularly specific to std_concepts being defined here.
Should it be moved to `config.h.in`?

include/unifex/std_concepts.hpp
```diff
@@ -0,0 +1,324 @@
...
+  template <typename T>
+  UNIFEX_CONCEPT
+    copy_constructible =
+      move_constructible<T> &&
+      UNIFEX_FRAGMENT(unifex::_copy_constructible, T);
+
+  template <typename T>
+  UNIFEX_CONCEPT_FRAGMENT(
+    _move_assignable,
+      requires()(0) &&
```
lewissbaker, line 281: What does this syntax mean?

include/unifex/swap.hpp
```diff
@@ -0,0 +1,115 @@
...
+UNIFEX_INLINE_VAR constexpr _swap::_fn swap {};
+} // namespace _swap_cpo
+using _swap_cpo::swap;
+
+template <typename, typename, typename>
+UNIFEX_INLINE_VAR constexpr bool is_swappable_with_v = false;
+
+template <typename T, typename U>
+UNIFEX_INLINE_VAR constexpr bool is_swappable_with_v<
+    T, U, decltype(swap(static_cast<T&&(*)()noexcept>(nullptr)(),
```
lewissbaker, line 101: Maybe define this function-pointer-pattern as a macro in `type_traits.hpp`?
eg.
```
#define UNIFEX_DECLVAL(T) static_cast<T(*)()noexcept>(nullptr)()
```
then usage here (and everywhere else it's used) could be:
```
template <typename T, typename U>
UNIFEX_INLINE_VAR constexpr bool is_swappable_with_v<
    T, U, decltype(swap(UNIFEX_DECLVAL(T&&), UNIFEX_DECLVAL(U&&)))> = true;
```

include/unifex/via.hpp
```diff
@@ -23,12 +23,15 @@
 #include <unifex/submit.hpp>
 #include <unifex/tag_invoke.hpp>
 #include <unifex/get_stop_token.hpp>
+#include <unifex/std_concepts.hpp>
```
lewissbaker, line 26: Not actually adding any std_concepts usage here?

include/unifex/std_concepts.hpp
ericniebler: It's the same as `requires() { 0; }`. It is an empty requirement. The way `UNIFEX_CONCEPT_FRAGMENT` is currently defined, a `requires` clause is required and must be first. I could make a `requires` clause optional at the expense of a huge pile of PP gunk. It wouldn't be hard, it's just unpleasant.

include/unifex/swap.hpp
ericniebler: I thought you guys wanted fewer macros. :-)

include/unifex/detail/concept_macros.hpp
```diff
@@ -0,0 +1,347 @@
...
+  UNIFEX_PP_CAT3(UNIFEX_PP_EAT_TYPENAME_SELECT_, \
+    UNIFEX_PP_EVAL(UNIFEX_PP_CHECK, UNIFEX_PP_CAT3( \
+      UNIFEX_PP_EAT_TYPENAME_PROBE_, \
+      X)))
+#define UNIFEX_PP_EAT_TYPENAME_(...) \
+  UNIFEX_PP_EVAL2(UNIFEX_PP_EAT_TYPENAME_SELECT_, __VA_ARGS__,)(__VA_ARGS__)
+#define UNIFEX_PP_EAT_TYPENAME_SELECT_0(...) __VA_ARGS__
+#define UNIFEX_PP_EAT_TYPENAME_SELECT_1(...) \
+  UNIFEX_PP_CAT3(UNIFEX_PP_EAT_TYPENAME_, __VA_ARGS__)
+#define UNIFEX_PP_EAT_TYPENAME_typename
```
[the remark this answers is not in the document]
ericniebler: It is more complicated then strictly necessary to accommodate MSVC's broken preprocessor. :-/

PR #110, opened 2020-04-20
Add receiver and receiver_of concepts; use them in constraints

ericniebler:


ericniebler:
This PR regresses build times by about 5-10%. I'll investigate to see if I can do better.

ericniebler:
@lewissbaker, @kirkshoop this is ready for its close-up.

include/unifex/receiver_concepts.hpp
```diff
@@ -195,6 +198,41 @@ constexpr bool is_receiver_cpo_v = is_one_of_v<
 template <typename T>
 using is_receiver_cpo = std::bool_constant<is_receiver_cpo_v<T>>;
 
+template <typename R, typename E>
+UNIFEX_CONCEPT_FRAGMENT(
+  _receiver,
+    requires(remove_cvref_t<R>&& r, E&& e) //
+    (
+      noexcept(set_done(std::move(r))),
```
lewissbaker, line 206: How does this ensure the `noexcept()` expression evaluates to `true` and not just that this is a valid expression?

include/unifex/retry_when.hpp
```diff
@@ -97,7 +97,7 @@ class _trigger_receiver<Source, Func, Receiver, Trigger>::type {
   }
 
   template(typename R = Receiver)
-    (requires is_callable_v<decltype(unifex::set_done), Receiver>)
+    (requires receiver<Receiver>)
```
lewissbaker, line 100: I think this should be using `R` instead of `Receiver` to force lazy evaluation of the `receiver` constraint.

include/unifex/receiver_concepts.hpp
ericniebler: This is a little preprocessor-based DSL. Within a `requires` clause in a `UNIFEX_CONCEPT_FRAGMENT`, `noexcept(expr)` has the same effect as `{ expr } noexcept;` in C++20. (Except for the fact that gcc has a bug in the ABI that prevents this from working in concepts emulation mode. In that case, it only checks that `expr` is well-formed.)

PR #111, opened 2020-04-21
use unifex::remove_cvref_t instead of std::remove_cvref_t

ericniebler:


include/unifex/stop_when.hpp
```diff
@@ -340,22 +340,22 @@ namespace unifex
                   stop_when_source_receiver<
                       Source,
                       Trigger,
-                      std::remove_cvref_t<Receiver>>> &&
+                      remove_cvref_t<Receiver>>> &&
```
lewissbaker, line 343: Missing `#include <unifex/type_traits.hpp>` which defines this?

PR #114, opened 2020-04-24
Handful of fixes for msvc

lewissbaker:
Fix some easy problems that were causing msvc to fail to compile.
There are still more things but this gets a bit closer.

- Fix some access-control issues.
- Fix mismatched class/struct usage.
- Adds `is_receiver_query_cpo_v<CPO>` which is now used to filter receivers
  forwarding queries rather than `!is_receiver_cpo_v<CPO>`. This now explicitly
  excludes the `connect` CPO in an attempt to try to fix some recursive template
  instantiations under msvc.
- Work around lack of msvc support for fold-expressions in `noexcept` clauses.

### DIFF
--- a/include/unifex/receiver_concepts.hpp
+++ b/include/unifex/receiver_concepts.hpp
@@ -26,6 +26,10 @@
 #include <unifex/detail/prologue.hpp>
 
 namespace unifex {
+    namespace _connect {
+        struct _fn;
+    }
+
 namespace _rec_cpo {
   inline constexpr struct _set_value_fn {
   private:
@@ -194,6 +198,18 @@ inline constexpr bool is_receiver_cpo_v = is_one_of_v<
     _rec_cpo::_set_next_fn,
     _rec_cpo::_set_error_fn,
     _rec_cpo::_set_done_fn>;
+
+
+// HACK: Approximation for CPOs that should be forwarded through receivers
+// as query operations.
+template <typename T>
+inline constexpr bool is_receiver_query_cpo_v = !is_one_of_v<
+    remove_cvref_t<T>,
+    _rec_cpo::_set_value_fn,
+    _rec_cpo::_set_next_fn,
+    _rec_cpo::_set_error_fn,
+    _rec_cpo::_set_done_fn,
+    _connect::_fn>;
 
 template <typename T>
 using is_receiver_cpo = std::bool_constant<is_receiver_cpo_v<T>>;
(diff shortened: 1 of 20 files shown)


include/unifex/receiver_concepts.hpp
```diff
@@ -195,6 +199,18 @@ inline constexpr bool is_receiver_cpo_v = is_one_of_v<
...
+// HACK: Approximation for CPOs that should be forwarded through receivers
+// as query operations.
+template <typename T>
+inline constexpr bool is_receiver_query_cpo_v = !is_one_of_v<
+    remove_cvref_t<T>,
+    _rec_cpo::_set_value_fn,
+    _rec_cpo::_set_next_fn,
+    _rec_cpo::_set_error_fn,
+    _rec_cpo::_set_done_fn,
+    _connect::_fn>;
```
ericniebler, line 212: This isn't exactly what I had in mind. ;-)

include/unifex/sequence.hpp
```diff
@@ -89,9 +90,8 @@ namespace unifex
       }
 
       template(typename CPO, typename R)
-          (requires (!is_receiver_cpo_v<CPO>) AND
-            same_as<R, successor_receiver> AND
-            is_callable_v<CPO, const Receiver&>)
```
kirkshoop, line 94: what is the story for is_callable_v being removed?

include/unifex/timed_single_thread_context.hpp
```diff
@@ -249,9 +249,9 @@ class timed_single_thread_context {
   friend cancel_callback;
   friend scheduler;
   template <typename Duration, typename Receiver>
-  friend class _timed_single_thread_context::_after_op;
+  friend struct _timed_single_thread_context::_after_op;
   template <typename Receiver>
-  friend class _timed_single_thread_context::_at_op;
+  friend struct _timed_single_thread_context::_at_op;
```
kirkshoop, line 254: _timed_single_thread_context::scheduler also needs this

include/unifex/receiver_concepts.hpp
lewissbaker: I know. It was a stop-gap to see if that fixed the issue I was seeing.
I'll replace it with something more opt-in and category-like later.

include/unifex/sequence.hpp
lewissbaker: That wasn't supposed to sneak in.

I was trying to see if it was the cause of some recursive type-alias errors I was seeing in msvc.
It should ideally be redundant given the `callable_result_t<>` usage in the return-type.

I can add it back in.

PR #115, opened 2020-04-29
Refactor CPOs for improved diagnostics

ericniebler:
Also:
- Define the receiver concepts without the macros when the compiler supports concepts.
- Unroll some macros so diagnostics are shorter.

### DIFF
--- a/include/unifex/dematerialize.hpp
+++ b/include/unifex/dematerialize.hpp
@@ -177,33 +177,29 @@ namespace _demat {
 
 namespace _demat_cpo {
   inline const struct _fn {
-   private:
-    template <bool>
-    struct _impl {
-      template <typename Sender>
-      auto operator()(Sender&& predecessor) const
-          noexcept(is_nothrow_tag_invocable_v<_fn, Sender>) {
-        return unifex::tag_invoke(_fn{}, (Sender&&) predecessor);
-      }
-    };
-   public:
+  private:
     template <typename Sender>
+    using _result_t =
+      typename conditional_t<
+        tag_invocable<_fn, Sender>,
+        meta_tag_invoke_result<_fn>,
+        meta_quote1<_demat::sender>>::template apply<Sender>;
+  public:
+    template(typename Sender)
+      (requires tag_invocable<_fn, Sender>)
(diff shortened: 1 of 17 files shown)


include/unifex/dematerialize.hpp
```diff
@@ -177,33 +177,29 @@ namespace _demat {
...
+    template(typename Sender)
+      (requires tag_invocable<_fn, Sender>)
     auto operator()(Sender&& predecessor) const
-        noexcept(is_nothrow_callable_v<
-          _impl<is_tag_invocable_v<_fn, Sender>>, Sender>) {
-      return _impl<is_tag_invocable_v<_fn, Sender>>{}((Sender&&) predecessor);
+        noexcept(is_nothrow_tag_invocable_v<_fn, Sender>)
+        -> _result_t<Sender> {
+      return unifex::tag_invoke(_fn{}, (Sender&&) predecessor);
     }
```
kirkshoop, line 194: after looking at a couple of these - 
  result_t and the operator() for tag_invoke look like they can be defined once in a base class and reused, right?

ericniebler: Yes, but I think there is value to having it this way. It is far less surprising for users to see diagnostics pointing to the algorithm they just tried to use, rather than some common base class shared by all the algorithms.

include/unifex/receiver_concepts.hpp
```diff
@@ -211,6 +190,28 @@ inline constexpr bool is_receiver_query_cpo_v = !is_one_of_v<
 template <typename T>
 using is_receiver_cpo = std::bool_constant<is_receiver_cpo_v<T>>;
 
+#if UNIFEX_CPP_CONCEPTS
```
kirkshoop, line 193: is this supposed to be UNIFEX_CXX_CONCEPTS?

include/unifex/dematerialize.hpp
kirkshoop: I assumed the base would have the algo context in the template params

include/unifex/receiver_concepts.hpp
ericniebler: Yep.

PR #118, opened 2020-05-01
Add C++17 support to unifex, and add C++17 tests to the CI matrix

ericniebler:
This comments out a PMR-related test in test/submit_allocator_customization_test.cpp that was segfaulting in CI for both gcc and clang, but only for C++20, and only on Linux. I cannot reproduce the crash locally.

I notice that googletest is getting built with -std=c++2a and unifex is getting built with -std=gnu++2a. I _suppose_ that could make a difference, but it doesn't seem to on my machine.

### DIFF
--- a/.github/workflows/libunifex-ci.yml
+++ b/.github/workflows/libunifex-ci.yml
@@ -32,7 +32,7 @@ jobs:
             os: ubuntu-latest,
             build_type: Debug,
             cc: "gcc-9", cxx: "g++-9",
-            cmake_args: "-D CMAKE_CXX_STANDARD:STRING=20"
+            cmake_args: "-D CMAKE_CXX_STANDARD:STRING=20 -D CMAKE_CXX_FLAGS:STRING=\"-fsanitize=address -fno-omit-frame-pointer\""
           }
         - {
             name: "Linux GCC 9 Optimised (C++20)", artifact: "Linux.tar.xz",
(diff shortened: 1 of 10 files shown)


.github/workflows/libunifex-ci.yml
```diff
@@ -16,66 +16,140 @@ jobs:
       matrix:
         config:
         - {
```
lewissbaker, line 18: We have a lot of build variants now! This is great :)

cmake/FindCoroutines.cmake
```diff
@@ -108,12 +108,12 @@ cmake_push_check_state()
 
 set(CMAKE_REQUIRED_QUIET ${Coroutines_FIND_QUIETLY})
 
-# All of our tests required C++20 or later
+# All of our tests required C++17 or later
 if(("x${CMAKE_CXX_COMPILER_ID}" MATCHES "x.*Clang" AND "x${CMAKE_CXX_SIMULATE_ID}" STREQUAL "xMSVC") OR "x${CMAKE_CXX_COMPILER_ID}" STREQUAL "xMSVC")
-    set(_CXX_COROUTINES_STD20 "/std:latest")
+    set(_CXX_COROUTINES_STD17 "/std:c++17")
```
lewissbaker, line 113: How does this flag end up being overridden by the -DCMAKE_CXX_STANDARD flag?
Is that some CMake magic behaviour?

lewissbaker: Ah, never mind. I see it's just checking whether it _can_ use coroutines at all.
This flag isn't actually being passed to libunifex sources.

lewissbaker:
@ericniebler Have you confirmed that at least some of the C++20 build variants are now building with C++ concepts feature enabled?

ericniebler:
> @ericniebler Have you confirmed that at least some of the C++20 build variants are now building with C++ concepts feature enabled?

I have not, and I'm pretty sure that none of the builds currently are using real concepts. I can probably do that without adding to the matrix by, say, passing `-fconcepts` for the debug gcc c++20 builds.

ericniebler:
> I can probably do that without adding to the matrix by, say, passing -fconcepts for the debug gcc c++20 builds.

Actually, let me do that in a separate PR. It turns out gcc-9 concepts has a few quirks that I need workarounds for. (For the record, I have been testing the concepts stuff locally with gcc-trunk and clang-trunk.)

PR #124, opened 2020-05-03
Forward progress delegation

LeeHowes:
Adds P1898R1 style forward progress delegation using example executors that try to reserve capacity, if there is capacity submit to an internal thread context, if not delegate downstream to the scheduler provided by the receiver.

Adds sync_wait_with_context to wrap a threading context to achieve this rather than making sync_wait always provide a scheduler, to avoid breaking other tests. P1898R1 would require that sync_wait provide a scheduler that runs work on the calling thread.

### DIFF
--- a/examples/fp_delegation.cpp
+++ b/examples/fp_delegation.cpp
@@ -127,24 +127,21 @@ delegating_scheduler delegating_context::get_scheduler() noexcept {
 }
 } // namespace
 
-int main() {
+// sync_wait_with_context simulates P1897R3's sync_wait by always providing a
+// context that will run the work, although for simplicity of reuse here it is
+// a thread rather than inline on the caller
+template<class Sender>
+auto sync_wait_with_context(Sender&& s) {
   timed_single_thread_context ctx;
+  return with_query_value((Sender&&)s, get_scheduler, ctx.get_scheduler());
+}
+
+int main() {
   delegating_context inner_delegating_ctx{2};
   delegating_context outer_delegating_ctx{3};
 
-  // Check that the schedule() operation can pick up the current
-  // scheduler from the receiver which we inject by using 'with_query_value()'.
-  sync_wait(with_query_value(schedule(), get_scheduler,
-                             ctx.get_scheduler()));
-
-  // Check that the schedule_after(d) operation can pick up the current
-  // scheduler from the receiver.
-  sync_wait(with_query_value(
-      schedule_after(200ms), get_scheduler, ctx.get_scheduler()));
-
-  // Check that this can propagate through multiple levels of
-  // composed operations.
-  sync_wait(with_query_value(
+  // Try inner context, then outer context, delegating to ctx if necessary
+  sync_wait_with_context(
       transform(
           for_each(via_stream(outer_delegating_ctx.get_scheduler(),
                               transform_stream(via_stream(inner_delegating_ctx.get_scheduler(),
@@ -156,8 +153,7 @@ int main() {
                                                  return value * value;
                                                })),
                    [](int value) { std::printf("got %i\n", value); }),
-          []() { std::printf("done\n"); }),
-      get_scheduler, ctx.get_scheduler()));
+          []() { std::printf("done\n"); }));
 
   std::printf("inner_delegating_ctx operations: %d\n", inner_delegating_ctx.get_count());
   std::printf("outer_delegating_ctx operations: %d\n", outer_delegating_ctx.get_count());
--- a/examples/fp_delegation.cpp
+++ b/examples/fp_delegation.cpp
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <unifex/for_each.hpp>
+#include <unifex/range_stream.hpp>
+#include <unifex/scheduler_concepts.hpp>
+#include <unifex/sync_wait.hpp>
+#include <unifex/timed_single_thread_context.hpp>
+#include <unifex/transform_stream.hpp>
+#include <unifex/via_stream.hpp>
+#include <unifex/with_query_value.hpp>
+
+#include <chrono>
+
+using namespace unifex;
+using namespace std::chrono_literals;
+
+namespace {
+template <typename Receiver>
+struct _delegating_op {
+  class type;
+};
+
+template <typename OperationState>
+class delegating_operation final {
+  public:
+  inline void start() noexcept {
+    std::printf("start()\n");
+    target_op_.start();
+  }
+
+  OperationState target_op_;
+};
+
+class delegating_sender {
+
+  public:
+  template <
+      template <typename...> class Variant,
+      template <typename...> class Tuple>
+  using value_types = Variant<Tuple<>>;
+
+  template <template <typename...> class Variant>
+  using error_types = Variant<>;
+
+  template <typename Receiver>
+  auto connect(Receiver&& receiver) {
+    auto target_scheduler = unifex::get_scheduler(std::as_const(receiver));
+    auto target_op = unifex::connect(unifex::schedule(target_scheduler), (Receiver&&)receiver);
+    std::printf("connect\n");
+    return delegating_operation<
+      remove_cvref_t<decltype(target_op)>>{
+        std::move(target_op)};
+  }
+};
+
+class delegating_scheduler {
+  public:
+  auto schedule() const noexcept {
+    return delegating_sender{};
+  }
+};
+} // namespace
+
+int main() {
+  timed_single_thread_context ctx;
+
+  // Check that the schedule() operation can pick up the current
+  // scheduler from the receiver which we inject by using 'with_query_value()'.
+  sync_wait(with_query_value(schedule(), get_scheduler,
+                             ctx.get_scheduler()));
+
+  // Check that the schedule_after(d) operation can pick up the current
+  // scheduler from the receiver.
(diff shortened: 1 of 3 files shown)


examples/fp_delegation.cpp
```diff
@@ -0,0 +1,178 @@
...
+      context_->run();
+      std::get<LocalOperationState>(op_).start();;
+    }
+  }
+
+  std::variant<DelegatedOperationState, LocalOperationState> op_;
+  delegating_context* context_ = nullptr;
+};
+
+class delegating_sender {
```
lewissbaker, line 91: This kind of variant-sender could probably be factored out into a separate `variant_sender` helper class that implements the more general "this could be one of several senders" patterns.

include/unifex/sync_wait.hpp
```diff
@@ -125,27 +105,18 @@ using receiver_t = typename _receiver<T>::type;
...
-  const bool completesSynchronously =
-      blockingResult == blocking_kind::always ||
-      blockingResult == blocking_kind::always_inline;
-
   using promise_t = _sync_wait::promise<Result>;
   promise_t promise;
-  if (!completesSynchronously) {
-    promise.doneEvent_.emplace();
-  }
+  manual_event_loop ctx;
```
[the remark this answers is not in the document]
LeeHowes: Yes, that occurred to me.

examples/fp_delegation.cpp
```diff
@@ -0,0 +1,178 @@
...
+  int reservation_count_{0};
+  std::atomic<int> run_count_{0};
+  int capacity_;
+  timed_single_thread_context single_thread_context_;
+};
+
+template <typename DelegatedOperationState, typename LocalOperationState>
+class delegating_operation final {
+  public:
+  delegating_operation(DelegatedOperationState&& op, delegating_context* context) : op_{std::move(op)}, context_{context} {}
```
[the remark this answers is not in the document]
LeeHowes: Yes, I mentioned that this was a TODO in our messenger exchange. Low priority while I try to understand why tag_invoke is failing to build.

include/unifex/sync_wait.hpp
```diff
@@ -125,27 +105,18 @@ using receiver_t = typename _receiver<T>::type;
...
 
   // Store state for the operation on the stack.
   auto operation = connect(
-      ((Sender &&) sender),
-      _sync_wait::receiver_t<Result>{promise});
+      with_query_value((Sender&&)sender, get_scheduler, ctx.get_scheduler()),
```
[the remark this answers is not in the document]
LeeHowes: Indeed, although this seemed marginally easier to read as a way to get the customisation in there.

I don't really understand why this approach would cause tag_invoke to be ambiguous, though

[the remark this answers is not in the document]
LeeHowes: As it turns out, that does fix the ambiguity... hmm.

include/unifex/manual_event_loop.hpp
```diff
@@ -18,6 +18,7 @@
 #include <unifex/blocking.hpp>
 #include <unifex/get_stop_token.hpp>
 #include <unifex/receiver_concepts.hpp>
+#include <unifex/scheduler_concepts.hpp>
```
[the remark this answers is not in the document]
LeeHowes: Oh, I must have been experimenting.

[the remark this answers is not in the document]
LeeHowes: Or I put it in the wrong file. Should have bee in sync_wait.hpp.

examples/fp_delegation.cpp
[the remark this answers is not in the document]
LeeHowes: I'll commit an update in a bit that should fix these.

include/unifex/sync_wait.hpp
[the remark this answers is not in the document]
LeeHowes: I've made a note on the P1898R1 PR to constrain this in the text.

As more than Receivers can be scheduler_providers maybe we have to define it in terms of the lifetime of the provider, rather than the set_ calls though.

examples/fp_delegation.cpp
LeeHowes: Yes. Maybe not for this example, though. It could be a useful primitive, and if we decide a first step is to make the delegating context a core unifex piece rather than just in this example.

examples/fp_delegation.cpp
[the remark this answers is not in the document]
LeeHowes: Done.

include/unifex/sync_wait.hpp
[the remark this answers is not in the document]
lewissbaker: Yes, I think defining the validity of the scheduler in terms of being limited to the lifetime of the provider seems like it might generalise to other use-cases.

include/unifex/sync_wait.hpp
```diff
@@ -125,27 +111,18 @@ using receiver_t = typename _receiver<T>::type;
...
-    promise.doneEvent_.emplace();
-  }
+  manual_event_loop ctx;
 
   // Store state for the operation on the stack.
   auto operation = connect(
-      ((Sender &&) sender),
-      _sync_wait::receiver_t<Result>{promise});
+      (Sender&&)sender,
+      _sync_wait::receiver_t<Result>{promise, ctx});
```
lewissbaker, line 121: We've lost the optimisation here that efficiently handles senders that complete synchronously without synchronisation.

Is there a way we can retain this while still allowing the forward-progress delegation properties of this implementation?

examples/fp_delegation.cpp
```diff
@@ -0,0 +1,177 @@
...
+  int capacity_;
+  timed_single_thread_context single_thread_context_;
+};
+
+template <typename DelegatedOperationState, typename LocalOperationState>
+class delegating_operation final {
+  public:
+  template<class InitFunc>
+  delegating_operation(InitFunc&& func, delegating_context* context) :
+    op_{std::in_place_type_t<remove_cvref_t<decltype(func())>>{}, func()}, context_{context} {
```
lewissbaker, line 75: This still seems like it might require the operation returned by `func()` to support a move-constructor in order to initialise the `op_`. You'd need to, instead, pass a wrapper around `func()` that had an implicit-conversion to the return-type of `func()` so that the variant implementation would call that in a context that was direct-initialising its data member with guaranteed copy-elision.

examples/fp_delegation.cpp
```diff
@@ -0,0 +1,177 @@
...
+  struct LocalContextType {
+    OpType op_;
+
+    inline void start() noexcept {
+      op_.start();
+    }
+  };
+
+  template <typename Receiver>
+  auto connect(Receiver&& receiver) {
```
lewissbaker, line 114: Would you also be able to implement this pattern with, say, a `fallback_scheduler()`?

eg. if we have a capacity-constrained scheduler like you have here, which has a schedule() operation that might fail with an error, we could turn it into a delegating scheduler by building a `fallback_scheduler(capacity_constrainted_scheduler(), current_scheduler())` where `current_scheduler` is a placeholder that forwards `schedule()` to the scheduler returned from the receiver's `get_scheduler()`.

I don't think it's quite as general as a scheduler that explicitly uses the calling context's scheduler, but it would seem to support the use-case in this example.

examples/fp_delegation.cpp
LeeHowes: So... I think you're saying that we don't want to call func here, we want to pass a type that defers the call to func until the variant needs it. Is that right?

examples/fp_delegation.cpp
LeeHowes: I think that would work. Are you asking for an implementation or just thinking aloud?

include/unifex/sync_wait.hpp
LeeHowes: If it is *guaranteed* to complete inline, then constexpr if checking that should be safe. It has to be guaranteed, though.

examples/fp_delegation.cpp
lewissbaker: > So... I think you're saying that we don't want to call func here, we want to pass a type that defers the call to func until the variant needs it. Is that right?

Yes. You need to pass something that is implicitly convertible to the result of calling func().
Alternatively you can use manual_lifetime_union and its .construct_from() helper to manually construct in-place.

examples/fp_delegation.cpp
lewissbaker: Mainly just thinking aloud.

I think for this to address the forward-progress requirements more generally, though, if we want to allow the user to pass an arbitrary executor to use for a parallel algorithm, then we'll need to create a different kind of 

This one assumes that we only need to fallback to the sync_wait() thread's executor if the other executor's schedule() completes with set_error(). However, if the other executor has accepted the work but just not scheduled it yet (eg. all its threads are currently blocked on other things) then we'd still want the sync_wait() executor to be able to pick up the work and execute it and to be able to cancel the work scheduled to the user-provided executor if the sync_wait() executor finishes the work first.

examples/fp_delegation.cpp
```diff
@@ -0,0 +1,188 @@
...
+  int capacity_;
+  timed_single_thread_context single_thread_context_;
+};
+
+template<typename T>
+struct FuncWrapper {
+  T func_;
+  operator std::invoke_result_t<T&&>() {
+    return func_();
+  }
```
lewissbaker, line 75: You are using `invoke_result_t<T&&>` but invoking `T&`

examples/fp_delegation.cpp
```diff
@@ -0,0 +1,188 @@
...
+    // Attempt to reserve a slot otherwise delegate to the downstream scheduler
+
+    using LC = LocalContextType<remove_cvref_t<decltype(unifex::connect(
+      unifex::schedule(context_->single_thread_context_.get_scheduler()), (Receiver&&)receiver))>>;
+    using op = delegating_operation<
+        remove_cvref_t<decltype(unifex::connect(
+          unifex::schedule(unifex::get_scheduler(std::as_const(receiver))), (Receiver&&)receiver))>,
+        LC>;
+    if(context_->reserve()) {
+      auto local_op = [receiver = (Receiver&&)receiver, context = context_]() mutable {
```
lewissbaker, line 135: No need to capture receiver by value here.

include/unifex/sync_wait.hpp
lewissbaker: I don't think we have the mechanism to query for 'blocking()` at compile-time yet.
I need to add some helper classes for supporting this which were mentioned in the tag_invoke paper.

PR #125, opened 2020-05-05
Define the scheduler concept in accordance with P0443R13

ericniebler:


include/unifex/scheduler_concepts.hpp
```diff
@@ -25,48 +25,144 @@
...
+    static auto _select() noexcept {
+      if constexpr (_with_tag_invoke<Scheduler>) {
+        return meta_tag_invoke_result<_fn>{};
+      } else if constexpr (_with_member_schedule<Scheduler>) {
+        return meta_quote1<_member_schedule_result_t>{};
+      } else if constexpr (executor<Scheduler>) {
+        return meta_quote1<_as_sender_t>{};
+      } else {
+        return type_always<void>{};
+      }
```
lewissbaker, line 98: Is the use of this _select() indirection here to avoid instantiating the return-types for `operator()` overloads that are not selected below?

include/unifex/trampoline_scheduler.hpp
```diff
@@ -134,6 +134,12 @@ class scheduler {
   schedule_sender schedule() const noexcept {
     return schedule_sender{maxRecursionDepth_};
   }
+  friend bool operator==(scheduler a, scheduler b) noexcept {
+    return a.maxRecursionDepth_ == b.maxRecursionDepth_;
+  }
+  friend bool operator!=(scheduler a, scheduler b) noexcept {
+    return a.maxRecursionDepth_ != b.maxRecursionDepth_;
+  }
```
lewissbaker, line 142: I wonder whether all trampoline_schedulers should be considered equal. Since they all dispatch work to the same trampoline queue in the end.

Do we have a definition of what equality of schedulers means?

include/unifex/scheduler_concepts.hpp
ericniebler: Yes.

include/unifex/trampoline_scheduler.hpp
ericniebler: I wondered that too. I think it means: refers to the same underlying execution context, in which case these schedulers should always compare equal. I'll change it.

PR #131, opened 2020-05-21
Portability fixes so that msvc 2019 compiles libunifex with /std:c++latest

ericniebler:
Requires `/std:c++latest /permissive- /Zc:externConstexpr`. `/std:c++17` will cause msvc to ICE. Leaving that for another PR.

### DIFF
--- a/.github/workflows/libunifex-ci.yml
+++ b/.github/workflows/libunifex-ci.yml
@@ -159,7 +159,6 @@ jobs:
             cc: "cl", cxx: "cl",
             environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
             cmake_args: "-D CMAKE_CXX_STANDARD:STRING=20",
-            experimental: true
           }
         - {
             name: "Windows MSVC 2019 Optimised (C++20)", artifact: "Windows-MSVC.tar.xz",
@@ -168,7 +167,6 @@ jobs:
             cc: "cl", cxx: "cl",
             environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
             cmake_args: "-D CMAKE_CXX_STANDARD:STRING=20",
-            experimental: true
           }
 
     steps:
--- a/.github/workflows/libunifex-ci.yml
+++ b/.github/workflows/libunifex-ci.yml
@@ -237,7 +237,7 @@ jobs:
             COMMAND "${{ matrix.config.environment_script }}" && set
             OUTPUT_FILE environment_script_output.txt
           )
-          set(ENV{CXXFLAGS} "/permissive-")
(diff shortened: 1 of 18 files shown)


include/unifex/detail/unifex_fwd.hpp
```diff
@@ -24,7 +24,7 @@ namespace unifex {
     template <typename E, typename R, typename = void>
     extern const bool _can_execute;
 
-    template <typename Sender, typename Fn, typename = void>
+    template <typename S, typename F, typename = void>
```
ericniebler, line 27: This trivial change is actualy necessary (!!!) because of https://developercommunity.visualstudio.com/content/problem/1042144/bug.html

include/unifex/manual_lifetime.hpp
```diff
@@ -40,7 +40,7 @@ class manual_lifetime {
   }
 
   template <typename Func>
-  T& construct_from(Func&& func) noexcept(noexcept(T(((Func &&) func)()))) {
```
ericniebler, line 43: This change is necessary because in the previous formulation MSVC tried to test the `noexcept`-ness of the move constructor, and for operation states, the move constructor is deleted. Since we rely on RVO, the move constructor is never invoked, so we only need to test the `noexcept`-ness of the function call.

include/unifex/schedule_with_subscheduler.hpp
```diff
@@ -51,10 +51,9 @@ namespace _subschedule {
       typename _detail::_return_value<remove_cvref_t<T>>::type;
 
     template <typename Scheduler>
-    using _default_result_t =
-      decltype(transform(
-        UNIFEX_DECLVAL(callable_result_t<decltype(schedule), Scheduler&>),
-        UNIFEX_DECLVAL(_return_value<Scheduler>)));
+    using _default_result_t = _tfx::sender<
+        callable_result_t<decltype(schedule), Scheduler&>,
+        _return_value<Scheduler>>;
```
ericniebler, line 56: This change is necessary for a mysterious reason I was never able to infer. Without this change, weird errors involving "`unknown-type`" break all uses of `schedule_with_subscheduler`.

lewissbaker, line 56: I don't think this is going to give the right result in the case that `transform()` has been customised for `schedule(Scheduler&)`.
It's assuming that `transform()` will fall-back to the default implementation.

A workaround here would be to manually call the `_tfx::sender` constructor in default implementation below, rather than calling `transform()`.

.github/workflows/libunifex-ci.yml
```diff
@@ -159,7 +159,6 @@ jobs:
             cc: "cl", cxx: "cl",
             environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
             cmake_args: "-D CMAKE_CXX_STANDARD:STRING=20",
-            experimental: true
```
lewissbaker, line 162: ❤️

PR #133, opened 2020-05-27
Forward progress delegation blocking optimisation

LeeHowes:
Bring back optimisation for sync_wait to allow inline tasks to complete without involving the scheduler.

### DIFF
--- a/include/unifex/sync_wait.hpp
+++ b/include/unifex/sync_wait.hpp
@@ -96,6 +96,7 @@ struct _receiver {
 
     friend auto
     tag_invoke(tag_t<get_scheduler>, const type& r) noexcept {
+      assert(r.ctx_ != nullptr);
       return r.ctx_->get_scheduler();
     }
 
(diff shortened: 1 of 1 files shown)


include/unifex/sync_wait.hpp
```diff
@@ -96,12 +96,14 @@ struct _receiver {
 
     friend auto
     tag_invoke(tag_t<get_scheduler>, const type& r) noexcept {
-      return r.ctx_.get_scheduler();
+      return r.ctx_->get_scheduler();
```
lewissbaker, line 99: `assert(r.ctx_ != nullptr);` ?

Or do we need to allow this to be called but have it fail if anything is scheduled on the returned scheduler?

include/unifex/sync_wait.hpp
```diff
@@ -113,16 +115,25 @@ template <typename Result, typename Sender>
 std::optional<Result> _impl(Sender&& sender) {
   using promise_t = _sync_wait::promise<Result>;
   promise_t promise;
-  manual_event_loop ctx;
+  auto blockingResult = blocking(sender);
+  const bool completesInline = blockingResult == blocking_kind::always_inline;
+
+  manual_lifetime<manual_event_loop> ctx;
```
lewissbaker, line 121: This is not exception-safe and will leak in the case that `connect()` below throws.
You should be able to just use a `std::optional` and just call `.emplace()` here instead.

LeeHowes: Quite right, I was expecting this to come up. For some reason I didn't think of emplace when I was playing with optional.

include/unifex/sync_wait.hpp
LeeHowes: That is interesting... given that ctx is runtime information, presumably we have to have a runtime failure case here.

Do we have any other operations that can fail to provide useful information via tag_invoke?

LeeHowes: I've changed it to assert and it does pass the current tests. Something that is guaranteed to complete inline presumably is not going to get_scheduler.

That doesn't feel right, though. It could easily be brittle if there is any chance that scheduler_provider can fail to be correct at runtime.

Returning an Inline scheduler here makes no sense either, allowing delegation to an inline scheduler would be a bad contract.

Maybe scheduler_provider should have get_scheduler return an optional scheduler?
Or we pair it with a has_scheduler operation?

LeeHowes: Discussed a scheduler wrapper approach that would send an error on submit if the real one was unavailable.

AnujYamdagni:

LeeHowes:
Yes, let's close it.

PR #139, opened 2020-06-12
Add windows low_latency_iocp_context execution context

lewissbaker:
This adds a proof-of-concept design of a single-threaded I/O scheduler
that tries to avoid always needing two syscalls to perform an async I/O
operation (one to start it and one to receive the completion-notification)
for operations that complete quickly with low-latency by introducing
an extra polling step that polls for I/O completion of recently started
I/O operations before going back to `GetQueuedCompletionStatusEx()`.

If the operation completes as part of the polling then we can immediately
process the result and call the receiver. However, we still need to wait until
we receive the completion-notification before we can free/reuse the
OVERLAPPED structures so in this case we detach the io-state from the
operation-state and let the io-state be garbage collected and returned
to the pool of available io-states once the completion-events have been
received.

A pool of io-states are preallocated when the I/O context is created and
placed onto a free-list that allows efficient O(1) allocation of a new io-state
when launching a new operation. If the io-state pool is exhausted then
new I/O operations are delayed from starting until existing I/O operations
complete and free up an io-state.

This design was based on my understanding of ideas on low-latency I/O
patterns I saw implemented in @ned14's llfio library, although it's possible
that in translating them to libunifex and sender/receiver that I have I missed
some key aspects of that design. Feedback welcome!

### DIFF
--- a/include/unifex/win32/low_latency_iocp_context.hpp
+++ b/include/unifex/win32/low_latency_iocp_context.hpp
@@ -1,0 +1,814 @@
+/*
+ * Copyright 2020-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <unifex/io_concepts.hpp>
+#include <unifex/manual_lifetime.hpp>
+#include <unifex/pipe_concepts.hpp>
+#include <unifex/receiver_concepts.hpp>
+#include <unifex/scheduler_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
+#include <unifex/span.hpp>
+#include <unifex/std_concepts.hpp>
+
+#include <unifex/detail/intrusive_list.hpp>
+#include <unifex/detail/intrusive_stack.hpp>
+#include <unifex/detail/atomic_intrusive_queue.hpp>
+
+#include <unifex/win32/detail/types.hpp>
+#include <unifex/win32/detail/safe_handle.hpp>
+
+#include <cstdint>
+#include <array>
+#include <thread>
+#include <cassert>
+#include <system_error>
+#include <tuple>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex::win32 {
+
+class low_latency_iocp_context {
+    class scheduler;
+
+    class readable_byte_stream;
+    class writable_byte_stream;
+
+    class schedule_sender;
+
+    template<typename Receiver>
+    struct _schedule_op {
+        class type;
+    };
+    template<typename Receiver>
+    using schedule_op = typename _schedule_op<Receiver>::type;
+
+    template<typename Buffer>
+    struct _read_file_sender {
+        class type;
+    };
+    template<typename Buffer>
+    using read_file_sender = typename _read_file_sender<Buffer>::type;
+
+    template<typename Buffer>
+    struct _write_file_sender {
+        class type;
+    };
+    template<typename Buffer>
+    using write_file_sender = typename _write_file_sender<Buffer>::type;
+
+    template<typename Buffer, typename Receiver>
+    struct _read_file_op {
+        class type;
+    };
+    template<typename Buffer, typename Receiver>
+    using read_file_op = typename _read_file_op<Buffer, Receiver>::type;
+
+    template<typename Buffer, typename Receiver>
+    struct _write_file_op {
+        class type;
+    };
+    template<typename Buffer, typename Receiver>
+    using write_file_op = typename _write_file_op<Buffer, Receiver>::type;
+
+public:
+    // Initialise the IOCP context and pre-allocate storage for I/O
+    // state needed to support at most 'maxIoOperations' concurrent I/O
+    // operations.
+    explicit low_latency_iocp_context(std::size_t maxIoOperations);
+
+    low_latency_iocp_context(low_latency_iocp_context&&) = delete;
+    low_latency_iocp_context(const low_latency_iocp_context&) = delete;
+    low_latency_iocp_context& operator=(const low_latency_iocp_context&) = delete;
+    low_latency_iocp_context& operator=(low_latency_iocp_context&&) = delete;
+
+    ~low_latency_iocp_context();
+
+    // Drive the event loop until a request to stop is communicated via
+    // the passed StopToken.
+    template<typename StopToken>
+    void run(StopToken st);
+
+    // Obtain a handle to this execution context that can be used to
+    // schedule work and open I/O resources.
+    scheduler get_scheduler() noexcept;
+
+private:
+    struct vectored_io_state;
+
+    // This value chosen so that vectored_io_state is 1024 bytes on 64-bit architectures
+    // and 512 bytes on 32-bit architectures.
+    static constexpr std::size_t max_vectored_io_size = 31;
+
+    struct operation_base {
+        explicit operation_base(low_latency_iocp_context& ctx) noexcept
+        : context(ctx)
+        {}
+
+        using callback_t = void(operation_base*) noexcept;
+        low_latency_iocp_context& context;
+        callback_t* callback = nullptr;
+        operation_base* next = nullptr;
+        operation_base* prev = nullptr;
+    };
+
+    using operation_queue = intrusive_list<operation_base, &operation_base::next, &operation_base::prev>;
+
+    struct io_operation : operation_base {
+        explicit io_operation(low_latency_iocp_context& context, handle_t fileHandle, bool skipNotificationOnSuccess) noexcept
+        : operation_base(context)
+        , fileHandle(fileHandle)
+        , skipNotificationOnSuccess(skipNotificationOnSuccess)
+        , ioState(nullptr)
+        {}
+
+        const handle_t fileHandle;
+        const bool skipNotificationOnSuccess;
+        vectored_io_state* ioState;
+
+        // Cancel outstanding I/O operations (if any)
+        void cancel_io() noexcept;
+
+        // Poll for whether or not the operation has completed.
+        // Returns 'true' if the operation is completed (in which case the operation
+        // has 'acquire' semantics), 'false' otherwise. 
+        bool is_complete() noexcept;
+
+        // Start reading the next 'buffer.size()' bytes into 'buffer'.
+        //
+        // Returns 'true' if a additional read-operations can be submitted.
+        // This will only be the case if all of the following are true:
+        // - a read of the entirety of 'buffer' was successfully started
+        // - we haven't reached the maximum number of I/O operations in
+        //   this batch.
+        //
+        // Once all I/O operations have been started, the caller should
+        // schedule a 'poll' of this operation by setting this->callback
+        // and calling context.schedule_poll(this). This will schedule the
+        // callback to be run immediately (if it completed synchronously).
+        bool start_read(span<std::byte> buffer) noexcept;
+
+        // Start writing the context of 'buffer' to fileHandle.
+        bool start_write(span<const std::byte> buffer) noexcept;
+
+        std::size_t get_result(std::error_code& ec) noexcept;
+    };
+
+    struct vectored_io_state {
+        // Parent operation.
+        //
+        // May be nullptr if this operation already completed due to a poll.
+        // If this is the case then when all pending completion notifications
+        // are received then this will just go straight back on the free-list.
+        io_operation* parent = nullptr;
+
+        // Intrusive list ptr.
+        vectored_io_state* next = nullptr;
+        vectored_io_state* prev = nullptr;
+
+        // Total number of operations started
+        std::uint8_t operationCount = 0;
+
+        // Number of operations not yet received completion-notification
+        // via the IOCP. This structure is not free to be reused until
+        // this number reaches zero.
+        std::uint8_t pendingCompletionNotifications = 0;
+
+        // Whether or not the 'parent' has already been notified of completion.
+        bool completed = false;
+
+        std::array<overlapped, max_vectored_io_size> operations;
+    };
+
+    struct stop_operation : operation_base {
+        explicit stop_operation(low_latency_iocp_context& ctx) noexcept
+        : operation_base(ctx) {
+            this->callback = &request_stop_callback;
+        }
+
+        ~stop_operation() {
+            if (isEnqueued) {
+                // Flush any items in the remote-queue into the ready queue
+                // just in case this operation is still in the remote queue.
+                (void)context.try_dequeue_remote_work();
+                
+                // This operation should now be in the ready queue, remove it.
+                context.readyQueue_.remove(this);
+            }
+        }
+
+        void start() noexcept {
+            if (context.is_running_on_io_thread()) {
+                stopRequestedFlag = true;
+            } else {
+                isEnqueued = true;
+                context.schedule_remote(this);
+            }
+        }
+
+        static void request_stop_callback(operation_base* op) noexcept {
+            auto& self = *static_cast<stop_operation*>(op);
+            self.stopRequestedFlag = true;
+            self.isEnqueued = false;
+        }
+
+        bool stopRequestedFlag = false;
+
+    private:
+        bool isEnqueued = false;
+    };
+
+    struct stop_callback {
+        stop_operation& op;
+
+        void operator()() noexcept {
+            op.start();
+        }
+    };
+
+    void run_impl(bool& stopFlag);
+
+    // Dequeue items from remote queue and move them to the
+    // ready-to-run queue.
+    // Returns true if any items were dequeued.
+    bool try_dequeue_remote_work() noexcept;
+
+    bool poll_is_complete(vectored_io_state& state) noexcept;
+
+    // Obtain the I/O state that contains a given overlapped structure.
+    vectored_io_state* to_io_state(overlapped* o) noexcept;
+
+    bool is_running_on_io_thread() const noexcept {
+        return activeThreadId_.load(std::memory_order_relaxed) == std::this_thread::get_id();
+    }
+
+    void schedule(operation_base* op) noexcept;
+    void schedule_local(operation_base* op) noexcept;
+    void schedule_remote(operation_base* op) noexcept;
+
+    // If an I/O state is available, attaches an unused I/O state to
+    // op->ioState and returns true, otherwise returns false if no
+    // I/O state is currently available.
+    //
+    // If 'true' is returned then the caller can populate op->ioState
+    // and initiate the I/O using its OVERLAPPED structures.
+    [[nodiscard]] bool try_allocate_io_state_for(io_operation* op) noexcept;
+
+    // Schedule the specified 'op' to be called back (calling op->callback)
+    // when an I/O state becomes available and has been allocated to 'op'.
+    //
+    // Do this only after an unsuccessful call to try_allocate_io_state_for()
+    // asynchronously wait until some other I/O operation completes and frees
+    // up an other I/O state.
+    void schedule_when_io_state_available(io_operation* op) noexcept;
+
+    // Mark the io state as released and let it return to the pool.
+    void release_io_state(vectored_io_state* state) noexcept;
+
+    // Schedule this operation to be polled next time we run out of work
+    // in the ready-queue before going back to the OS for completion-events.
+    void schedule_poll_io(io_operation* op) noexcept;
+
+    // Attempt to associate the specified file-handle with this I/O context
+    // so that its I/O completion events are posted to this context's IOCP.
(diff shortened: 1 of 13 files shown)


include/unifex/win32/low_latency_iocp_context.hpp
```diff
@@ -0,0 +1,814 @@
...
+
+        // Number of operations not yet received completion-notification
+        // via the IOCP. This structure is not free to be reused until
+        // this number reaches zero.
+        std::uint8_t pendingCompletionNotifications = 0;
+
+        // Whether or not the 'parent' has already been notified of completion.
+        bool completed = false;
+
+        std::array<overlapped, max_vectored_io_size> operations;
```
ned14, line 194: Note that you can pack these more tightly than `OVERLAPPED`, which is a Win32 extension of `IO_STATUS_BLOCK`. You only actually need `IO_STATUS_BLOCK`.

source/win32/low_latency_iocp_context.cpp
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    if (remaining > 0) {
+        constexpr std::uint32_t completionBufferSize = 128;
+        OVERLAPPED_ENTRY completionBuffer[completionBufferSize];
+        std::memset(&completionBuffer, 0, sizeof(completionBuffer));
+
+        do {
+            DWORD numEntriesRemoved = 0;
+            BOOL ok = ::GetQueuedCompletionStatusEx(
```
ned14, line 73: Be aware that this batch function is umpteen times slower than the single packet edition. It makes sense if you are pulling down many completions at a time, it amortises out, but if you know that you're not, looping processing them singly can be very considerably faster.

source/win32/low_latency_iocp_context.cpp
```diff
@@ -0,0 +1,682 @@
...
+        volatile io_status_block* iosb = reinterpret_cast<io_status_block*>(o);
+        if (iosb->Status == STATUS_PENDING) {
+            return false;
+        }
+    }
+
+    // Make sure that we have visibility of the results of the I/O operations
+    // (assuming that the OS used a "release" memory semantic when writing
+    // to the 'Status' field).
+    std::atomic_thread_fence(std::memory_order_acquire);
```
ned14, line 259: I suspect this is overkill, any syscall is an implicit sequentially consistent thread fence.

source/win32/low_latency_iocp_context.cpp
```diff
@@ -0,0 +1,682 @@
...
+        throw std::system_error{
+            static_cast<int>(errorCode), std::system_category(),
+            "CreateIoCompletionPort"};
+    }
+
+    const BOOL ok = ::SetFileCompletionNotificationModes(
+        fileHandle,
+        FILE_SKIP_COMPLETION_PORT_ON_SUCCESS |
+        FILE_SKIP_SET_EVENT_ON_HANDLE);
+    if (!ok) {
```
ned14, line 392: You can't do this unfortunately. Whether skipping completion port on success is available or not depends on the NIC driver. Some NIC drivers don't implement it. So you really do need to fallback if it fails.

source/win32/low_latency_iocp_context.cpp
```diff
@@ -0,0 +1,682 @@
...
+void low_latency_iocp_context::io_operation::cancel_io() noexcept {
+    assert(ioState != nullptr);
+
+    // Cancel operations in reverse order so that later operations
+    // are cancelled first and don't accidentally end up with earlier
+    // operations being cancelled and later ones completing due to
+    // a race.
+    for (std::uint16_t i = ioState->operationCount; i != 0; --i) {
+        win32::overlapped* o = &ioState->operations[i-1];
+        (void)::CancelIoEx(fileHandle, reinterpret_cast<OVERLAPPED*>(o));
```
ned14, line 409: Cancelling i/o does not need to occur immediately, it may take time. Yet you are returning immediately. This equals potential memory corruption.

source/win32/low_latency_iocp_context.cpp
```diff
@@ -0,0 +1,682 @@
...
+        op.Internal = 0;
+        op.InternalHigh = 0;
+        op.Offset = 0;
+        op.OffsetHigh = 0;
+        op.hEvent = nullptr;
+
+        // Truncate over-large chunks to a number of bytes that will still
+        // preserve alignment requirements of the underlying device if this
+        // happens to be an unbuffered storage device.
+        // In this case we truncate to the largest multiple of 64k less than
```
ned14, line 463: I think 4Kb would be enough. Any single device will support 4Kb forever. Only RAID arrays have larger fundamental i/o sizes, and that could be anything at all. For example, 128Kb stripe sizes are very common.

ned14:
> This design was based on my understanding of ideas on low-latency I/O
patterns I saw implemented in @ned14's llfio library, although it's possible
that in translating them to libunifex and sender/receiver that I have I missed
some key aspects of that design. Feedback welcome!

Your design is pretty close to what I presented to WG21 in Prague.

However, my "second attempt" based on throwing away that first attempt, I made quite a few changes:

1. Use of atomics, or any thread synchronisation at all, is configurable. This makes a big difference, for single threaded use I beat ASIO by 40% with synchronisation, I beat ASIO by 60% without them. That's a +50% gain.

2. I packed the cache lines touched during i/o down by about half from what they were. From a quick skim of this PR, I think you're touching as many cache lines as my original design did.

3. I did use volatile to inspect the `IO_STATUS_BLOCK` structure members before, but in my second attempt I was able to dispense with that by having whomever processes the i/o completion set the (possibly atomic) state for the i/o. As the only way an `IO_STATUS_BLOCK` can get updated is due to waiting on the handle, or reaping i/o completions for a port associated with the handle, you can mux in the right checking code immediately after any pertinent syscall. That, in turn, eliminates much of the need for volatile or atomics.

Finally, I'd love to see a comparative benchmark between my code and your code. I am currently working a 70 hour week due to working for a startup, and we deploy live next month. So I won't be able to properly get back to this until mid July. But do feel free to ask any questions.

include/unifex/win32/low_latency_iocp_context.hpp
lewissbaker: Thanks, I had considered doing this but couldn't seem to find winsock APIs equivalent to WSASend/WSARecv that take IO_STATUS_BLOCK instead of OVERLAPPED and didn't want to have to maintain separate io-state pools for files/pipes vs sockets.

Although the socket APIs have native support for vectorised I/O so it will generally only ever use the first OVERLAPPED anyway.

Maybe this could be a union of a single `overlapped` and array of `io_status_block` and just let the I/O operation choose which one to use based on the APIs it's calling?

source/win32/low_latency_iocp_context.cpp
lewissbaker: I thought the idea was that when we poll we aren't making a syscall and so we need to do some synchronisation to ensure we have visibility of the (possibly other kernel thread's) writes to the buffer (at least for reads anyway). Maybe the barrier can be omitted from here?

An alternative I considered was just calling GetOverlappedResultEx() with a zero timeout and bAlertable=FALSE. Do you know if that ends up doing everything in user-space? Or were you manually reading the NTSTATUS because GetOverlappedResult() is too slow?

include/unifex/win32/low_latency_iocp_context.hpp
ned14: That's because Winsock is actually an abomination :). Internally, Windows literally cannot do scatter-gather i/o. The kernel structures don't have the facility. So, the winsock loadable does a poor emulation by issuing an IRP per buffer.

You're much better off, in my opinion, in dispensing with winsock entirely and just use Windows RIO throughout. They're far better designed APIs (they look very like io_uring, unsurprisingly), they work in terms of registered i/o buffers and submission and completion ring buffers stored in memory shared with the kernel. Your implementation would simply memcpy user buffers into a portion of those registered i/o buffers, submit an i/o request to RIO, it'll complete via IOCP using the normal `IO_STATUS_BLOCK` if you ask it to do that.

In other words, I'm saying use RIO to do socket i/o just like you do file i/o. Then everything is using the same identical IOCP mechanism, no special casing the network i/o. It's how Windows works internally in any case, so you'll be skipping an entire layer of software.

Once you're into the RIO weeds, you may wish to pay attention to Microsoft's shiny new RIO + Compute integration. I know all the Microsoft folk are super excited about this, and if you can demonstrate that libunifex uses their shiny new toys, that's a nice number of library implementer votes in favour in committee.

source/win32/low_latency_iocp_context.cpp
lewissbaker: This is specifically for named pipes and files, which afaict should always support skipping completion port...?

I am aware of the need to query the capability of this for sockets, but have not yet added socket support.

When I do I plan to add a new `associate_socket()` function that does the capability a check as I did in [cppcoro](https://github.com/lewissbaker/cppcoro/blob/69e575480afe97fd342a44f580ba8ff3b4a9a58a/lib/socket.cpp#L146).

source/win32/low_latency_iocp_context.cpp
ned14: I can't say if this particular atomic fence is necessary or not, I haven't traced through all the possible call paths. 

I don't call `GetOverlappedResultEx()` because it has a trivial implementation, see https://doxygen.reactos.org/dd/df8/iocompl_8c.html#ad5f55ae9b6f196ec2ae6cf775176fabe. And by not using it, I don't need to use `OVERLAPPED` which is larger than `IO_STATUS_BLOCK` because of the extra event handle, which you don't need as you can just loop waiting on the i/o handle and checking whether the status value has updated (I actually pump a single IOCP completion each loop too, if you're waiting anyway, might as well do something useful. But then mine is as-if blocking i/o, unlike yours).

source/win32/low_latency_iocp_context.cpp
lewissbaker: This function is called as a request to cancel the operation which is called from the stop_callback in response to a call to `request_stop()`. The caller does not assume that the cancellation will be complete when this function returns.

The storage for the ioState and its OVERLAPPED structures is not released until we have received completion-notifications for all pending I/Os.

lewissbaker:
> 1. Use of atomics, or any thread synchronisation at all, is configurable. This makes a big difference, for single threaded use I beat ASIO by 40% with synchronisation, I beat ASIO by 60% without them. That's a +50% gain.

This implementation only supports single-threaded I/O. Other than the atomic fence in `is_complete()` helper I don't _think_ I have any atomic usage on the critical path.

The exception is if you want to support cancellation of I/O, which requires synchronisation to attach/detach stop-callbacks.

> 2. I packed the cache lines touched during i/o down by about half from what they were. From a quick skim of this PR, I think you're touching as many cache lines as my original design did.

Yes, I think I can probably slim down the use of `OVERLAPPED` to just use `IO_STATUS_BLOCK` for files/pipes but I think I'll still need to use `OVERLAPPED` for sockets to be able to use WSARecv/WSASend.

Was that the only place you had cache-line concerns?

> 3. I did use volatile to inspect the `IO_STATUS_BLOCK` structure members before, but in my second attempt I was able to dispense with that by having whomever processes the i/o completion set the (possibly atomic) state for the i/o. As the only way an `IO_STATUS_BLOCK` can get updated is due to waiting on the handle, or reaping i/o completions for a port associated with the handle, you can mux in the right checking code immediately after any pertinent syscall. That, in turn, eliminates much of the need for volatile or atomics.

Are you saying that polling the `IO_STATUS_BLOCK`'s `Status` field without an intervening call to e.g. `GetOverlappedResult()` won't actually update the `Status` field if the I/O completed in the background?

Should I be calling `GetOverlappedResult()` when polling for early completion instead of reading the NTSTATUS directly?

> Finally, I'd love to see a comparative benchmark between my code and your code.

A very rough initial test that runs a single reader and single writer communicating over a named-pipe on the same thread gets a ballpark of ~1.5M IOPS on a Windows VM running on a Mac laptop (with 1 byte reads, 100 byte writes, 16KB pipe buffer).

This is not doing any vectored I/O, though, and I'm not measuring any latency, etc.

I'll need to go back to look at your benchmarks to see what aspects they are testing. But, likewise, it may be some time before I get a chance to build a comparative benchmark.

source/win32/low_latency_iocp_context.cpp
lewissbaker: So should I be keeping track of the number of pending I/Os and only calling the batch API if this number is above some threshold?

I can imagine cases where, even though I have a reasonable number of outstanding I/Os, that I might still only have one at a time coming in.

What heuristics should this code be using to determine whether to call the single vs batch API?

ned14:
> This implementation only supports single-threaded I/O. Other than the atomic fence in is_complete() helper I don't think I have any atomic usage on the critical path.

One sees the maximum benefit of not using atomics if the compiler can statically deduce that no atomics could ever occur, as if it ever gets a hint of there being potentially an atomic could be called, it'll penalise codegen. The way I usually do this is to use a templated internal implementation which selects between atomics or non-atomic editions, and then propagate the runtime selection of that type as high up the use chain as is possible, ideally via vptr dispatch which can be devirtualised. That then encourages the compiler to generate two entirely separate implementations, rather than mixing little sections of separate implementation with constant branching to choose between them.

I appreciate that that design approach will seem a bit funky to you guys who tend to write mostly header only code. But the inversion gives me maximally thin headers, which means seconds to build during development, whilst a unity build still produces very high quality codegen albeit with very long compile times.

(As an aside, I spent a large part of yesterday trying to persuade GCC to not ICE during unity builds in Release of our custom database. I really, really, really wish I could use clang which "just works", but my employer refuses to use clang. So I expect to spend much of today undoing our unity release build, just for GCC alone. Sigh)

> Yes, I think I can probably slim down the use of OVERLAPPED to just use IO_STATUS_BLOCK for files/pipes but I think I'll still need to use OVERLAPPED for sockets to be able to use WSARecv/WSASend.

Windows RIO! It's *so* much nicer than WSA ...

> Was that the only place you had cache-line concerns?

I should emphasise that my comment was based purely on gut feeling from a quick skim. I didn't do any analysis like counting pointer dereferences.

> A very rough initial test that runs a single reader and single writer communicating over a named-pipe on the same thread gets a ballpark of ~1.5M IOPS on a Windows VM running on a Mac laptop (with 1 byte reads, 100 byte writes, 16KB pipe buffer).

On my completely different hardware, so these numbers are completely incomparable, I'm achieving 375,000 single byte read/write ops/sec. This is from before i/o initiation to after i/o release i.e. after IOCP release. If I do no syscalls at all, and simply go through the motions as if i/o had happened, I achieve 8.3 million ops/sec. I have manually walked the assembler generated by the no syscalls case, and I believe it close to optimal given the problem being solved. In the end, tracking the state of the i/o operation through its lifecycle costs what it costs, and I don't think it can be reduced by much further. However, in terms of good news, thanks to superscalar execution I think that the CPU could do a complete syscall in exactly the same time i.e. it can parallelise doing the actual i/o in approx the same time as we take to track lifecycle state. Obviously that's unproven without testing, but 5 million peak iops for Sender-Receiver I think is a reasonable aspiration.

(To put that into context, kernels currently can route ~2 million packets per second per core without hardware assistance, and it is currently believed that this will be a hard maximum going forth)

I'm not sure what you guys are actually doing at Facebook exactly, but you may wish to look into Vector Packet Processing (VPP) which is an open source packet switch. Using that software design, they can route 11 million packets per second per core. As a comparator, my employer's packet routing achieves 100k packets/sec/core, and I know for a fact that we are currently class leading in finance because we keep getting new customers who are wowed by our software packet switching performance. Needless to say my recommendations to replace the current software design with something like VPP have fallen on deaf ears. They'll learn when someone out innovates them.

source/win32/low_latency_iocp_context.cpp
ned14: In my first design, I kept a count, and if outstanding i/o were below 16 or so, I looped the single-completion API.

In my second design, I eliminated the count, and made when completions get done parameterised. I now implement completions by doing completions depending on context, so instead of there being a global `run()`, they get dispatched ad-hoc, wherever they need to be. As the caller knows whether they need to spin fast or don't care, they can dictate whether to complete one, two, or as many as possible completions. I found this design much better. But I appreciate that that isn't an Executor, and so it probably doesn't help you.

include/unifex/win32/low_latency_iocp_context.hpp
lewissbaker: @ned14 I've started looking into converting the use of `OVERLAPPED` into `IO_STATUS_BLOCK` but have been running into the question of which function I should call instead of `CancelIoEx()` to cancel an individual pending I/O.

Can I still call `CancelIoEx()` with the `IO_STATUS_BLOCK*` cast to an `OVERLAPPED*`? Or is there some other `NtCancel...()` function that I need to call instead?

ned14: `NtCancelIoFileEx` takes a handle and two pointers to `IO_STATUS_BLOCK`. `OVERLAPPED` is simply an `IO_STATUS_BLOCK` (two void*'s) with extra stuff after it, two extra void*.

Only Win32 ever uses the extra stuff in `OVERLAPPED`. The NT kernel exclusively works in `IO_STATUS_BLOCK`. If you therefore only use the NT kernel APIs, you gain.

There are two big advantages to using `IO_STATUS_BLOCK`:

1. They are half the size, and you generally need one per i/o, so that's a nice improvement on CPU cache loading.

2. You can place your own extra stuff after each `IO_STATUS_BLOCK` instead e.g. a per-i/o pointer. The special branch of LLFIO I did Jan 2020 I think put a pointer to the Sender and a pointer to the `llfio::io_handle` or something like that. I might be completely wrong on this, but I think `IO_STATUS_BLOCK` requires an 8 byte alignment, so on 64 bit you can legally have three-void* `IO_STATUS_BLOCK`, but on 32 bit you must pad. The current LLFIO, by the way, has two-void* sized `IO_STATUS_BLOCK` and what we do is use aligned malloc to ensure the whole block of `IO_STATUS_BLOCK` is aligned to some boundary. We can then AND round down the pointer to the head of the block, and at the very front we store a pointer to `llfio::io_handle`, and from there we know how to handle things. Packing the `IO_STATUS_BLOCK` more densely was found to be a nice source of performance gain, especially as you need lots of them.

lewissbaker: Thanks for the help with NtCancelIoFileEx().

I've updated the PR to now deal directly with IO_STATUS_BLOCK structures, which has halved the memory used for each I/O state and it now uses the lower-level NtReadFile/NtWriteFile/NtCancelIoFileEx/NtRemoveIoCompletion/etc. APIs.

The vectored_io_state now holds 4 pointers worth of header stuff per 'sender operation', and also contains an array of 30 IO_STATUS_BLOCK entries. So each sender can issue a batch of up to 30 buffers for a scatter/gather-like operation.

ned14: Very nice!

My Christmas holiday side project this year is to add to LLFIO some sort of mechanism which:

1. Is Sender-Receiver, Executors and Schedulers compatible (but stops just short of those, same as `llfio::io_multiplexer`).
2. Wraps the system provided self-dynamically-scaling kernel thread pool into a low level portable API which somehow (?) will integrate with any `llfio::io_multiplexer`.
3. Can pace the work submitted to the system provided kernel thread pool according to current i/o operation latency on the relevant backing hardware devices.

All this is quite hard to achieve, which is why I can't really justify writing one on work time, because chances are good I won't succeed quickly. And besides, I think WG21 is currently rather too heavy on theoretical abstraction, and too light on real world practically useful implementations of all this stuff they propose standardising, in my opinion.

I suppose I better explain what specifically I want to solve here:

1. I want the sockets to all be using Windows RIO.
2. I want Windows RIO's IOCP interface to be wired into the Win32 i/o aware thread pool, which self adjusts its concurrency.
3. I want all kernel thread level concurrency to use the Win32 i/o aware thread pool.
4. With all those three Windows facilities wired together as the Microsoft folk designed and optimised the snot out of them, I want all that to appear as natural and portable low level APIs in standard C++.
5. And for that natural and portable low level API to also "do the right thing" with Grand Central Dispatch on Mac OS and other kernel thread concurrency + i/o platform-specific integrations. You might notice here that if the API is designed right you get "for free" hardware accelerated TLS sockets, or any other kind of socket-like abstraction combined with concurrency framework which your system can provide.
6. And finally, I want the socket i/o concurrency solution to work well (on Linux and Windows at least) with memory mapped file i/o.

That final point will need a bit of explaining. The problem needing solved is that the system provided dynamically self adjusting thread pool generally tries to keep as many threads active as there is hardware concurrency. So, if a thread stalls on a page fault or syscall or i/o, more kernel threads are created, whereas if more kernel threads are running non-blocked than hardware concurrency, then kernel threads are removed. The problem I need to solve is that if you point such a pool at lots of memory mapped file i/o, and you're not doing a ton of CPU work, as storage becomes congested, the thread pool "runs away" with adding lots more concurrency. This makes the problem exponentially worse until the system appears to hang because you overload storage.

So what I basically need is some sort of Kalman filter on total process concurrency whereby we cut back concurrency on non-socket i/o if that i/o is congested, but can still increase concurrency for non-i/o bound work and socket i/o, because the system thread pool has special support for being aware of socket i/o.

Anyway all the above turned into rather longer than I intended, but perhaps now you can understand the difficulty of what I'll be attempting, but also why if I'm successful, how extremely useful it would be. At work in our custom DB, we currently hand tune the thread counts which is brittle and prone to pathological corner case performance drop outs. If we could replace that with what I have in mind, the system would self-optimise itself, and even better, would self adjust over time to remain optimal.

Anyway merry christmas to all of you, and have a great holiday!

PR #141, opened 2020-06-21
Find if

LeeHowes:
Simple serial implementation of find_if over iterators.

### DIFF
--- a/examples/find_if.cpp
+++ b/examples/find_if.cpp
@@ -44,7 +44,8 @@ int main() {
           just(begin(input), end(input), 3),
           [&](const int& v, int another_parameter) {
             return v == another_parameter;
-          }),
+          },
+          unifex::seq),
       [](std::vector<int>::iterator v, int another_parameter) {
         assert(another_parameter == 3);
         return v;
--- a/examples/find_if.cpp
+++ b/examples/find_if.cpp
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <unifex/just.hpp>
+#include <unifex/let.hpp>
+#include <unifex/scheduler_concepts.hpp>
+#include <unifex/sync_wait.hpp>
+#include <unifex/timed_single_thread_context.hpp>
+#include <unifex/find_if.hpp>
+#include <unifex/transform.hpp>
+#include <unifex/when_all.hpp>
+
+#include <chrono>
+#include <iostream>
+
+using namespace unifex;
+using namespace std::chrono;
+using namespace std::chrono_literals;
+
+int main() {
+  std::vector<int> input{1, 2, 3, 4};
+  // Apply linear find_if.
+  // As for std::find_if it returns the first instance that matches the
+  // predicate where the algorithm takes an iterator pair as the first
+  // two arguments, and forwards all other arguments to the predicate and
+  // onwards to the result.
+  // Precise API shape for the data being passed through is TBD, this is
+  // one option only.
+  std::optional<std::vector<int>::iterator> result = sync_wait(
+    transform(
+      find_if(
+          just(begin(input), end(input), 3),
+          [&](const int& v, int another_parameter) {
+            return v == another_parameter;
+          }),
+      [](std::vector<int>::iterator v, int another_parameter) {
+        assert(another_parameter == 3);
+        return v;
+      }));
+
+  std::cout << "all done " << **result << "\n";
+
+  return 0;
+}
--- a/include/unifex/find_if.hpp
+++ b/include/unifex/find_if.hpp
@@ -36,16 +36,6 @@
 
 namespace unifex {
 namespace _find_if {
-namespace detail {
-  template <typename Result, typename = void>
-  struct result_overload {
-    using type = type_list<Result>;
-  };
-  template <typename Result>
-  struct result_overload<Result, std::enable_if_t<std::is_void_v<Result>>> {
-    using type = type_list<>;
-  };
-}
 
 template <typename Receiver, typename Func, typename FuncPolicy>
 struct _receiver {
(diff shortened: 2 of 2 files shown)


examples/find_if.cpp
```diff
@@ -0,0 +1,57 @@
...
+  // As for std::find_if it returns the first instance that matches the
+  // predicate where the algorithm takes an iterator pair as the first
+  // two arguments, and forwards all other arguments to the predicate and
+  // onwards to the result.
+  // Precise API shape for the data being passed through is TBD, this is
+  // one option only.
+  std::optional<std::vector<int>::iterator> result = sync_wait(
+    transform(
+      find_if(
+          just(begin(input), end(input), 3),
```
lewissbaker, line 44: Ideally this would be taking a `range` as input, but I understand that's difficult at the moment given that `std::ranges` is not yet widely available in all standard libraries.

The other thing I'm vacillating over here is whether this algorithm should take a `sender_of<range>` or just the `range` as input.

The former might allow some better chaining ergonomics and optimisation possibilities, but makes it harder to use for simpler cases, e.g. from a coroutine, as you'd often need to wrap the input in a `just()` wrapper when you already have a range.

The latter would still support chaining, but would require something like the `let()` algorithm to connect the output of one sender to the argument to the function.
e.g. from a coroutine:
```
std::vector<int> v = co_await produce_some_vector();
auto it = co_await unifex::find_if(v, [](int i) { return i == 3; }, unifex::seq);
if (it != v.end()) {
  *it = 4;
  co_return true;
}
co_return false;
```
but from a pure sender approach:
```
let(
  produce_some_vector(),
  [](std::vector<int>& v) {
    return transform(
      find_if(v, [](int i) { return i == 3;}),
      [&](std::vector<int>::iterator it) {
        if (it != v.end()) {
          *it = 4;
          return true;
        }
        return false;
      });
  });
```
or with pizza-operator:
```
produce_some_vector()
|> let([](std::vector<int>& v) {
    return
      find_if(v, [](int i) { return i == 3;})
      |> transform([&](std::vector<int>::iterator it) {
        if (it != v.end()) {
          *it = 4;
          return true;
        }
        return false;
      });
  });
```


For now I think what you have is good enough to get started, we can refine the shape later.

include/unifex/find_if.hpp
```diff
@@ -0,0 +1,205 @@
...
+namespace _find_if {
+namespace detail {
+  template <typename Result, typename = void>
+  struct result_overload {
+    using type = type_list<Result>;
+  };
+  template <typename Result>
+  struct result_overload<Result, std::enable_if_t<std::is_void_v<Result>>> {
+    using type = type_list<>;
+  };
```
lewissbaker, line 47: These are not currently used

include/unifex/find_if.hpp
```diff
@@ -0,0 +1,205 @@
...
+      typename conditional_t<
+        tag_invocable<_fn, Sender, Func>,
+        meta_tag_invoke_result<_fn>,
+        meta_quote3<_find_if::sender>>::template apply<Sender, Func, FuncPolicy>;
+  public:
+    template(typename Sender, typename Func, typename FuncPolicy)
+      (requires tag_invocable<_fn, Sender, Func, FuncPolicy>)
+    auto operator()(Sender&& predecessor, Func&& func, FuncPolicy policy) const
+        noexcept(is_nothrow_tag_invocable_v<_fn, Sender, Func, FuncPolicy>)
+        -> _result_t<Sender, Func, FuncPolicy> {
```
lewissbaker, line 188: Can this just be `tag_invoke_result_t<_fn, Sender, Func, FuncPolicy>`?

include/unifex/find_if.hpp
```diff
@@ -0,0 +1,205 @@
...
+      return unifex::tag_invoke(_fn{}, (Sender&&)predecessor, (Func&&)func, (FuncPolicy&&)policy);
+    }
+    template(typename Sender, typename Func, typename FuncPolicy)
+      (requires (!tag_invocable<_fn, Sender, Func, FuncPolicy>))
+    auto operator()(Sender&& predecessor, Func&& func, FuncPolicy policy) const
+        noexcept(std::is_nothrow_constructible_v<
+          _find_if::sender<Sender, Func, FuncPolicy>, Sender, Func, FuncPolicy>)
+        -> _result_t<Sender, Func, FuncPolicy> {
```
lewissbaker, line 196: And similarly, can the return type here just be `_find_if::sender<Sender, Func, FuncPolicy>`?

include/unifex/find_if.hpp
```diff
@@ -0,0 +1,205 @@
...
+      (requires tag_invocable<_fn, Sender, Func, FuncPolicy>)
+    auto operator()(Sender&& predecessor, Func&& func, FuncPolicy policy) const
+        noexcept(is_nothrow_tag_invocable_v<_fn, Sender, Func, FuncPolicy>)
+        -> _result_t<Sender, Func, FuncPolicy> {
+      return unifex::tag_invoke(_fn{}, (Sender&&)predecessor, (Func&&)func, (FuncPolicy&&)policy);
+    }
+    template(typename Sender, typename Func, typename FuncPolicy)
+      (requires (!tag_invocable<_fn, Sender, Func, FuncPolicy>))
+    auto operator()(Sender&& predecessor, Func&& func, FuncPolicy policy) const
+        noexcept(std::is_nothrow_constructible_v<
```
lewissbaker, line 194: Using `std::is_nothrow_constructible_v` here with the sender type to determine validity can cause problems because it forces an instantiation of the sender type even if it might not be valid and this could lead to errors that are not SFINAE errors.

Instead, I think you need to duplicate the requirements here so that they are in terms of the template parameters directly.
e.g.
```
noexcept(
  std::is_nothrow_constructible_v<remove_cvref_t<Sender>, Sender> &&
  std::is_nothrow_constructible_v<remove_cvref_t<Func>, Func> &&
  std::is_nothrow_constructible_v<remove_cvref_t<FuncPolicy>, FuncPolicy>)
```

include/unifex/find_if.hpp
```diff
@@ -0,0 +1,205 @@
...
+    std::invoke(visit, r.receiver_);
+  }
+};
+
+template <typename Predecessor, typename Func, typename FuncPolicy>
+struct _sender {
+  struct type;
+};
+template <typename Predecessor, typename Func, typename FuncPolicy>
+using sender = typename _sender<remove_cvref_t<Predecessor>, std::decay_t<Func>, FuncPolicy>::type;
```
lewissbaker, line 121: Generally I've been trying to avoid applying the `remove_cvref_t` and `decay_t` transformations on template args in the type-aliases here as it makes it more difficult to tell whether it's doing the correct thing at the use-site.

examples/find_if.cpp
LeeHowes: There are wider questions with this one.

Should we be able to pass a container rather than a view?
just(vector) | find_if(...) | what?

if we propagate an iterator at that point we've lost the container. So in that world. the let form seems necessary and if we have let why have it take a sender<view> rather than a view. What if we do want to perform a find on a data structure that is being generated, though?

Kirk suggested it should return some pair of a contained and an iterator, but then of course there are questions about correctness of the iterator on return of the container.

include/unifex/find_if.hpp
LeeHowes: Yes, that's a lot simpler.

PR #142, opened 2020-06-23
Simple find_if using bulk_scheduler

LeeHowes:


include/unifex/find_if.hpp
```diff
@@ -24,63 +25,162 @@
...
+                      return;
+                    }
+                  }
+                  // If not found, return the very end value
+                  perChunkState[index] = end_it;
+                },
+                unifex::par
+              )
+            ),
+            [&perChunkState, end_it, &values...]() mutable -> std::tuple<Iterator, Values...> {
```
lewissbaker, line 169: It's unfortunate you have to pack the results into a tuple just so you can return them from a lambda, only to have to then pipe them through another 'unpack' adapter to get them into multiple set_value() args.

I wonder if this whole thing might be more simply implemented as a custom many_receiver that explicitly implements set_next/set_value/set_error/set_done, etc. rather than using transform() here.

LeeHowes: Yes, that would certainly be possible. It just would mean more work in custom code while I would expect that an unpack algorithm would be a reasonable generic addition.

include/unifex/find_if.hpp
```diff
@@ -105,6 +205,70 @@ struct _receiver<Receiver, Func, FuncPolicy>::type {
...
+      typename Predecessor::template value_types<concat_type_lists_unique_t, find_if_apply>::type;
+
+  template<typename InPlaceConstruct>
+  _operation_state(InPlaceConstruct f) : predOp_{f(*this)} {
+  }
+
+  ~_operation_state() noexcept {
+    if (!started_) {
+      innerOp_.destruct();
+    }
```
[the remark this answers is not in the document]
LeeHowes: Oh, you're right. It's not that it just needs inverting, this was based on the implementation of let.

The implementation here is a bit different, though, and predOp_ is just initialised inline rather than in a manual lifetime. innerOp is iniitalised and then immediately started and then the receiver functions destroy it anyway, so this would be a double free. So this is simply unnecessary in this implementation.

include/unifex/find_if.hpp
```diff
@@ -152,13 +313,21 @@ struct _sender<Predecessor, Func, FuncPolicy>::type {
...
+    return _operation_state<Predecessor, Receiver, Func, FuncPolicy>{
+      [&](_operation_state<Predecessor, Receiver, Func, FuncPolicy>& os){
+        return unifex::connect(
+          static_cast<Sender&&>(s).pred_,
+          receiver_type<remove_cvref_t<Receiver>>{
+            static_cast<Sender&&>(s).func_,
+            static_cast<Receiver&&>(r),
+            static_cast<Sender&&>(s).funcPolicy_,
+            os
+        });
```
lewissbaker, line 328: Normally I put the call to the nested connect() inside the operation-state constructor.
I found it a little harder to follow the control flow with the way you've factored things here - not because it's wrong, but probably more because it's different.

PR #143, opened 2020-06-25
Add cancellation support to bulk schedule

LeeHowes:
Adds bulk_with_stop_source which creates a stop_token and injects it into the bulk algorithm. Each set_next call has the token available to it and a call to request_stop may stop further progress of the bulk_schedule. Stopping the downstream token should have the same effect.


include/unifex/bulk_schedule.hpp
```diff
@@ -58,11 +61,19 @@ class _schedule_receiver<Integral, Receiver>::type {
             #pragma loop(ivdep)
 #endif
             for (Integral i(0); i < count_; ++i) {
+                if(stop_token.stop_requested()) {
```
LeeHowes, line 64: This breaks vectorisation. I'm open to not interrupting in vectorised form. It's optimistic anyway.

It's still an open question precisely what the contract should be. Should set_next calls be started in order and cancellation only happen before they are started?

test/bulk_schedule_test.cpp
```diff
@@ -48,3 +49,33 @@ TEST(bulk, bulk_transform) {
...
+            unifex::bulk_transform(
+                unifex::bulk_with_stop_source(
+                    unifex::bulk_schedule(sched, count)),
+                [&](std::size_t index, auto& cancel_future_operations) noexcept {
+                    // Stop after third index
+                    if(index == compare_index) {
+                        cancel_future_operations.request_stop();
+                    }
+                    output[index] = index;
+                }, unifex::par_unseq)));
```
LeeHowes, line 73: This should probably be seq to be technically correct given the ordering assumption the example makes.

include/unifex/bulk_schedule.hpp
lewissbaker: For the default implementation of bulk_schedule(), which is built on top of schedule(), the handling of cancellation before we start executing should be handled by the underlying schedule(), which would end up just calling set_done() instead of set_value() here.

As a start, we could have code-paths here for the `is_stop_never_possible_v` and `!st.stop_possible()` checks where we hoist the check for the negative case outside the loop.

For the case where we do want to check periodically for a request to stop, we could try a nested loop where the inner loop was vectorised and we only checked for stop once per outer loop iteration.

The number of iterations to be performed by the inner loop could be tuned for the architecture (eg. lane-width for target architecture) or could possibly be tuned by a property/CPO on the receiver. e.g. a `get_batch_size_hint(receiver)` CPO.

LeeHowes: > For the default implementation of bulk_schedule(), which is built on top of schedule(), the handling of cancellation before we start executing should be handled by the underlying schedule(), which would end up just calling set_done() instead of set_value() here.

Is your point that I don't need a stop_requested() query outside the loop? I'm not doing that here, though, so I'm not quite sure what you're asking for.

Of course, we can do all those optimisations. I would expect a good QOI implementation to do so.

I could hack something in here just to demonstrate a point with the nested loop, to show we don't have to check every iteration. Maybe use the interference size constant as a hint, or just hard code 16 as a fairly reasonable vectorisable size, given that it depends completely on the function anyway. Adding CPOs to do those queries seems a bigger step out of scope of this diff.

I can add the stop_possible queries. That's clearly useful information.

include/unifex/bulk_schedule.hpp
```diff
@@ -48,22 +55,53 @@ class _schedule_receiver<Integral, Receiver>::type {
...
+        auto stop_token = get_stop_token(receiver_);
+        const bool stop_possible = !is_stop_never_possible_v<decltype(stop_token)> && stop_token.stop_possible();
+
+        if(stop_possible) {
+            for (Integral chunk_start(0); chunk_start < count_; chunk_start += bulk_cancellation_chunk_size) {
+                if(stop_token.stop_requested()) {
+                    unifex::set_done(std::move(receiver_));
+                    return;
+                }
+                Integral chunk_end = std::min(chunk_start + bulk_cancellation_chunk_size, count_);
```
lewissbaker, line 67: Have you verified whether this computation here still allows the below loop to be unrolled and vectorised?

I wonder if we need to have separate "chunk" loops which use a constant number of iterations and a tail-loop that handles the remaining iterations?

e.g. maybe something like this?
```
const Integral chunksEnd = count_ - (count_ % bulk_cancellation_chunk_size);
Integral i(0);
while (i < chunksEnd) {
  if (stopToken.stop_requested()) {
    unifex::set_done(std::move(receiver_));
    return;
  }

  const auto chunkEnd = i + bulk_cancellation_chunk_size;
  if constexpr (isVectorisable) {
    // pragma(ivdep)
    for (; i < chunkEnd; ++i) {
      unifex::set_next(receiver_, Integral(i));
    }
  } else {
    // no pragmas
    for (; i < chunkEnd; ++i) {
      unifex::set_next(receiver_, Integral(i));
    }
  }
}

for (; i < count_; ++i) {
  unifex::set_next(receiver_, Integral(i));
}
```

include/unifex/bulk_schedule.hpp
```diff
@@ -15,16 +15,23 @@
...
 
+#include <algorithm>
+
 #include <unifex/scheduler_concepts.hpp>
 #include <unifex/sender_concepts.hpp>
 #include <unifex/receiver_concepts.hpp>
 #include <unifex/tag_invoke.hpp>
 #include <unifex/get_execution_policy.hpp>
 #include <unifex/execution_policy.hpp>
+#include <unifex/inplace_stop_token.hpp>
```
lewissbaker, line 26: This include is unnecessary.

include/unifex/bulk_with_stop_source.hpp
```diff
@@ -0,0 +1,231 @@
...
+    auto operator()(Source&& s) const
+        noexcept(std::is_nothrow_constructible_v<remove_cvref_t<Source>, Source>)
+        -> stop_source_sender<remove_cvref_t<Source>> {
+        return stop_source_sender<remove_cvref_t<Source>>{(Source&&)s};
+    }
+};
+
+} // namespace _bulk_with_stop_source
+
+inline constexpr _bulk_with_stop_source::_fn bulk_with_stop_source{};
```
lewissbaker, line 227: I have a feeling this algorithm could be made more general to be applicable to all senders, not just bulk senders.

e.g. Say a `let_stop_source(fn)` that returns a sender that constructs an inplace_stop_source and invokes `fn` with an lvalue reference to it, which is then responsible for returning a sender which produces the result of the expression. It then connects the returned sender with a receiver that has the inplace_stop_token corresponding to that inplace_stop_source and also hooks up the incoming receiver's stop token to also request stop on the inplace_stop_source.

This could then be combined with a bulk operation as follows.
```
let_stop_source([&](inplace_stop_source& stopSource) {
  return bulk_join(
    bulk_transform(
      bulk_schedule(n),
      [&](size_t i) {
        if (shouldStop(i)) {
          stopSource.request_stop();
        }
      }));
});
```

lewissbaker: We could even make this a little more generic by having, say, a `let_value_from(factory, fn)` that is like `let_value()` but that constructs the value using `factor()` instead of from another sender.
This would allow constructing in-place an immovable object like `inplace_stop_source`.

It's possible that my experiments on sender/receiver async cleanup with async rvo might directly yield the ability to use `let_value()` with a `generate(fn)` algorithm to achieve the same. But that's something for another day...

LeeHowes: I think you're right. I don't tend to think in terms of let but it does seem likely to be cleaner in this case. I should do that more often - putting values in the data stream isn't always the right approach once you have good structure.

include/unifex/bulk_schedule.hpp
LeeHowes: I have not, largely because I would probably lean towards only including the simple version of the code here without the optimisation, until we see production use cases that benefit. I would hope that the unrolling for vectorisation you describe though is the first thing the compiler tries - the original loop would have suffered this way as well.

include/unifex/bulk_with_stop_source.hpp
LeeHowes: I left it as the stop_source injecting version of let. We can make it more flexible later if necessary.

include/unifex/let_with_stop_source.hpp
```diff
@@ -0,0 +1,217 @@
...
+        op_(op), receiver_{std::forward<Receiver>(r)}
+    {}
+
+    template<typename... Values>
+    void set_next(Values&&... values) &
+        noexcept(is_nothrow_next_receiver_v<Receiver, Values...,  unifex::inplace_stop_source&>) {
+
+        // Forward, injecting stop_source
+        unifex::set_next(receiver_, std::forward<Values>(values)..., op_.stop_source_);
+    }
```
lewissbaker, line 64: Do we need to make this algorithm a many-sender to inject the stop-source via set_next()?

Can we instead just rely on the injection of the stop-source via the SuccessorFactory and make this algorithm a single-sender-only algorithm?

Usually, any algorithm that is injecting cancellation of a bulk operation will also control the reduce operation and so will be able to apply this algorithm outside, eg. the `bulk_join()`.

LeeHowes: No we don't. I don't think that code is used, I must have not deleted it when cleaning up.

include/unifex/let_with_stop_source.hpp
```diff
@@ -0,0 +1,217 @@
...
+struct _stop_source_operation_callback;
+template<typename SuccessorFactory, typename Receiver, typename Token>
+struct _stop_source_operation_callback {
+    _stop_source_operation_callback(operation<SuccessorFactory, Receiver>& op, Token&& token) :
+        op_{op}, token_{token} {}
+    void operator()() {
+        op_.stop_source_.request_stop();
+    }
+    operation<SuccessorFactory, Receiver>& op_;
+    Token token_;
```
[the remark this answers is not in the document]
LeeHowes: Leftover, presumably...

include/unifex/let_with_stop_source.hpp
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+private:
+    UNIFEX_NO_UNIQUE_ADDRESS SuccessorFactory func_;
+};
+
+template<typename SuccessorFactory, typename Receiver, typename Token>
+struct _stop_source_operation_callback;
+template<typename SuccessorFactory, typename Receiver, typename Token>
+struct _stop_source_operation_callback {
```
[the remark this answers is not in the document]
LeeHowes: Actually, it's worse than that. Somewhere in doing this I didn't make the change to chain the callbacks. Fixing both in the update.

LeeHowes:
@lewissbaker Documentation added and cleanup in response to comments.

test/bulk_schedule_test.cpp
```diff
@@ -48,3 +49,39 @@ TEST(bulk, bulk_transform) {
...
+                        }
+                        output[index] = index;
+                    }, unifex::seq));
+        }));
+
+    for (std::size_t i = 0; i <= compare_index; ++i) {
+        EXPECT_EQ(i, output[i]);
+    }
+    for (std::size_t i = compare_index+1; i < count; ++i) {
+        EXPECT_EQ(0, output[i]);
```
lewissbaker, line 85: I wonder whether we should provide this guarantee for `bulk_schedule()` that, at least for sequential execution policy, that the indices are delivered in ascending order.

Would a valid implementation of `bulk_schedule()` be allowed to deliver the indices in a different order which might break the assumptions of this test?

test/bulk_schedule_test.cpp
```diff
@@ -48,3 +49,39 @@ TEST(bulk, bulk_transform) {
...
+    unifex::single_thread_context ctx;
+    auto sched = ctx.get_scheduler();
+
+    const std::size_t count = 1000;
+
+    std::vector<int> output(count, 0);
+    // Cancel after two chunks
+    // For the serial implementation this will stop the third chunk onwards from
+    // being dispatched.
+    const std::size_t compare_index = unifex::bulk_cancellation_chunk_size*2 - 1;
```
lewissbaker, line 63: Is this `bulk_cancellation_chunk_size` something you expect to be part of the public interface of `bulk_schedule()`?

I imagine different schedulers might choose different strategies for `bulk_schedule()` which might have different behaviour for cancellation here.

LeeHowes: I expect so, yes. I wouldn't suggest standardising this, it should be implementation-defined.

Even the vectorisation length is implemetation-defined on analysis of the set_next chain.

test/bulk_schedule_test.cpp
LeeHowes: There are questions around this that we should discuss.

Delivering the indices in some order is one question. Launching the tasks in some order is another - which may or may not be tied to the index. I don't think it matters for the purposes of this test, it is definitely a question we should bring up for standardisation.

I don't know if there is an answer with the current bulk algorithms either, but to be fair there are many unanswered questions there.

include/unifex/let_with_stop_source.hpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    template(typename Error)
+        (requires is_error_receiver_v<Receiver, Error>)
+    void set_error(Error&& error) noexcept {
+        unifex::set_error(std::move(receiver_), (Error&&)error);
+    }
+
+    template(typename R = Receiver)
+        (requires is_done_receiver_v<Receiver>)
```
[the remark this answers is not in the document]
LeeHowes: I have no idea. It cannot hurt to change, though.

doc/api_reference.md
```diff
@@ -239,6 +240,34 @@ whole will complete with the result of the successor.
...
+let_with_stop_source(some_operation(),
+    [](unifex::inplace_stop_source& stop_source) {
+      return other_operation(stop_source);
+    });
+```
+
+If `request_stop` is called on `stop_source` this may cancel some or
+all of `other_operation`. By this mechanism `let_with_stop_source`
+provides the opportunity to cancel work, partially or completely,
+as part of some larger operation.
```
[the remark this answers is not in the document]
LeeHowes: The vagueness of not definitely cancelling was what I was going for. You are right that it certainly does request, though.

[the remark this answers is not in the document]
LeeHowes: Is the fact that requesting cancellation is not guaranteed to cancel something we assume? Your wording has dropped that.

Presumably it should be assumed generally, though, or we'd have to repeat the text everywhere.

include/unifex/let_with_stop_source.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+    template<typename SuccessorFactory2>
+    explicit type(SuccessorFactory2&& func) : func_((SuccessorFactory2&&)func)
+    {}
+
+    template(typename Self, typename Receiver)
+        (requires same_as<remove_cvref_t<Self>, type>)
+    friend auto tag_invoke(tag_t<unifex::connect>, Self&& self, Receiver&& r)
+        noexcept(
+            std::is_nothrow_constructible_v<SuccessorFactory, member_t<Self, SuccessorFactory>> &&
```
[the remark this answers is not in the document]
LeeHowes: I don't understand how I missed this. I think I have done so now.

include/unifex/let_with_stop_source.hpp
```diff
@@ -0,0 +1,217 @@
...
+};
+
+template<typename SuccessorFactory, typename Receiver>
+struct _stop_source_operation<SuccessorFactory, Receiver>::type {
+    type(SuccessorFactory&& func, Receiver&& r) :
+        stop_source_{},
+        stop_callback_(*this, unifex::get_stop_token(r)),
+        innerOp_(
+            unifex::connect(
+                func(stop_source_),
```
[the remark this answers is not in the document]
LeeHowes: I think I only partially responded to that comment and missed the other part. Done.

include/unifex/let_with_stop_source.hpp
```diff
@@ -0,0 +1,217 @@
...
+    void start() noexcept {
+        unifex::start(innerOp_);
+    }
+
+    UNIFEX_NO_UNIQUE_ADDRESS unifex::inplace_stop_source stop_source_;
+    UNIFEX_NO_UNIQUE_ADDRESS _stop_source_operation_callback<
+            SuccessorFactory, Receiver, remove_cvref_t<decltype(unifex::get_stop_token(std::declval<Receiver>()))>>
+        stop_callback_;
+    connect_result_t<
+        std::invoke_result_t<SuccessorFactory, unifex::inplace_stop_source&>,
```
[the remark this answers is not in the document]
LeeHowes: I'm confused about github. It isn't showing these comments in the conversation view. No wonder I missed them.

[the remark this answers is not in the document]
LeeHowes: Did you get your lvalues and rvalues the right way round here? I made a change.It may be right.

include/unifex/let_with_stop_source.hpp
```diff
@@ -123,7 +123,7 @@ class _stop_source_sender<SuccessorFactory>::type {
         (requires same_as<remove_cvref_t<Self>, type>)
     friend auto tag_invoke(tag_t<unifex::connect>, Self&& self, Receiver&& r)
         noexcept(
-            std::is_nothrow_constructible_v<SuccessorFactory, member_t<Self, SuccessorFactory>> &&
+            std::is_nothrow_invocable_v<SuccessorFactory, inplace_stop_source&> &&
```
lewissbaker, line 126: Sorry to nit, but I think this needs to be `is_nothrow_invocable_v<member_t<Self, SuccessorFactory>, inplace_stop_source&>` to match the value-category of the template-parameter passed to the operation-state.

include/unifex/let_with_stop_source.hpp
```diff
@@ -169,7 +169,7 @@ struct _stop_source_operation<SuccessorFactory, Receiver>::type {
     UNIFEX_NO_UNIQUE_ADDRESS unifex::inplace_stop_source stop_source_;
     UNIFEX_NO_UNIQUE_ADDRESS callback_type<_stop_source_operation_callback> stop_callback_;
     connect_result_t<
-        std::invoke_result_t<SuccessorFactory, unifex::inplace_stop_source&>,
+        std::invoke_result_t<SuccessorFactory&&, unifex::inplace_stop_source&>,
```
lewissbaker, line 172: The `&&`-qualification here should be unnecessary (both refer to an rvalue)

doc/api_reference.md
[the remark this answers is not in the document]
lewissbaker: Yes, I think some of the wording around whether it might or might not respond to cancellation in a timely manner is intrinsic to the topic of cancellation and probably doesn't need to be spelled out in every algorithm.

From the perspective of this algorithm, all it can do is signal a request to cancel the child operation. It doesn't have any control over what the child operation does in response to this request.

PR #148, opened 2020-07-13
avoid dangling reference bug in unifex::let

ericniebler:


include/unifex/let.hpp
```diff
@@ -125,19 +125,20 @@ struct _predecessor_receiver<Operation>::type {
...
   void set_value(Values&&... values) && noexcept {
     bool destroyedPredOp = false;
+    auto& op = op_;
     try {
       auto& valueTuple =
-          op_.values_.template get<decayed_tuple<Values...>>();
+          op.values_.template get<decayed_tuple<Values...>>();
       valueTuple.construct((Values &&) values...);
       destroyedPredOp = true;
-      op_.predOp_.destruct();
```
ericniebler, line 133: After this line, `this` is defunct, and so is `op_`. Cacheing a reference to the operation state in a local is the fix.

lewissbaker: Thanks for fixing this. We generally do this elsewhere but missed here for some reason.

PR #149, opened 2020-07-14
add reset and release to scope_guard

ericniebler:


include/unifex/scope_guard.hpp
```diff
@@ -16,27 +16,41 @@
...
-      std::is_nothrow_move_constructible_v<Func>)
+public:
+  scope_guard(Func&& func) noexcept
       : func_((Func &&) func) {}
 
   ~scope_guard() {
+    reset();
+  }
+
+  void dismiss() {
```
lewissbaker, line 40: noexcept?

include/unifex/scope_guard.hpp
```diff
@@ -16,27 +16,41 @@
...
+    dismissed_ = true;
+  }
+
+  void reset() {
```
lewissbaker, line 44: `noexcept`?

include/unifex/scope_guard.hpp
```diff
@@ -16,27 +16,41 @@
...
     static_assert(noexcept(((Func &&) func_)()));
-    ((Func &&) func_)();
+    if (!std::exchange(dismissed_, true)) {
+      ((Func &&) func_)();
+    }
   }
```
kirkshoop, line 49: should dismiss/reset/~scope_guard be marked noexcept?

include/unifex/scope_guard.hpp
lewissbaker: The name `reset()` doesn't really scream "run this now" to me when I read it at the call-site. I suspect there would be occasions where I would mistake `.reset()` for doing the same thing as `.dismiss()`.

ericniebler: Maybe `dismiss` should be `release` to more strongly evoke the `reset`/`release` of standard smart pointers.

include/unifex/scope_guard.hpp
ericniebler: Well, not the destructor. It's implicitly `noexcept`.

PR #150, opened 2020-07-14
Fixing lifetime issues of active union members

ericniebler:


include/unifex/let.hpp
```diff
@@ -124,45 +124,38 @@ struct _predecessor_receiver<Operation>::type {
...
+        op.values_.template destruct<decayed_tuple<Values...>>();
+      };
+      auto& succOp =
+          unifex::activate_from<successor_operation<Values...>>(op.succOp_, [&] {
+              return unifex::connect(
+                  std::apply(std::move(op.func_), valueTuple),
+                  successor_receiver<Operation, Values...>{op});
+            });
+      unifex::start(succOp);
+      destroyValues.release();
```
ericniebler, line 143: @kirkshoop, @lewissbaker: please take a hard look at this to make sure I preserved the logic.

kirkshoop: LGTM

kirkshoop:
I would like to make the activate and deactivate pattern safer. I do not like that there are no compile-time or runtime checks that the target is a member of a union.

perhaps something like

```
  union operations {
    manual_lifetime<connect_result_t<Predecessor, predecessor_receiver<operation>>> predOp_;
    typename predecessor_type::template
        value_types<manual_lifetime_union, successor_operation>
            succOp_;
  };
  manual_lifetime_named_union<operations> operations_;
```

Where manual_lifetime_named_union is a struct that is constrained to only accept a union and has a member that is that union. 

The activate/activate_from/deactivate could be members of manual_lifetime_named_union and do a runtime check to make sure that the address of the arg to activate `operations_.activate(operations.get().predOp_, ..)` is in the contained union.

Or `manual_lifetime_union_member<operations::predOp_> predOp{operations_};` could be a type with the activate/activate_from/deactivate member for predOp and only do compile-time checks.

Also activate/activate_from/deactivate would be members of manual_lifetime_union.

Thoughts?

ericniebler:
I think the idea has promise -- I would also like to explore ways to make this safer, with reasonable compile times, and without space or runtime overhead -- but I'm not sure we're there yet. I'd rather land this PR now to fix the UB, and noodle on the larger issue of unions and manual lifetime management.

PR #151, opened 2020-07-26
Find if cancellation

LeeHowes:
Add cancellation to parallel find_if to cancel unstarted work after result found.


include/unifex/find_if.hpp
```diff
@@ -24,63 +26,219 @@
...
+    // This version is two phase to avoid a non-trivial atomic in the middle.
+    // With more built-in algorithms it can be simplified:
+    //  * let_with to allocate non-movable state in the operation state.
+    //  * unpack to deal with tuple to pack conversion
+    // It could also be simplified by making more of the code custom, but I wanted
+    // to demonstrate reuse of internal algorithms to build something more compelex
+    // and cancellable.
+    template<typename Iterator, typename... Values>
+    auto operator()(
+        const unpack_receiver<Receiver>& receiver,
```
lewissbaker, line 146: Seems unusual to pass in the unpack_receiver when its only used for the scheduler.
Could we just pass in the scheduler here instead?

include/unifex/find_if.hpp
```diff
@@ -24,63 +26,219 @@
...
+                )
+              );
+            return
+              unifex::transform(
+                unifex::transform_done(
+                  std::move(bulk_phase),
+                  [&stopSource, &found_flag](){
+                    if(*found_flag == true) {
+                      // If the item was found, then continue as if not cancelled
+                      return just();
```
lewissbaker, line 218: How can this return both just() and just_done()?

Edit: Ah, I see what you did there... :s

include/unifex/just.hpp
```diff
@@ -42,8 +44,14 @@ template <typename Receiver, typename... Values>
 struct _op<Receiver, Values...>::type {
   UNIFEX_NO_UNIQUE_ADDRESS std::tuple<Values...> values_;
   UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+  bool done_;
 
   void start() & noexcept {
+    if(done_) {
```
lewissbaker, line 50: I'm not a fan of the change to make this a runtime decision about whether this holds a value or the one signal.

Better would be to build a `variant_sender<Senders...>` type that can hold a sender of one of several different types.

include/unifex/just.hpp
```diff
@@ -110,6 +122,18 @@ namespace _just_cpo {
 } // namespace _just_cpo
 using _just_cpo::just;
 
+namespace _just_done_cpo {
+  inline const struct just_done_fn {
+    template <typename... Values>
+    constexpr auto operator()(Values&&... values) const
```
lewissbaker, line 128: Why does `just_done()` accept values when it will never complete with a value?

include/unifex/find_if.hpp
LeeHowes: Yes, although as it is unused in the other overload there's marginal inefficiency there.

include/unifex/find_if.hpp
LeeHowes: Yes... that's a fair complaint :) I think there is value in just not needing the magic to have multiple variants though.

include/unifex/just.hpp
LeeHowes: Yes, indeed. Unfortunately each time we think we need another algorithm it's days of work to implement it. We can front load that work, or alternatively get something that works and tidy it later.

include/unifex/just.hpp
LeeHowes: Clunky agreed. It was the simplest hack to infer a matching type. I could have templated the CPO.

include/unifex/just.hpp
LeeHowes: Alternatively, as we know what we actually want to do for this, I could remove the just hack and have the cancellation always uncancel, with a comment and a TODO that we will fix with the variant sender later.

include/unifex/just.hpp
LeeHowes: Ok, it is no more. There is a TODO instead.

PR #152, opened 2020-08-05
Let with

LeeHowes:
let_with algorithm that allows allocation in-place into the operation state, allowing the operation state to store non-moveable types like std::atomic without heap allocation.

### DIFF
--- a/include/unifex/find_if.hpp
+++ b/include/unifex/find_if.hpp
@@ -17,6 +17,7 @@
 
 #include <unifex/config.hpp>
 #include <unifex/just.hpp>
+#include <unifex/let_with.hpp>
 #include <unifex/let_with_stop_source.hpp>
 #include <unifex/execution_policy.hpp>
 #include <unifex/receiver_concepts.hpp>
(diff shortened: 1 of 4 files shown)


include/unifex/let_with.hpp
```diff
@@ -0,0 +1,183 @@
...
+    {}
+
+    template(typename Self, typename Receiver)
+        (requires same_as<remove_cvref_t<Self>, type>)
+    friend auto tag_invoke(tag_t<unifex::connect>, Self&& self, Receiver&& r)
+        noexcept(
+            std::is_nothrow_invocable_v<
+                member_t<Self, SuccessorFactory>,
+                std::invoke_result_t<StateFactory>&> &&
+            std::is_nothrow_constructible_v<remove_cvref_t<Receiver>, Receiver>) {
```
lewissbaker, line 122: This should be:
```
noexcept(
  std::is_nothrow_invocable_v<member_t<Self, StateFactory>> &&
  std::is_nothrow_invocable_v<member_t<Self, SuccessorFactory>, invoke_result_t<member_t<Self, StateFactory>>&> &&
  std::is_nothrow_constructible_v<Receiver, Receiver2> &&
  std::is_nothrow_invocable_v<
    decltype(unifex::connect),
    invoke_result_t<member_t<Self, SuccessorFactory>, invoke_result_t<member_t<Self, StateFactory>>&>,
    let_with_Receiver<operation<StateFactory, SuccessorFactory, Receiver>, remove_cvref_t<Receiver>>>)
```

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,183 @@
...
+private:
+    UNIFEX_NO_UNIQUE_ADDRESS StateFactory state_factory_;
+    UNIFEX_NO_UNIQUE_ADDRESS SuccessorFactory func_;
+};
+
+
+template<typename StateFactory, typename SuccessorFactory, typename Receiver>
+struct _let_with_operation<StateFactory, SuccessorFactory, Receiver>::type {
+    type(StateFactory&& state_factory, SuccessorFactory&& func, Receiver&& r) :
+        state_{state_factory()},
```
lewissbaker, line 140: Should be:
```
state_(static_cast<StateFactory&&>(state_factory)())
```
to be consistent with value-category of invocation of `func` below.

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,183 @@
...
+        innerOp_(
+              unifex::connect(
+                static_cast<SuccessorFactory&&>(func)(state_),
+                let_with_receiver<
+                        operation<StateFactory, SuccessorFactory, Receiver>,
+                        remove_cvref_t<Receiver>>{
```
lewissbaker, line 146: The application of `remove_cvref_t<Receiver>` here seems redundant given that you've already applied remove_cvref_t in the `operation` type-alias above.

Can we just apply this in one place?
My preference is generally to require template parameter to operation-state type to be unqualified and to strip the qualifications explicitly in `connect()` implementation.

This generally requires the operation-state constructor to make the receiver parameter's type a template so that it can be perfectly forwarded to construct the unqualified receiver data-member.

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,183 @@
...
+                    *this,
+                    static_cast<Receiver&&>(r)})) {
+    }
+
+    void start() noexcept {
```
lewissbaker, line 151: Explicitly lvalue qualify?

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,183 @@
...
+        unifex::start(innerOp_);
+    }
+
+    std::invoke_result_t<StateFactory> state_;
+    connect_result_t<
+        std::invoke_result_t<SuccessorFactory, std::invoke_result_t<StateFactory>&>,
+        let_with_receiver<
+            operation<StateFactory, SuccessorFactory, Receiver>,
+            remove_cvref_t<Receiver>>>
```
lewissbaker, line 160: remove_cvref_t<Receiver> is redundant here.

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+
+template<typename Operation, typename Receiver>
+using let_with_receiver = typename _let_with_receiver<Operation, Receiver>::type;
+
+template<typename Operation, typename Receiver>
+class _let_with_receiver<Operation, Receiver>::type {
+public:
+    explicit type(Operation& op, Receiver&& r) :
+        op_(op), receiver_{std::forward<Receiver>(r)}
```
lewissbaker, line 54: You're not actually using `op_` in this class.

Do you really need to wrap the receiver?
Maybe you can get away with just passing the `Receiver` directly into the child operation's connect() method instead?

include/unifex/find_if.hpp
```diff
@@ -157,86 +158,90 @@ struct _receiver<Predecessor, Receiver, Func, FuncPolicy>::type {
...
         max_num_chunks : ((distance+min_chunk_size)/min_chunk_size);
       auto chunk_size = (distance+num_chunks)/num_chunks;
-      // Found flag on the heap
-      // TODO: We can store this in the operation state with a let_with algorithm
-      // as atomics are not moveable.
-      auto found = std::make_unique<std::atomic<bool>>(false);
+
+      // Found flag and vector that will be constructed in-place in the operation state
+      struct State {
+        std::atomic<bool> found_flag;
```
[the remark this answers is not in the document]
LeeHowes: Indeed, but that's a change to find_if outside of this pure let_with change. I'll leave that for the future.

include/unifex/find_if.hpp
```diff
@@ -157,86 +158,90 @@ struct _receiver<Predecessor, Receiver, Func, FuncPolicy>::type {
...
-                        stopSource.request_stop();
-                        return;
+                      for(auto it = chunk_begin_it; it != chunk_end_it; ++it) {
+                        if(std::invoke(func, *it, values...)) {
+                          // On success, store the value in the output array
+                          // and cancel future work.
+                          // This works on the assumption that bulk_schedule will launch
+                          // tasks (or at least, test for cancellation) in
+                          // iteration-space order, and hence only cancel future work,
+                          // to maintain the find-first property.
```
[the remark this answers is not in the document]
LeeHowes: "bulk_schedule must launch workers in order of index, and may only cancel launch of workers after all workers with a smaller index may be considered launched." or something like that.

We did try this for OpenCl though and people did not want to enforce this requirement, for flexibility, even though it's extremely useful for algorithm design and all the hardware at the time I believe did satisfy it.

include/unifex/let_with.hpp
LeeHowes: Can do. Let is missing this though.

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,183 @@
...
+        let_with_receiver<
+            operation<StateFactory, SuccessorFactory, Receiver>,
+            remove_cvref_t<Receiver>>>
+        innerOp_;
+};
+
+struct _fn {
+
+    template<typename StateFactory, typename SuccessorFactory>
+    auto operator()(StateFactory&& state_factory, SuccessorFactory&& successor_factory) const
```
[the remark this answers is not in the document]
LeeHowes: I think it is worth it. It removes the need for the inner let in the algorithm too.

It's a bit fiddly with the parameter pack magic to unpack a tuple. I'll see if I get get it to work.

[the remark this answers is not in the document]
LeeHowes: It's more fiddly even if we want the parameters to be in this order.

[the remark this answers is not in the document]
LeeHowes: Do you approve of either:
reversing the order:
```

let_with([](auto&x, auto& y) {
  return someSender(x, y);
},
f, g);
```
which isn't too crazy as this isn't a pred sender, but a bit inconsistent.


Require a tuple for more than one arg:
```
let_with(f, [](auto&x) {
  return someSender(x);
});

let_with(make_tuple(f, g), [](auto&x, auto& y) {
  return someSender(x, y);
});
```

Or have another suggestion for a non-terminal parameter pack?

One option might be to have a single pack and a helper function that packs it into a tuple, splits out the last element, swaps the two and carries on happily.

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+#include <unifex/type_list.hpp>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+
+namespace _let_with {
+
+template<typename StateFactory, typename InnerOp, typename Receiver>
+struct _let_with_operation {
```
ericniebler, line 33: In the interest of bringing down identifier length (it matters for compile times and for the length of diagnostics), I suggest you drop the `_let_with_` prefix on the types, since it is redundant with the `_let_with` namespace.

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+    class type;
+};
+
+template<typename StateFactory, typename SuccessorFactory>
+using let_with_sender = typename _let_with_sender<StateFactory, SuccessorFactory>::type;
+
+template<typename StateFactory, typename SuccessorFactory>
+class _let_with_sender<StateFactory, SuccessorFactory>::type {
+public:
+    using InnerOp = std::invoke_result_t<SuccessorFactory, std::invoke_result_t<StateFactory>&>;
```
ericniebler, line 52:
```suggestion
    using InnerOp = std::invoke_result_t<SuccessorFactory, callable_result_t<StateFactory>&>;
```

... since a nullary invocable must be a simple callable (not a member pointer).

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+    using error_types = typename InnerOp::template error_types<Variant>;
+
+    template<typename StateFactory2, typename SuccessorFactory2>
+    explicit type(StateFactory2&& stateFactory, SuccessorFactory2&& func) :
+        stateFactory_((StateFactory2&&)stateFactory),
+        func_((SuccessorFactory2&&)func)
+    {}
+
+    template(typename Self, typename Receiver)
+        (requires same_as<remove_cvref_t<Self>, type>)
```
ericniebler, line 67: What are the requirements on `Receiver`?

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+    friend auto tag_invoke(tag_t<unifex::connect>, Self&& self, Receiver&& r)
+        noexcept(
+            std::is_nothrow_invocable_v<member_t<Self, StateFactory>> &&
```
ericniebler, line 70:
```suggestion
            is_nothrow_callable_v<member_t<Self, StateFactory>> &&
```

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+            std::is_nothrow_invocable_v<
+                member_t<Self, SuccessorFactory>,
+                std::invoke_result_t<member_t<Self, StateFactory>>&> &&
```
ericniebler, line 73:
```suggestion
                callable_result_t<member_t<Self, StateFactory>>&> &&
```

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+            std::is_nothrow_invocable_v<
+                decltype(unifex::connect),
```
ericniebler, line 75:
```suggestion
            is_nothrow_connectable_v<
```

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+                std::invoke_result_t<
+                    member_t<Self, SuccessorFactory>,
+                    std::invoke_result_t<member_t<Self, StateFactory>>&>,
```
ericniebler, line 78:
```suggestion
                    callable_result_t<member_t<Self, StateFactory>>&>,
```

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+              unifex::connect(
+                static_cast<SuccessorFactory&&>(func)(state_),
+                static_cast<Receiver&&>(r))) {
+    }
+
+    void start() & noexcept {
+        unifex::start(innerOp_);
+    }
+
+    std::invoke_result_t<StateFactory> state_;
```
ericniebler, line 108:
```suggestion
    callable_result_t<StateFactory> state_;
```

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+    connect_result_t<
+        std::invoke_result_t<SuccessorFactory, std::invoke_result_t<StateFactory>&>,
```
ericniebler, line 110:
```suggestion
        std::invoke_result_t<SuccessorFactory, callable_result_t<StateFactory>&>,
```

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,134 @@
...
+
+    template<typename StateFactory, typename SuccessorFactory>
+    auto operator()(StateFactory&& stateFactory, SuccessorFactory&& successor_factory) const
+        noexcept(std::is_nothrow_constructible_v<remove_cvref_t<SuccessorFactory>, SuccessorFactory> &&
+                 std::is_nothrow_constructible_v<remove_cvref_t<StateFactory>, StateFactory>)
+        -> let_with_sender<remove_cvref_t<StateFactory>, remove_cvref_t<SuccessorFactory>> {
+        return let_with_sender<remove_cvref_t<StateFactory>, remove_cvref_t<SuccessorFactory>>{
+            (StateFactory&&)stateFactory, (SuccessorFactory&&)successor_factory};
+    }
+};
```
ericniebler, line 125: It's best from an ADL-isolation point of view to put the CPO types into their own namespace. Since they get used in ADL calls, we don't want them accidentally dragging in any overloads that might happen to be in its associated namespace. (This matters more on MSVC which erroneously dumps hidden friend functions into the enclosing namespace.)

include/unifex/let_with.hpp
```diff
@@ -0,0 +1,141 @@
...
+                member_t<Self, SuccessorFactory>,
+                callable_result_t<member_t<Self, StateFactory>>&> &&
+            is_nothrow_connectable_v<
+                callable_result_t<
+                    member_t<Self, SuccessorFactory>,
+                    callable_result_t<member_t<Self, StateFactory>>&>,
+                remove_cvref_t<Receiver>>) {
+
+        return operation<
+                member_t<Self, StateFactory>, member_t<Self, SuccessorFactory>, Receiver>(
```
ericniebler, line 81: Does this need to change here now?

LeeHowes: Yes, I think it does.

ericniebler:
I think this is ready to go. @lewissbaker, has Lee addressed all of your concerns?

PR #155, opened 2020-08-08
Fix let_with_stop_source lifetimes

LeeHowes:
I think this is the correct behaviour for the let variants. The expectation of the user is that the lifetime of the successor function is until the end of the async operation (so that capture state can be treated as being alive as well as parameter state).

This also changes the behaviour to call the successor function as an l-value, because that is the consistent behaviour for an object that remains alive after the function runs.

### DIFF
--- a/include/unifex/let_with_stop_source.hpp
+++ b/include/unifex/let_with_stop_source.hpp
@@ -171,7 +171,7 @@ struct _stop_source_operation<SuccessorFactory, Receiver>::type {
     UNIFEX_NO_UNIQUE_ADDRESS unifex::inplace_stop_source stop_source_;
     UNIFEX_NO_UNIQUE_ADDRESS callback_type<_stop_source_operation_callback> stop_callback_;
     connect_result_t<
-        std::invoke_result_t<SuccessorFactory&&, unifex::inplace_stop_source&>,
+        callable_result_t<SuccessorFactory, unifex::inplace_stop_source&>,
         stop_source_receiver<operation<SuccessorFactory, Receiver>, remove_cvref_t<Receiver>>>
         innerOp_;
 };
--- a/include/unifex/let_with_stop_source.hpp
+++ b/include/unifex/let_with_stop_source.hpp
@@ -105,7 +105,7 @@ class _stop_source_sender<SuccessorFactory>::type {
     using additional_arguments = type_list<type_list<Values...>>;
 
 public:
-    using InnerOp = std::invoke_result_t<SuccessorFactory&&, inplace_stop_source&>;
+    using InnerOp = std::invoke_result_t<SuccessorFactory, inplace_stop_source&>;
 
     template<template<typename...> class Variant, template<typename...> class Tuple>
     using value_types = typename InnerOp::template value_types<Variant, Tuple>;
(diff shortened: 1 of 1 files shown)


include/unifex/let_with_stop_source.hpp
```diff
@@ -129,7 +129,7 @@ class _stop_source_sender<SuccessorFactory>::type {
     }
 
 private:
-    UNIFEX_NO_UNIQUE_ADDRESS SuccessorFactory func_;
+    UNIFEX_NO_UNIQUE_ADDRESS remove_cvref_t<SuccessorFactory> func_;
```
ericniebler, line 132: Is it correct to strip qualifiers here but not on line 108 or lines 124-127? It's probably better to strip qualifiers once in the definition of `stop_source_sender`:

```c++
template<typename SuccessorFactory>
using stop_source_sender =
    typename _stop_source_sender<remove_cvref_t<SuccessorFactory>>::type;
```

ericniebler:
Apart from the question about cvref qualifiers, this looks good to me.

include/unifex/let_with_stop_source.hpp
LeeHowes: Lewis has been tending to strip them as late as possible. 

You may be right that what's here is not correct, though - it is what I realised was missing for correctness to fix a bug.

LeeHowes: I think it is right here to store the func as a value, but to keep the cvrefness in the templates otherwise. I've updated it on that basis.

include/unifex/let_with_stop_source.hpp
```diff
@@ -105,7 +105,7 @@ class _stop_source_sender<SuccessorFactory>::type {
     using additional_arguments = type_list<type_list<Values...>>;
 
 public:
-    using InnerOp = std::invoke_result_t<SuccessorFactory, inplace_stop_source&>;
+    using InnerOp = std::invoke_result_t<SuccessorFactory&&, inplace_stop_source&>;
```
lewissbaker, line 108: Adding `&&` here should be unnecessary.

include/unifex/let_with_stop_source.hpp
lewissbaker: I have been tending to have the qualifiers stripped by the callsite that is instantiating the sender/receiver class rather than implicitly inside aliases or when defining data-members.

One of the reasons is to try to avoid instantiating senders/receivers multiple times when constructed with different value-categories. eg. if we are stripping qualifiers at the data-member level then invoking `let_with_stop_source(f)` and `let_with_stop_source(std::move(f))` will return two different sender-types even though all of the logic is identical except for the sender constructor which moves rather than copies.

But if we always return `_stop_source_sender<decay_t<SuccessorFactory>>` from `_fn::operator()` then you'll just get the one instantiation of the class and only instantiate the constructor multiple times.

include/unifex/let_with_stop_source.hpp
```diff
@@ -164,10 +165,11 @@ struct _stop_source_operation<SuccessorFactory, Receiver>::type {
 
     template<class F>
     using callback_type = typename stop_token_type_t<Receiver>::template callback_type<F>;
+    UNIFEX_NO_UNIQUE_ADDRESS remove_cvref_t<SuccessorFactory> func_;
     UNIFEX_NO_UNIQUE_ADDRESS unifex::inplace_stop_source stop_source_;
     UNIFEX_NO_UNIQUE_ADDRESS callback_type<_stop_source_operation_callback> stop_callback_;
     connect_result_t<
-        std::invoke_result_t<SuccessorFactory, unifex::inplace_stop_source&>,
+        std::invoke_result_t<SuccessorFactory&&, unifex::inplace_stop_source&>,
```
ericniebler, line 172: This can also be a `callable_result_t` since `SuccessorFactor` will never be a pointer to a member on an `inplace_stop_source`.

PR #156, opened 2020-08-08
Let with variadic

LeeHowes:
This is a variadic generalisation of let_with such that rather than one in-place initialising function, we can take a list of them.

eg:
```
let_with(
        [] { return 42; },
        [] { return std::atomic<int>{42}; },
        [&](int& val, std::atomic<int>& x) {
          ++x;
          printf("addressof x = %p, val = %i\n", (void*)&x, x.load());
          return async([&]() -> int {
            ++x;
            printf("successor tranform\n");
            printf("addressof x = %p, val = %i\n", (void*)&x, x.load());
            return x.load() + val;
          });
```

I also used it to update find_if to avoid a nesting of a let_with and a let.

let_with_stop_source is still separate because it relies on not only allocating a stop source but hooking the callback in, so it has custom logic.

### DIFF
--- a/doc/api_reference.md
+++ b/doc/api_reference.md
@@ -240,16 +240,17 @@ whole will complete with the result of the successor.
 If the predecessor completes with done/error then `func` is not invoked
 and the operation as a whole completes with that done/error signal.
 
-### `let_with(Invocable state_factory, Invocable func) -> Sender`
+### `let_with(Invocable state_factory1... Invocable state_factoryN, Invocable func) -> Sender`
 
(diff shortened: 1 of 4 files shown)


ericniebler:
@LeeHowes looks like this now has some merge conflicts.

LeeHowes:
I collapsed the branch, rebased and force pushed. I think it'll still have some MSVC build issues as ever.

AnujYamdagni (changes requested):

PR #157, opened 2020-08-09
Add Operator Composition

kirkshoop:
add support for composition by operator.

```cpp
  just()
    | transform([&]{ ++count; })
    | sync_wait();
```


include/unifex/compose_with_target.hpp
```diff
@@ -0,0 +1,71 @@
...
+    return unifex::tag_invoke(
+        _compose_with_target_fn{}, (Cpo &&) cpo, (ArgN &&) argN...);
+  }
+  template(typename Cpo, typename... ArgN)
+    (requires (!tag_invocable<_compose_with_target_fn, Cpo, ArgN...>))
+  auto operator()(Cpo&& cpo, ArgN&&... argN) const 
+      noexcept(
+          std::is_nothrow_constructible_v<std::tuple<remove_cvref_t<ArgN>...>, ArgN...> &&
+          std::is_nothrow_constructible_v<apply_target<Cpo, ArgN...>, std::tuple<remove_cvref_t<ArgN>...>>) 
+      -> apply_target<remove_cvref_t<Cpo>, remove_cvref_t<ArgN>...> {
```
ericniebler, line 64: Take a look at how the other CPOs define their return types using a `_result_t` helper and some lazy type computation. That prevents the compiler from instantiating two return types when doing overload resolution on `operator()`.

include/unifex/operator_composition.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <unifex/config.hpp>
+#include <unifex/std_concepts.hpp>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+  struct enable_operator_composition {
+    template (typename Target, typename TargetComposeFn)
+        (requires invocable<TargetComposeFn, Target>)
```
ericniebler, line 27:
```suggestion
        (requires callable<TargetComposeFn, Target>)
```

include/unifex/operator_composition.hpp
```diff
@@ -0,0 +1,37 @@
...
+    friend auto operator|(Target&& target, TargetComposeFn&& fn) 
+      noexcept(
+        std::is_nothrow_invocable_v<TargetComposeFn, Target>)
```
ericniebler, line 30:
```suggestion
        std::is_nothrow_callable_v<TargetComposeFn, Target>) 
```

include/unifex/operator_composition.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+  struct enable_operator_composition {
+    template (typename Target, typename TargetComposeFn)
+        (requires invocable<TargetComposeFn, Target>)
+    friend auto operator|(Target&& target, TargetComposeFn&& fn) 
+      noexcept(
+        std::is_nothrow_invocable_v<TargetComposeFn, Target>) 
+      -> std::invoke_result_t<TargetComposeFn, Target> {
```
ericniebler, line 31:
```suggestion
      -> callable_result_t<TargetComposeFn, Target> {
```

include/unifex/compose_with_target.hpp
```diff
@@ -0,0 +1,71 @@
...
+public:
+  template(typename Cpo, typename... ArgN)
+    (requires tag_invocable<_compose_with_target_fn, Cpo, ArgN...>)
+  auto operator()(Cpo&& cpo, ArgN&&... argN) const
+      noexcept(
+          is_nothrow_tag_invocable_v<_compose_with_target_fn, Cpo, ArgN...>)
+      -> tag_invoke_result_t<_compose_with_target_fn, Cpo, ArgN...> {
+    return unifex::tag_invoke(
+        _compose_with_target_fn{}, (Cpo &&) cpo, (ArgN &&) argN...);
+  }
```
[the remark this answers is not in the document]
kirkshoop: this is intended to allow customization for how arguments are packed and applied. tuple may not always be the best option.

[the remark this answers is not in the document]
ericniebler: Is that really sufficient justification for adding a customization point? I'm very stingy with customization points. I would much prefer to leave this one out. Pretty please?

include/unifex/compose_with_target.hpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+  };
+
+public:
+
+  template <typename Cpo, typename... ArgN>
+  auto operator()(Cpo&& cpo, ArgN&&... argN) const 
+      noexcept(
+          std::is_nothrow_constructible_v<std::tuple<remove_cvref_t<ArgN>...>, ArgN...> &&
+          std::is_nothrow_constructible_v<apply_target<remove_cvref_t<Cpo>, ArgN...>, std::tuple<remove_cvref_t<ArgN>...>>)
```
ericniebler, line 56: This is testing whether the `apply_target` is nothrow constructible from a `tuple`. I don't think that's what you meant.

kirkshoop: heh, that was an remanent of an earlier form. nice catch!

include/unifex/compose_with_target.hpp
```diff
@@ -0,0 +1,69 @@
...
+  }
+};
+
+inline const struct _fn {
+public:
+
+  template <typename Cpo, typename... ArgN>
+  auto operator()(Cpo&& cpo, ArgN&&... argN) const 
+      noexcept(
+          std::is_nothrow_constructible_v<apply_target<remove_cvref_t<Cpo>, ArgN...>, ArgN...>)
```
ericniebler, line 54: This will always be `false` because `apply_target` doesn't have a conditionally `noexcept` constructor.

include/unifex/compose_with_target.hpp
```diff
@@ -0,0 +1,69 @@
...
+    return apply_target<remove_cvref_t<Cpo>, ArgN...>{(ArgN &&) argN...};
+  }
+} compose_with_target{};
+
+} // namespace _compose_with_target
+
+using _compose_with_target::compose_with_target;
+
+template <typename Cpo, typename... ArgN>
+using compose_with_target_result_t = _compose_with_target::apply_target<remove_cvref_t<Cpo>, ArgN...>;
```
ericniebler, line 65: This is just `bind_back`, isn't it? Maybe rename it?

include/unifex/compose_with_target.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace _compose_with_target {
+
+template <typename Cpo, typename... ArgN>
+struct apply_target {
+private:
+  std::tuple<remove_cvref_t<ArgN>...> argN_;
+public:
+  template <typename... CN>
+  explicit apply_target(CN&&... cn) : argN_((CN&&) cn...) {}
```
ericniebler, line 35: `constexpr`? `noexcept`?

include/unifex/compose_with_target.hpp
kirkshoop: I renamed it to bind_back and added an operator(). the operator| now seems like an aberration, even though it is only only reason this function exists.

include/unifex/bind_back.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <tuple>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+
+namespace _bind_back {
+
+template <typename Cpo, typename... ArgN>
+struct apply_target {
```
ericniebler, line 30: Ah, I just realized, this needs ADL isolation.

PR #159, opened 2020-08-11
Fifo bulk experiment

LeeHowes:
Adds tag_invoke forwards to the bulk algorithms to:
 * Query what fifo context the algorithm will run on (if any)
 * To request eager execution - to run start early

The bulk scheduler uses this information to not call set_value, and the sequence algorithm is customised to eagerly enqueue subsequent operations if the fifo schedulers match.

The result is that this relies on FIFO ordering in manual_event_loop to chain work and skips the set_value calls entirely.

There are three extensions to this, that for now just add noise:
 * Add a tag_invoke hook for each bulk algorithm and an input fifo algorithm to wrap the bulk algorithm in a trivial type wrapper that adds the tag_invoke forwards (this might be a little noisy, but mechanical as there are not many bulk algorithms)
 * Have sequence's tag_invoke hook rely on that to create fifo_sequence based on an incoming fifo_sender, rather than explicitly calling a separate fifo_sequence algorithm as I did here. Clearly having this work with the standard algorithm names is the right high QOI solution, and tag_invoke allows that.
 * Move the set_value call in the bulk op into a separate task that is entirely separately enqueued. The reason for this is that in something like a CUDA runtime, we would probably not want to just conditionally not run it, but instead to not enqueue it and at the point where we do chain from one queue to another, enqueue it *then*, potentially as a CPU callback. This is a fairly simple extension of what I did here.


AnujYamdagni (changes requested):

LeeHowes:
Yes, we can close this.

PR #161, opened 2020-08-12
Make task.hpp compile

ispeters:
If task.hpp is the only Unifex header included in a translation unit, it
won't compile because there's no declaration of the sender_traits
template. This change fixes the problem by including sender_concepts.hpp.


ericniebler:
Thanks.

PR #162, opened 2020-08-12
fix on() to support current_scheduler

kirkshoop:
enables current scheduler to be used with `on()`

```cpp
with_query_value(
    when_all(
            on(transform(just(x), x_to_x2), current_scheduler()),
            . . . ),
    get_scheduler,
    sched
);
```

### DIFF
--- a/include/unifex/on.hpp
+++ b/include/unifex/on.hpp
@@ -42,7 +42,9 @@ namespace _on {
         (requires sender<Sender> AND scheduler<Scheduler> AND //
           (!tag_invocable<_fn, Sender, Scheduler>))
     auto operator()(Sender&& sender, Scheduler&& scheduler) const {
-      return sequence(schedule(scheduler),
+      auto sndr = schedule(scheduler);
+      return sequence(
+        std::move(sndr),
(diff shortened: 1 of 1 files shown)


include/unifex/on.hpp
```diff
@@ -42,10 +42,11 @@ namespace _on {
         (requires sender<Sender> AND scheduler<Scheduler> AND //
           (!tag_invocable<_fn, Sender, Scheduler>))
     auto operator()(Sender&& sender, Scheduler&& scheduler) const {
-      return with_query_value(
-          sequence(schedule(), (Sender&&)sender),
+      return sequence(schedule(scheduler),
```
ericniebler, line 45:
```suggestion
      auto sched = schedule(scheduler);
      return sequence(std::move(sched),
```

You're potentially moving from `scheduler` in the same expression. That's potential UB considering the indeterminate order of function argument evaluation.

kirkshoop: yes, I looked at that and did not remember indeterminate eval. :(

used your suggestion but with s/sched/sndr/

Issue #16, opened 2019-11-12
Add support for senders whose result-type is dependent on the receiver type


lewissbaker:
One of the patterns supported by senders is to allow contextual information from the calling site (receiver) to be injected into the callee (sender) using the receiver passed to the `connect()` operation.

For example, the `get_stop_token()` CPO can be called by a sender to query the stop-token contextual information from the receiver passed to `connect()`. Similarly, the `get_allocator(context)` CPO allows a sender to query the current allocator.

There will be cases where the result-type of a sender is dependent on the contextual information provided through the receiver. For example, a reduce algorithm that cumulates elements from a stream into a `std::vector` might produce a `std::vector<T, Allocator>` that uses the allocator obtained from  calling `get_allocator(context)`.

This means that the `sender::value_types` and `sender::error_types` type-aliases may not yet have enough information to determine the types of the values they will send. However, once we have access to a receiver (ie. in the `connect()` function or the operation-state object returned from it)  we can correctly calculate the value_types/error_types that will be sent.

Thus, to support using these types of senders with algorithms that need to know what types of values/errors will be sent we should probably allow/require defining the value_types/error_types on the operation-state type instead-of/as-well-as on the sender type.

PR #163, opened 2020-08-12
add current_scheduler to scheduler_concepts.hpp

ericniebler:


include/unifex/scheduler_concepts.hpp
```diff
@@ -337,6 +337,21 @@ namespace _now {
 } // namespace _now
 using _now::now;
 
+namespace _current {
+  inline constexpr struct _scheduler {
+    auto schedule() const noexcept {
```
kirkshoop, line 342: I was thinking that this should forward `tag_invoke()` rather than have member `schedule()`. This way it could support time and priority schedulers as well.

ericniebler: The meaning of `unifex::schedule()` is fixed. There is no dispatching with `tag_invoke` at this point. Supporting time schedulers is a good idea, though. I think a `schedule_after` member is the right way to support that, though.

It doesn't look like there is support yet for `unifex::schedule_at(TimePoint)` to be used with `current_scheduler`. I could add that.

I'm also tempted to move this functionality out of the customization points and into `current_scheduler` directly. The meaning of `unifex::schedule()` (with no arguments) is pretty unintuitive. But I suppose I can leave that for another time.

kirkshoop: I agree about `schedule()` - I am not sure about the `tag_invoke()` statement. I was thinking of a `tag_invoke()` overload that returns a sender that stores the cpo and arguments and uses them to call `tag_invoke()` in `connect()` with the target being `get_scheduler(receiver)`. I guess the challenge would be to constrain the `current_scheduler` `tag_invoke()` so that only schedule cpos were accepted. It might require some oddity like a common base tag for all the schedule cpos.

include/unifex/scheduler_concepts.hpp
```diff
@@ -301,7 +308,44 @@ namespace _schedule_at {
...
   } schedule_at {};
+
+  template <typename TimePoint>
+  class _sender<TimePoint>::type {
+  public:
+    template <template <typename...> class Variant, template <typename...> class Tuple>
+    using value_types = Variant<Tuple<>>;
+
+    template <template <typename...> class Variant>
+    using error_types = Variant<std::exception_ptr>;
```
lewissbaker, line 325: This sender is reporting that it will only ever invoke set_error() with `std::exception_ptr` but the ultimate scheduler's schedule_at() sender might end up sending some other error types.

We'd either need to static_assert that its error_types are a subset of the error-types reported here, or apply an adapter that remapped other error types into a std::exception_ptr.

We can leave this for another PR though, and hopefully it would be solved properly by #16.

include/unifex/scheduler_concepts.hpp
ericniebler: Yes, something like that would be possible. I've been meaning to extend unifex's `tag_invoke` with CPO categories like I first envisioned. We can use that, e.g., instead of the `is_receiver_cpo_v` trait also.

But that's work for another day. I'm going to merge this as-is for now.

lewissbaker: I was wondering whether new scheduler-based algorithms should explicitly add support for the `current_scheduler` scheduler when they are added, rather than adding them as members of the scheduler here.

Adding it for the scheduler methods defined in here is probably ok, but when extending, say, the I/O or file scheduler operations, they should probably add customisations for `current_scheduler()` within those CPOs rather than keep extending this class (which would create all sorts of circular dependencies).

PR #166, opened 2020-08-18
Suppress vectorization warning

ispeters:
When building with Clang with -Oz, the compiler will not vectorize the
loops in bulk_schedule.hpp that have requested vectorization.  This
results in a warning like so:

```
In file included from libunifex/test/bulk_schedule_test.cpp:17:
libunifex/include/unifex/bulk_schedule.hpp:96:17: error: loop not
vectorized: the optimizer was unable to perform the requested
transformation; the transformation might be disabled or specified as
part of an unsupported transformation ordering
[-Werror,-Wpass-failed=transform-warning]
                for (Integral i(0); i < count_; ++i) {
                ^
1 error generated.
```

This change suppresses the warning with a Clang-specific #pragma.

### DIFF
--- a/include/unifex/bulk_schedule.hpp
+++ b/include/unifex/bulk_schedule.hpp
@@ -65,8 +65,11 @@ public:
                 }
                 Integral chunk_end = std::min(chunk_start + static_cast<Integral>(bulk_cancellation_chunk_size), count_);
                 if constexpr (is_one_of_v<policy_t, unsequenced_policy, parallel_unsequenced_policy>) {
+UNIFEX_DIAGNOSTIC_PUSH
+
                     // Vectorisable version
 #if defined(__clang__)
+                    #pragma clang diagnostic ignored "-Wpass-failed"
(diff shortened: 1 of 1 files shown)


include/unifex/bulk_schedule.hpp
```diff
@@ -65,8 +65,11 @@ class _schedule_receiver<Integral, Receiver>::type {
                 }
                 Integral chunk_end = std::min(chunk_start + static_cast<Integral>(bulk_cancellation_chunk_size), count_);
                 if constexpr (is_one_of_v<policy_t, unsequenced_policy, parallel_unsequenced_policy>) {
+UNIFEX_DIAGNOSTIC_PUSH
+
                     // Vectorisable version
 #if defined(__clang__)
+                    #pragma clang diagnostic ignored "-Wpass-failed"
```
lewissbaker, line 72: Can you add a comment indicating why this diagnostic is being turned off?

PR #169, opened 2020-08-19
allow epoll and uring support to be forced

kirkshoop:
This allows `cmake -DUNIFEX_NO_EPOLL=1 ..` to disable epoll support in libunifex


cmake/unifex_flags.cmake
```diff
@@ -31,21 +31,40 @@ else()
   set(UNIFEX_NO_MEMORY_RESOURCE TRUE)
 endif()
 
+if(DEFINED UNIFEX_NO_LIBURING)
+  message(WARNING "[unifex warning]: forcing no_liburing=${UNIFEX_NO_LIBURING} !")
+elseif(DEFINED ENV{UNIFEX_NO_LIBURING})
+  message(WARNING "[unifex warning]: forcing no_liburing=$ENV{UNIFEX_NO_LIBURING} !")
```
ericniebler, line 37: Why is it a warning to do what the user asked? Maybe use `STATUS` instead.

cmake/unifex_flags.cmake
```diff
@@ -31,21 +31,40 @@ else()
...
+  set(UNIFEX_NO_LIBURING $ENV{UNIFEX_NO_LIBURING})
+  set(LIBURING_INCLUDE_DIRS $ENV{UNIFEX_LIBURING_INCLUDE_DIRS})
+  set(LIBURING_LIBRARIES $ENV{UNIFEX_LIBURING_LIBRARIES})
+else()
```
ericniebler, line 41: Indent everything in the `else()` branch, pls.

PR #170, opened 2020-08-20
Task cancellation

lewissbaker:
Adds support for stop-token-based cancellation to `task<T>` coroutines.

- Introduce new unifex::await_transform() CPO that coroutine types
  should forward to from their promise_type::await_transform() methods
  to allow types to customise how they should be awaited.
  Includes a default implementation for senders if not customised and
  no operator co_await() is provided.
- Introduce a new helper function unifex::connect_awaitable() that can
  be used as the implementation of connect() for awaitable types that
  want to be usable as senders.
- unifex::task<T> now directly implements the sender-concept and
  so you don't need to wrap it in awaitable_sender() when passing it
  to sender-based algorithms.
- Awaiting a task<T> now destroys the coroutine immediately upon
  resuming rather than at end of scope of the task object.
- The task<T> type now passes through the cancellation signal from
  the parent operation's get_stop_token() CPO to child operations
  co_awaited from within the coroutine. This should allow task-based
  coroutines to now be transparent to cancellation requests.

Also adds a new `inplace_stop_token_adapter` that makes it easier to adapt
an arbitrary incoming stop-token to an `inplace_stop_token` if required to
pass through type-erased interfaces, such as `task<T>` coroutines.


include/unifex/task.hpp
```diff
@@ -273,29 +194,58 @@ struct _task<T>::type {
...
+      assert(coro_);
       coro_.promise().continuation_ = h;
       coro_.promise().info_.emplace(
           continuation_info::from_continuation(h.promise()));
+      coro_.promise().stoken_ = stopTokenAdapter_.subscribe(get_stop_token(h.promise()));
       return coro_;
     }
-    decltype(auto) await_resume() {
+
+    std::optional<non_void_t<wrap_reference_t<decay_rvalue_t<T>>>> await_resume() {
```
ispeters, line 225: This is a surprising choice, I think.  It makes an uncancellable task inexpressible, right?

include/unifex/await_transform.hpp
```diff
@@ -0,0 +1,186 @@
...
+      break;
+    default:;
+    }
+  }
+
+  bool await_ready() const noexcept {
+    return false;
+  }
+
+  void await_suspend([[maybe_unused]] coro::coroutine_handle<Promise> h) noexcept {
```
ispeters, line 121: Typically you don't need `[[maybe_unused]]` if you don't name the argument.
```suggestion
  void await_suspend(coro::coroutine_handle<Promise>) noexcept {
```

include/unifex/await_transform.hpp
```diff
@@ -0,0 +1,186 @@
...
+    return unifex::tag_invoke(_fn{}, promise, (Value&&)value);
+  }
+
+  // Default implementation.
+  template(typename Promise, typename Value)
+    (requires (!tag_invocable<_fn, Promise&, Value>))
+  decltype(auto) operator()(Promise& promise, Value&& value) const {
+    if constexpr (detail::_awaitable<Value>) {
+      return (Value&&)value;
+    } else if constexpr (unifex::sender<Value>) {
```
ispeters, line 162: It's irking me that there are two `return (Value&&)value;` lines.  What do you think of collapsing this `if`?
```suggestion
    if constexpr (unifex::sender<Value> && !detail::_awaitable<Value>) {
```

include/unifex/connect_awaitable.hpp
```diff
@@ -0,0 +1,204 @@
...
+        }
+      };
+      return awaiter{(Func &&) func};
+    }
+
+    template(typename Value)
+      (requires callable<decltype(unifex::await_transform), promise_type&, Value>)
+    auto await_transform(Value&& value)
+        noexcept(is_nothrow_callable_v<decltype(unifex::await_transform), promise_type&, Value>)
+        -> callable_result_t<decltype(unifex::await_transform), promise_type&, Value> {
```
ispeters, line 94: My grasp of perfect forwarding is a bit fuzzy, but don't you need a `&&` in your `callable_result_t` here to match possibility of reference collapsing on the next line?
```suggestion
        -> callable_result_t<decltype(unifex::await_transform), promise_type&, Value&&> {
```

include/unifex/inplace_stop_token.hpp
```diff
@@ -157,6 +177,57 @@ class inplace_stop_callback final : private inplace_stop_callback_base {
...
+
+template<>
+class inplace_stop_token_adapter<unstoppable_token> {
+public:
+  inplace_stop_token subscribe(unstoppable_token) noexcept {
+    return inplace_stop_token{};
+  }
+
+  void unsubscribe() noexcept {}
+};
```
ispeters, line 229: If you gave `inplace_stop_token_adapter` a SFINAE hook, couldn't you handle all unstoppable tokens in one specialization?
```suggestion
template<typename StopToken>
class inplace_stop_token_adapter<StopToken, std::enable_if_t<is_never_stoppable<StopToken>>> {
public:
  inplace_stop_token subscribe(StopToken) noexcept {
    return inplace_stop_token{};
  }

  void unsubscribe() noexcept {}
};
```

include/unifex/await_transform.hpp
ericniebler: It's not the same, actually. When expressed as two separate `if constexpr` branches, if `detail::_awaitable<Value>` is true, the compiler never instantiates `unifex::sender<Value>`. This is important.

include/unifex/connect_awaitable.hpp
ericniebler: It's not needed. Either `Value` is an lvalue reference, in which case the `&&` collapses away, or else it `Value` is a non-reference type, in which case calling a function with a prvalue (e.g., an `int`) is the same (semantically) as calling it with an xvalue (e.g., and `int&&`). There is no observable difference.

include/unifex/inplace_stop_token.hpp
ericniebler: Are there multiple unstoppable tokens?

include/unifex/await_transform.hpp
ispeters: What do you think about dropping some comments in the code to explain that?  It seems like you folks are breaking ground on how to build template libraries while breaking ground on asynchrony; it might be valuable to document what you're learning.

include/unifex/inplace_stop_token.hpp
ispeters: The existence of a base template with specializations for `inplace_stop_token` and `unstoppable_token` suggests @lewissbaker is expecting non-Unifex tokens, and I assume there might be an unstoppable non-Unifex stop token.

include/unifex/await_transform.hpp
```diff
@@ -0,0 +1,186 @@
...
+namespace _await_transform {
+
+template <typename Promise, typename Sender>
+class _as_awaitable {
+  using value_t = single_value_result_t<Sender>;
+  enum class state { empty, value, exception, done };
+
+  class receiver {
+  public:
+    explicit receiver(_as_awaitable* op, coro::coroutine_handle<Promise> continuation) noexcept
```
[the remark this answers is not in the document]
ericniebler: It's a stylistic choice that the C++ Committee has decided on recently, and the justification is a little handwave-y, but it goes like this: are the constructor arguments semantically like the "value" of the resulting object? If so, the constructor should be implicit. (Think: a `pair<X,Y>` constructed from an `X` and a `Y`.) If not, the constructor should be explicit.

I'm not entirely convinced I buy this, but there it is.

include/unifex/await_transform.hpp
ericniebler: A comment to that effect would be a very nice addition. And we should probably write up somewhere how we use `tag_invoke` and extreme namespace isolation as a poor-man's "[deducing `this`](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0847r4.html) to dramatically speed up compile times. There's just so much to do and so little time. We're severely understaffed.

include/unifex/connect_awaitable.hpp
ericniebler: I mean, I can dream up cases where it matters. E.g., if `Value&&` is `int(&&)[4]`, then calling a function with `int(&&)[4]` is different than calling it with an `int[4]`. In the latter case, the array type necessarily decays to an `int*`. But we hedge against this in the definition of `callable_result_t`, which defensively slaps a `&&` on the arguments.

include/unifex/connect_awaitable.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+  inline const struct _fn {
+  private:
+    template <typename Awaitable, typename Receiver>
+    static auto connect_impl(Awaitable awaitable, Receiver receiver)
+        -> _await::sender_task<Receiver> {
+#if !UNIFEX_NO_EXCEPTIONS
+      std::exception_ptr ex;
+      try {
+#endif // !UNIFEX_NO_EXCEPTIONS
```
ericniebler, line 149: We should have CI runs that disable exceptions and RTTI.

include/unifex/inplace_stop_token.hpp
```diff
@@ -98,11 +100,17 @@ class inplace_stop_token {
 
   inplace_stop_token() noexcept : source_(nullptr) {}
 
+  inplace_stop_token(const inplace_stop_token& other) noexcept = default;
```
ericniebler, line 103: Why make it copyable?

include/unifex/task.hpp
```diff
@@ -214,6 +110,14 @@ struct _task<T>::type {
       state_ = state::exception;
     }
 
+    template(typename Value)
+      (requires is_callable_v<decltype(unifex::await_transform), promise_type&, Value>)
```
ericniebler, line 114:
```suggestion
      (requires callable<decltype(unifex::await_transform), promise_type&, Value>)
```

include/unifex/stop_when.hpp
```diff
@@ -92,18 +92,16 @@ namespace unifex
         return r.get_stop_token();
       }
 
-      template(typename CPO, typename... Args)
+      template(typename CPO)
           (requires is_receiver_query_cpo_v<CPO>)
       friend auto tag_invoke(
```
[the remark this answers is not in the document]
ericniebler: If I remember correctly, it was because the receiver query `tag_invoke` overloads were interfering with the `connect` `tag_invoke` overloads and causing spurious failures. The failures may have been msvc-specific, I can't quite recall.

include/unifex/connect_awaitable.hpp
```diff
@@ -0,0 +1,204 @@
...
+                    }
+                    return;
+                  }
+                }
+
+                assert(!result.has_value());
+
+                unifex::set_done(std::move(receiver));
+              };
+            }(co_await (Awaitable &&) awaitable);
```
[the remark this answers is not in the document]
lewissbaker: Yes, the `await_transform()` handles mapping this to `optional<unit>` for senders and the `task::_awaiter` handles this for `task`.

include/unifex/sender_concepts.hpp
```diff
@@ -402,9 +402,27 @@ struct single_value_type<> {
...
+
+template <>
+struct single_overload<> {
+private:
+  struct dummy {
+    using type = void;
+  };
+public:
+  using type = dummy;
+};
```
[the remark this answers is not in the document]
lewissbaker: I initially had `struct type` but I can't name the nested `::type` type-alias then because `type` is reserved for the constructor name.
I'll rename to something like `impl`.

include/unifex/task.hpp
lewissbaker: An uncancellable task will just never complete with `nullopt`.

I was going for a consistent shape to the result of `co_await` expressions - ie. every `co_await` in a coroutine completes in a `std::optional`.
This allows the `connect_awaitable()` implementation to always be able to unpack the result of the `co_await` and map an empty `optional` to `set_done()`.

include/unifex/inplace_stop_token.hpp
lewissbaker: This also crossed my mind, I was just being lazy at the time.
Thanks for the nudge. I'll do this.

include/unifex/inplace_stop_token.hpp
lewissbaker: The stop token concept should allow copyability.
It should have handle/pointer semantics to whatever the underlying state is.
`std::stop_token` is copyable.

include/unifex/stop_when.hpp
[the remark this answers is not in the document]
lewissbaker: In this instance I was trying to eliminate an ambiguous call to `visit_continuations()` where it was picking up the default `tag_invoke(_fn, const Receiver&, Func&&)` implementation for `visit_continuations()` and the receiver's forwarding tag_invoke overload `tag_invoke(CPO, const type& r, Args&&...)`. It may not have been necessary after changing the default implementation to not be defined as a `tag_invoke` overload.

But, yes, I removed `Args...` from most query-forwarding overloads to avoid these overloads being picked up by `connect()`. Keeping the arity strictly less than the arity of `connect()` means that these are not considered which avoided some unnecessary deep instantiations and some weird template errors.

include/unifex/await_transform.hpp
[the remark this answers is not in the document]
lewissbaker: The main difference comes with return-values:
```
receiver foo() {
  return {a, b}; // ill-formed if explicit
  return receiver{a, b}; // Need to explicitly name type
}
```
I find it helps force me to name the correct type and it looks more like a constructor-call.
Not strictly necessary in this case - we almost never need to return receivers from a function.

include/unifex/await_transform.hpp
lewissbaker: Comment has been added.

PR #172, opened 2020-08-20
Implement native unwind behaviour for coroutine cancellation

lewissbaker:
Coroutines now automatically unwind themselves if an awaited
operation completes with 'done'. Note that this 'unwind' process
is not catchable.
 
This signal will be transported up the stack until it hits a consumer
that does handle the 'done' signal. This typically means wrapping
the `task<T>` in a sender-algorithm that translates the done-signal
into some other signal that a coroutine can then handle.
eg. as an error/exception or as a value.
    
This means we don't have to unwrap `std::optional` values every
time we `co_await` something.
    
This relies on coroutine promise_types implementing a new
`unhandled_done()` method, similar to `unhandled_exception()`,
that handles the 'done' signal for that particular coroutine type.
    
The `task` type simply forwards the signal on to the awaiting
coroutine's `promise.unhandled_done()` method. This recursion
is terminated by the `connect_awaitable()` implementation which
implements `unhandled_done()` to forward on to the receiver's
`set_done()` implementation.


include/unifex/await_transform.hpp
```diff
@@ -73,7 +74,12 @@ class _as_awaitable {
     
     void set_done() && noexcept {
       op_->state_ = state::done;
-      continuation_.resume();
+      if constexpr (remove_cvref_t<Sender>::sends_done) {
```
ericniebler, line 77: We should stop looking for these things directly on the `Sender` type and instead use `sender_traits`.

include/unifex/connect_awaitable.hpp
```diff
@@ -160,24 +165,28 @@ namespace _await_cpo {
...
+                  constexpr size_t valueCount = Awaitable::template value_types<type_identity_t, count_types>::value;
+                  if constexpr (valueCount == 0) {
+                    unifex::set_value(std::move(receiver));
+                  } else if constexpr (valueCount == 1) {
+                    using value_type = typename Awaitable::template value_types<single_type, single_type>::type::type;
+                    unifex::set_value(std::move(receiver), static_cast<value_type&&>(result));
+                  } else {
+                    std::apply([&](auto&&... values) {
+                      unifex::set_value(std::move(receiver), static_cast<decltype(values)>(values)...);
+                    }, (result_type&&)result);
```
ericniebler, line 181: This lambda only really depends on the `Receiver` type, but due to how this code is structured it also depends incidentally on the `Awaitable` type. IMO, it's better to factor this out. Kill all the lambdas.

include/unifex/task.hpp
```diff
@@ -152,9 +161,13 @@ struct _task<T>::type {
       return p.stoken_;
     }
 
-    enum class state { empty, value, exception };
+    enum class state { empty, value, exception, done };
+
+    using done_callback = coro::coroutine_handle<>(void*) noexcept;
 
     coro::coroutine_handle<> continuation_;
+    done_callback* doneCallback_ = nullptr;
```
ericniebler, line 169: Did we ever consciously choose `camelCase` for variables? My preference is for `snake_case`. Obviously not a problem you should be solving here, though.

ericniebler:
@ GorNishanov should get a load of this.

include/unifex/await_transform.hpp
```diff
@@ -73,7 +74,12 @@ class _as_awaitable {
...
+        continuation_.promise().unhandled_done().resume();
+      } else {
+        // should be unreachable
+        std::terminate();
```
ispeters, line 81: If this is *actually* unreachable, aren't you converting a compile-time check into a runtime crash with this?  Why not lean on a `static_assert` instead?

include/unifex/task.hpp
```diff
@@ -216,13 +229,16 @@ struct _task<T>::type {
         coro::coroutine_handle<Promise> h) noexcept {
       assert(coro_);
       coro_.promise().continuation_ = h;
+      coro_.promise().doneCallback_ = [](void* p) noexcept {
+        return coro::coroutine_handle<Promise>::from_address(p).promise().unhandled_done();
+      };
```
ispeters, line 234: @ericniebler, is this another case where the lambda has an unnecessary implicit type dependency?  Looks to me like we really only need `Promise` here, but implicitly also depend on `T`.  Perhaps there should be a namespace-scope free function template that does this work; something like

```
namespace unifex::_task {
template <typename Promise>
coro::coroutine_handle<> task_done_callback(void* p) noexcept {
  return coro::coroutine_handle<Promise>::from_address(p).promise().unhandled_done();
}
```

Then this line could be
```suggestion
      coro_.promise().doneCallback_ = task_done_callback<Promise>;
```

ispeters:
This is beautiful!

include/unifex/task.hpp
lewissbaker: I have always been writing `camelCase` for non-public variable names.
Public data-members that form part of the interface are still `snake_case`.
Although I've noticed others have been using `snake_case` variable names, so we're not consistent.

include/unifex/await_transform.hpp
lewissbaker: Yes, I had meant to come back and replace this with use of a `UNIFEX_UNREACHABLE()` macro, but a `static_assert` would be better. I'm just not sure if we can guarantee this method will never be instantiated in all cases, even if it's not called. I guess at least we'd know if it did end up getting instantiated :)

include/unifex/await_transform.hpp
lewissbaker: I've fixed this - and uncommented the `sender_traits<S>::sends_done` definition so that it's usable ;)

include/unifex/connect_awaitable.hpp
lewissbaker: The lambda has been removed.

include/unifex/await_transform.hpp
lewissbaker: I've updated to a static_assert and it seems to still compile!

include/unifex/task.hpp
lewissbaker: I've added in the namespace-scope function.

PR #177, opened 2020-08-23
Add any sender of

ispeters:
Add a new _Sender_ called `any_sender_of<T...>` that uses `any_receiver_of<T...>` and `any_operation_state` to type-erase any _Sender_.

To support the new unit tests, I've also add a type called `unifex_test::mock_sender<Sigs...>`; it assumes that GMock is available.

### DIFF
--- a/include/unifex/any_sender_of.hpp
+++ b/include/unifex/any_sender_of.hpp
@@ -41,32 +41,26 @@ struct _rec_ref<Values...>::type {
   type(inplace_stop_token st, Op* op)
     : op_(op)
     , st_(st)
-    , set_value_fn_([](void* op, Values&&... values) noexcept {
-        static_cast<Op*>(op)->stopTokenAdapter_.unsubscribe();
-        UNIFEX_TRY {
-          unifex::set_value(
-              std::move(static_cast<Op*>(op)->rec_),
-              (Values&&) values...);
-        } UNIFEX_CATCH (...) {
-          unifex::set_error(
-              std::move(static_cast<Op*>(op)->rec_),
-              std::current_exception());
-        }
+    , set_value_fn_([](void* op, Values&&... values) {
+        static_cast<Op*>(op)->subscription_.unsubscribe();
+        unifex::set_value(
+            std::move(static_cast<Op*>(op)->rec_),
+            (Values&&) values...);
       })
     , set_error_fn_([](void* op, std::exception_ptr e) noexcept {
-        static_cast<Op*>(op)->stopTokenAdapter_.unsubscribe();
+        static_cast<Op*>(op)->subscription_.unsubscribe();
         unifex::set_error(
             std::move(static_cast<Op*>(op)->rec_),
             (std::exception_ptr&&) e);
       })
     , set_done_fn_([](void* op) noexcept {
-        static_cast<Op*>(op)->stopTokenAdapter_.unsubscribe();
+        static_cast<Op*>(op)->subscription_.unsubscribe();
         unifex::set_done(
             std::move(static_cast<Op*>(op)->rec_));
       })
   {}
 
-  void set_value(Values&&... values) && noexcept {
+  void set_value(Values&&... values) && {
     set_value_fn_(op_, (Values&&) values...);
   }
   void set_error(std::exception_ptr e) && noexcept {
@@ -83,7 +77,7 @@ private:
 
   void *op_;
   inplace_stop_token st_;
-  void (*set_value_fn_)(void*, Values&&...) noexcept;
+  void (*set_value_fn_)(void*, Values&&...);
   void (*set_error_fn_)(void*, std::exception_ptr) noexcept;
   void (*set_done_fn_)(void*) noexcept;
 };
@@ -134,6 +128,27 @@ struct _connect_fn<Values...>::type {
 template <typename... Values>
 inline constexpr typename _connect_fn<Values...>::type _connect{};
 
+template <typename StopToken>
+struct inplace_stop_token_adapter_subscription {
+  inplace_stop_token subscribe(StopToken stoken) noexcept {
+    isSubscribed_ = true;
+    return stopTokenAdapter_.subscribe(std::move(stoken));
+  }
+  void unsubscribe() noexcept {
+    if (isSubscribed_) {
+      isSubscribed_ = false;
+      stopTokenAdapter_.unsubscribe();
+    }
+  }
+  ~inplace_stop_token_adapter_subscription() {
+    unsubscribe();
+  }
+private:
+  bool isSubscribed_ = false;
+  UNIFEX_NO_UNIQUE_ADDRESS
+  inplace_stop_token_adapter<StopToken> stopTokenAdapter_{};
+};
+
 template <typename Receiver>
 struct _op_for {
   struct type;
@@ -144,8 +159,7 @@ struct _op_for<Receiver>::type {
   template <typename Fn>
   explicit type(Receiver r, Fn fn)
     : rec_((Receiver&&) r)
-    , stopTokenAdapter_{}
-    , state_{fn({stopTokenAdapter_.subscribe(unifex::get_stop_token(rec_)), this})}
+    , state_{fn({subscription_.subscribe(unifex::get_stop_token(rec_)), this})}
   {}
 
   void start() & noexcept {
@@ -154,8 +168,7 @@ struct _op_for<Receiver>::type {
 
   UNIFEX_NO_UNIQUE_ADDRESS
   Receiver rec_;
-  UNIFEX_NO_UNIQUE_ADDRESS
-  inplace_stop_token_adapter<stop_token_type_t<Receiver>> stopTokenAdapter_;
+  inplace_stop_token_adapter_subscription<stop_token_type_t<Receiver>> subscription_{};
   _operation_state state_;
 };
 


include/unifex/any_sender_of.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+template <typename Fun>
+rvo(Fun) -> rvo<Fun>;
+
+template <typename... Values>
+struct _any_connect_fn {
+  using type_erased_signature_t =
+      any_operation_state(this_&&, any_receiver_of<Values...>&&);
+
+  UNIFEX_TEMPLATE (typename Sender)
```
ericniebler, line 56: Since we're wrapped in `prologue.hpp` / `epilogue.hpp`, this can be:
```suggestion
  template (typename Sender)
```

include/unifex/any_sender_of.hpp
```diff
@@ -0,0 +1,103 @@
...
+    any_unique_t<
+        overload<void(this_&&, Values...)>(set_value),
+        overload<void(this_&&, std::exception_ptr) noexcept>(set_error),
+        overload<void(this_&&) noexcept>(set_done)>;
+
+using any_operation_state =
+    any_unique_t<
+        overload<void(this_&) noexcept>(start)>;
+
+namespace _any_sender {
```
ericniebler, line 36: In the interest of keeping identifier name length down:
```suggestion
namespace _any {
```

include/unifex/any_sender_of.hpp
```diff
@@ -0,0 +1,103 @@
...
+template <typename... Values>
+inline constexpr _any_connect_fn<Values...> _any_connect{};
+
+template <typename... Values>
+using _any_sender_of = any_unique_t<_any_connect<Values...>>;
+
+} // namespace _any_sender
+
+template <typename... Values>
+class any_sender_of : private _any_sender::_any_sender_of<Values...> {
```
ericniebler, line 82: We should use the ADL isolation technique here to prune the namespaces associated with `Values...` from the resulting `any_sender_of` type. This makes less work for the compiler when doing `tag_invoke` overload resolution. Can you reformulate as:

```c++
namespace _any {
  template <typename... Values>
  struct _sender {
    struct type;
  };

  template <typename... Values>
  struct _sender<Values...>::type {
    // as before
  };
} // namespace _any

template <typename... Values>
using any_sender_of = typename _any::_sender<Values>::type;
```

kirkshoop:
I would suggest a different pattern for assembling any_sender.

with a few changes we can punch through overloads of arbitrary receiver CPOs
see https://godbolt.org/z/c3zhf4
```cpp
using namespace unifex;

// For in-place constructing non-movable operation states.
// Relies on C++17's guaranteed copy elision.
template <typename Fun>
struct rvo {
    Fun fun_;
    operator callable_result_t<Fun>() && {
        return ((Fun&&) fun_)();
    }
};
template <typename Fun>
rvo(Fun) -> rvo<Fun>;

template <typename... Values>
using set_value_o = tag_t<overload<void(this_&&, Values...)>(set_value)>;
template <typename Error>
using set_error_o = tag_t<overload<void(this_&&, Error) noexcept>(set_error)>;

template <typename... Overloads>
using any_receiver_of =
    any_unique<
        Overloads...,
        set_error_o<std::exception_ptr>,
        tag_t<overload<void(this_&&) noexcept>(set_done)>>;

using any_operation_state =
    any_unique_t<
        overload<void(this_&) noexcept>(start)>;

template <typename... ReceiverOverloads>
struct _any_connect_fn {
    using type_erased_signature_t =
        any_operation_state(this_&&, any_receiver_of<ReceiverOverloads...>&&);

    UNIFEX_TEMPLATE (typename Sender)
        (requires sender_to<Sender, any_receiver_of<ReceiverOverloads...>>)
    friend any_operation_state
    tag_invoke(_any_connect_fn, Sender&& s, any_receiver_of<ReceiverOverloads...>&& r) {
        return any_operation_state{
            std::in_place_type<connect_result_t<Sender, any_receiver_of<ReceiverOverloads...>>>,
            rvo{[&]() { return connect((Sender&&) s, std::move(r)); }}
        };
    }

    UNIFEX_TEMPLATE (typename Self)
        (requires tag_invocable<_any_connect_fn, Self, any_receiver_of<ReceiverOverloads...>>)
    any_operation_state operator()(Self&& s, any_receiver_of<ReceiverOverloads...>&& r) const {
        return tag_invoke(*this, (Self&&) s, std::move(r));
    }
};
template <typename... ReceiverOverloads>
inline constexpr _any_connect_fn<ReceiverOverloads...> _any_connect{};

template <typename... ReceiverOverloads>
using _any_sender_of = any_unique_t<_any_connect<ReceiverOverloads...>>;

template <typename... ReceiverOverloads>
class any_sender_of : private _any_sender_of<ReceiverOverloads...>, sender_base {
public:
    using _any_sender_of<ReceiverOverloads...>::_any_sender_of;
    any_operation_state connect(any_receiver_of<ReceiverOverloads...> receiver) && {
        return _any_connect<ReceiverOverloads...>(
            static_cast<_any_sender_of<ReceiverOverloads...>&&>(*this),
            std::move(receiver));
    }
};

//====================
// test code
//====================

#include <cstdio>

struct sink {
    void set_value(int i) { std::printf("The answer is %d\n", i); }
    [[noreturn]]
    void set_error(std::exception_ptr) noexcept { std::terminate(); }
    void set_done() noexcept {}
};

int main() {
    any_sender_of<set_value_o<int>> s = just(42);
    auto op = connect(std::move(s), sink{});
    start(op); // OK, known to complete inline
}
```

include/unifex/any_sender_of.hpp
```diff
@@ -0,0 +1,103 @@
...
+public:
+  template <template <class...> class Variant, template <class...> class Tuple>
+  using value_types = Variant<Tuple<Values...>>;
+
+  template <template <class...> class Variant>
+  using error_types = Variant<std::exception_ptr>;
+
+  using _any_sender_of::_any_sender_of;
+
+  any_operation_state connect(any_receiver_of<Values...> receiver) && {
```
lewissbaker, line 94: I think we can possibly get rid of the need to heap-allocate the type-erased any_receiver_of here.

The idea would be to have the connect() method return an `any_operation_state_for<Receiver>` that stores the passed receiver object and then pass an `any_receiver_ref` that holds a type-erased pointer to this receiver.

This would require a few tweaks here, but would reduce it down to two heap-allocations (one for the sender and one for the operation-state, rather than three heap-allocations.

For a similar technique, see the `type_erased_stream` implementation which uses this to type-erase the receiver.

This customistaion would also provide an opportunity for hooking up the incoming receiver's stop_token up to an inplace_stop_token to be passed down to the implementation. eg. using the `inplace_stop_token_adapter<T>` to do the adaption.

include/unifex/any_sender_of.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  static constexpr bool sends_done = true;
+
+  using _sender_base<Values...>::_sender_base;
+
+  any_operation_state connect(any_receiver_of<Values...> receiver) && {
+    return _any::_any_connect<Values...>(
+        static_cast<_sender_base<Values...>&&>(*this),
+        std::move(receiver));
+  }
```
ispeters, line 100: If `receiver`'s type wasn't already erased in this context, you could know whether `set_value(receiver, values...)` is `noexcept` or not, right?  What do you think about defining `any_noexcept_receiver_of` to pair with `any_receiver_of`, and then extending `any_sender_of`'s vtable to support both `connect(any_receiver_of)` and `connect(any_noexcept_receiver_of)`?

I imagine the increased `noexcept`-ness would be a win, but it does generate a bunch of new code that couldn't be compiled out.  Could be a separate PR.

lewissbaker: This could be done, but it would end up double-instantiating the sender's connect() method for both the noexcept(true) and noexcept(false) cases - and then if we wanted to support multiple set_value-overloads, then for each combination of noexceptness.

If the goal of using noexcept would be to reduce binary size then I think this would probably be counter-productive.

Another approach would be to have the any_receiver_of always declare its set_value() overloads as noexcept and then if the concrete receiver's set_value is not noexcept then wrap the forwarding call to its set_value in a try/catch and call set_error() from the catch(...) clause in that case.

include/unifex/any_sender_of.hpp
ericniebler: @lewissbaker I've implemented the `any_receiver_ref` idea, but I haven't yet done the `inplace_stop_token` hookup. I may need your guidance here, but give me a chance to play with it first.

include/unifex/any_sender_of.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+
+template <typename... Values>
+using any_receiver_of =
+    any_unique_t<
+        overload<void(this_&&, Values...)>(set_value),
+        overload<void(this_&&, std::exception_ptr) noexcept>(set_error),
+        overload<void(this_&&) noexcept>(set_done)>;
```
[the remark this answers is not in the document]
ericniebler: Per your `any_receiver_ref` suggestion, I don't think we want `any_receiver_of` at all.

include/unifex/any_sender_of.hpp
ericniebler: > Another approach would be to have the any_receiver_of always declare its set_value() overloads as noexcept and then if the concrete receiver's set_value is not noexcept then wrap the forwarding call to its set_value in a try/catch and call set_error() from the catch(...) clause in that case.

I haven't done this. Maybe we can try it once we have `any_sender_of` in use internally and we can see what the impact is on code size.

include/unifex/any_sender_of.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+private:
+  friend inplace_stop_token tag_invoke(tag_t<get_stop_token>, const type& self) {
+    return self.st_;
+  }
+
+  void *op_;
+  void (*set_value_fn_)(void*, Values&&...);
+  void (*set_error_fn_)(void*, std::exception_ptr) noexcept;
+  void (*set_done_fn_)(void*) noexcept;
```
ericniebler, line 81: @lewissbaker @kirkshoop Would you prefer if I added an indirection and stored only a vtable ptr here? I think for 3 functions, this is better, but once we start adding type-erased receiver queries, a vtable is going to look attractive. We could even optimize for the common(?) case and leave the `set_value_fn_` pointer in the object but move `set_error_fn_` and `set_done_fn_` into a vtable.

lewissbaker: The any_unique implementation makes use of a vtable_holder_t type-alias that decides based on the number of vtable entries whether to store entries inline or externally.

kirkshoop: Additionally, I would not choose to pessimize set_done or set_error. There are cases where these are the hot path.

lewissbaker (approved):

PR #180, opened 2020-08-25
Remove use of 'virtual' for type-erasure

lewissbaker:
Updated the following types to directly use function-pointers instead of the `virtual` keyword
to implement the type-erasure necessary for maintaining queues of callbacks.

* `inplace_stop_token`
* `manual_event_loop`
* `thread_unsafe_event_loop`
* `timed_single_thread_context`
* `trampoline_scheduler`

This has the combined benefit of eliminating an extra indirection when invoking the
callback compared to a vtable-based solution as well as reducing binary size by
avoiding generation of RTTI for the operations being type-erased.

### DIFF
--- a/include/unifex/inplace_stop_token.hpp
+++ b/include/unifex/inplace_stop_token.hpp
@@ -16,9 +16,10 @@
 #pragma once
 
 #include <unifex/config.hpp>
-#include <unifex/spin_wait.hpp>
-#include <unifex/stop_token_concepts.hpp>
 #include <unifex/manual_lifetime.hpp>
+#include <unifex/spin_wait.hpp>
+#include <unifex/std_concepts.hpp>
+#include <unifex/stop_token_concepts.hpp>
 
 #include <atomic>
 #include <cassert>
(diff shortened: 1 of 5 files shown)


include/unifex/inplace_stop_token.hpp
```diff
@@ -150,9 +156,12 @@ inline inplace_stop_token inplace_stop_source::get_token() noexcept {
...
  public:
-  explicit inplace_stop_callback(inplace_stop_token token, F&& func) noexcept(
-      std::is_nothrow_move_constructible_v<F>)
-      : inplace_stop_callback_base(token.source_), func_((F &&) func) {
+  template(typename T)
+    (requires convertible_to<T, F>)
+  explicit inplace_stop_callback(inplace_stop_token token, T&& func) noexcept(
+      std::is_nothrow_constructible_v<F, T>)
+      : inplace_stop_callback_base(token.source_, &inplace_stop_callback::execute_impl)
+      , func_((T&&) func) {
```
ericniebler, line 164: How much of the code in the ctor depends on the template parameter? Can non-dependent bits be hoisted into the base?

lewissbaker: We could probably extract the body of the constructor to a helper method in the base class.

The main reason to have it inlined here would be to allow directly calling the function without indirection in the case that the stop-token was already ready. But this would be at the expense of code-size (more inlining).

lewissbaker: Fixed.

PR #182, opened 2020-08-26
ADL-isolate task<T>::awaiter<U> and _await_transform::_as_awaiter<T>

ericniebler:
Also, refactor `_task` for slightly shorter type names and refactor `_await_transform::_as_awaiter` so that all senders of the same value type share most of the implementation.

### DIFF
--- a/include/unifex/await_transform.hpp
+++ b/include/unifex/await_transform.hpp
@@ -92,7 +92,8 @@ protected:
       (requires (constructible_from<Value, Us...> ||
           (std::is_void_v<Value> && sizeof...(Us) == 0)))
     void set_value(Us&&... us) &&
-        noexcept(std::is_nothrow_constructible_v<Value, Us...>) {
+        noexcept(std::is_nothrow_constructible_v<Value, Us...> ||
+            std::is_void_v<Value>) {
       unifex::activate_union_member(result_->value_, (Us&&) us...);
       result_->state_ = state::value;
       continuation_.resume();
--- a/include/unifex/await_transform.hpp
+++ b/include/unifex/await_transform.hpp
@@ -41,8 +41,7 @@ namespace _await_transform {
 
 template <typename Promise, typename Sender>
 class _as_awaitable {
-  using value_t = typename sender_traits<remove_cvref_t<Sender>>::
-        template value_types<single_overload, single_value_type>::type::type;
+  using value_t = sender_single_value_return_type_t<remove_cvref_t<Sender>>;
   
   enum class state { empty, value, exception, done };
 
--- a/include/unifex/await_transform.hpp
+++ b/include/unifex/await_transform.hpp
@@ -37,74 +37,24 @@
 
 namespace unifex {
 
-namespace _await_transform {
+namespace _await_tfx {
+
+template <typename Promise, typename Value>
+struct _awaitable_base {
+  struct type;
+};
 
 template <typename Promise, typename Sender>
-class _as_awaitable {
-  using value_t = sender_single_value_return_type_t<remove_cvref_t<Sender>>;
-  
-  enum class state { empty, value, exception, done };
+struct _awaitable {
+  struct type;
+};
 
-  class receiver {
-  public:
-    explicit receiver(_as_awaitable* op, coro::coroutine_handle<Promise> continuation) noexcept
-    : op_(op)
-    , continuation_(continuation)
-    {}
+enum class state { empty, value, exception, done };
 
-    receiver(receiver&& r) noexcept
-    : op_(std::exchange(r.op_, nullptr))
-    , continuation_(std::exchange(r.continuation_, nullptr))
-    {}
-
-    template(class... Us)
-      (requires (constructible_from<value_t, Us...> || (std::is_void_v<value_t> && sizeof...(Us) == 0)))
-    void set_value(Us&&... us) &&
-        noexcept(std::is_nothrow_constructible_v<value_t, Us...>) {
-      unifex::activate_union_member(op_->value_, (Us&&) us...);
-      op_->state_ = state::value;
-      continuation_.resume();
-    }
-
-    void set_error(std::exception_ptr eptr) && noexcept {
-      unifex::activate_union_member(op_->exception_, std::move(eptr));
-      op_->state_ = state::exception;
-      continuation_.resume();
-    }
-    
-    void set_done() && noexcept {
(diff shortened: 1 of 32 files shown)


include/unifex/await_transform.hpp
```diff
@@ -37,17 +37,22 @@
 
 namespace unifex {
 
-namespace _await_transform {
+namespace _await_tfx {
 
 template <typename Promise, typename Sender>
-class _as_awaitable {
+struct _awaitable {
```
lewissbaker, line 48: I hadn't added ADL-isolation for `_as_awaitable` because I don't think it's used in any ADL-calls.
It's only used by the `operator co_await` mechanics of the compiler.

I guess it doesn't hurt to do this though, just in case...

include/unifex/await_transform.hpp
```diff
@@ -88,60 +120,60 @@ class _as_awaitable {
...
-      assert(state_ == state::exception);
-      std::rethrow_exception(exception_.get());
+      assert(result_.state_ == state::exception);
+      std::rethrow_exception(result_.exception_.get());
     }
   }
+};
+
+template <typename Sender>
+using _single_sender_value_type_t =
```
lewissbaker, line 151: This would be better named `_sender_single_value_type_t`.
The current naming reads to me like it's the value_type of a single-sender (as opposed to a many-sender).

Also, maybe move this out to `sender_concepts.hpp` along with `sender_value_types_t`, `sender_error_types_t` type-aliases?

include/unifex/await_transform.hpp
```diff
@@ -88,60 +120,60 @@ class _as_awaitable {
...
+    typename remove_cvref_t<Sender>::template
```
lewissbaker, line 152: You should be indirecting through `sender_traits` here.

include/unifex/task.hpp
```diff
@@ -124,25 +125,25 @@ struct _task<T>::type {
...
-        noexcept(is_nothrow_callable_v<decltype(unifex::await_transform), promise_type&, Value>)
-        -> callable_result_t<decltype(unifex::await_transform), promise_type&, Value> {
+        noexcept(is_nothrow_callable_v<decltype(unifex::await_transform), type&, Value>)
+        -> callable_result_t<decltype(unifex::await_transform), type&, Value> {
       return unifex::await_transform(*this, (Value&&)value);
     }
 
     template(typename Value)
-        (requires convertible_to<Value, T>)
+        (requires constructible_from<T, Value>)
```
lewissbaker, line 136: Why was this changed?
The use of `convertible_to` was intended to only allow returning things that were implicitly convertible to the return-type here (same as for normal functions).

include/unifex/task.hpp
```diff
@@ -180,76 +181,90 @@ struct _task<T>::type {
...
-    coro::coroutine_handle<promise_type> await_suspend(
-        coro::coroutine_handle<Promise> h) noexcept {
+    coro::coroutine_handle<ThisPromise> await_suspend(
+        coro::coroutine_handle<OtherPromise> h) noexcept {
       assert(coro_);
       coro_.promise().continuation_ = h;
-      coro_.promise().doneCallback_ = &forward_unhandled_done_callback<Promise>;
+      coro_.promise().doneCallback_ = &forward_unhandled_done_callback<OtherPromise>;
       coro_.promise().info_.emplace(
           continuation_info::from_continuation(h.promise()));
```
lewissbaker, line 213: I wonder how much cost the support for `visit_continuations` here has in terms of binary-size?

include/unifex/await_transform.hpp
ericniebler: Doesn't the compiler try to find `operator co_await` with ADL?

include/unifex/task.hpp
ericniebler: This function constructs a `T` from a `Value`, so `constructible_from<T, Value>` seems right to me. I think you're saying you want to constrain it _as if_ the continuation were a function that takes `T` and we're calling it with a `Value`. But that's not the reality.

include/unifex/task.hpp
ericniebler: You could put it in a branch and ask Ian to try it out.

include/unifex/await_transform.hpp
lewissbaker: Yes, I think you're right. It will attempt to find ADL-based overloads `operator co_await(_awaitable<Promise, Sender>)`.

include/unifex/task.hpp
lewissbaker: Say we have the following:
```
struct X {};
struct Y {
  explicit Y(X);
};
```
then with an ordinary function I will get a compile-error if I try to do:
```
Y f() {
  return X{}; // error: no implicit conversion from X to Y
  return {X{}}; // error: constructor is explicit need to name the type Y
  return Y{X{}}; // OK
}
```
and the idea is to do the same for coroutines:
```
task<Y> f() {
  co_return X{}; // error: no implicit conversion from X to Y
  co_return {X{}}; // error: Y constructor is explicit
  co_return Y{X{}}; // OK
}
```
Although with the current definition of `return_value()` it won't allow the second form.
(We need to add a non-template overload for this to work).

My understanding is that with your change here to use `constructible_from` instead of `convertible_to` is that this will allow the first form to be allowed (`Y` is indeed `constructible_from` `X`).

Do we need to require both `convertible_to` and `constructible_from` then?

I've just used `convertible_to` on its own in other places where I want to require implicit convertibility and if we do need to also use `constructible_from` then I'd like to understand better the ways in which `convertible_to` is insufficient.

ericniebler: Just because a type is convertible from another doesn't mean it is constructible from the other.

```C++
struct Y;
struct X {
    X() = default;
    explicit X(Y) = delete;
};
struct Y {
    operator X() const { return {}; }
};

static_assert(std::is_convertible_v<Y, X>);
static_assert(!std::is_constructible_v<X, Y>);
```

We should be requiring both.

include/unifex/connect_awaitable.hpp
```diff
@@ -163,7 +163,7 @@ namespace _await_cpo {
       try {
 #endif // !UNIFEX_NO_EXCEPTIONS
 
-        using result_type = single_value_result_t<Awaitable>;
+        using result_type = sender_single_value_result_t<Awaitable>;
```
lewissbaker, line 166: Also update the uses of `::value_types` in the lambda below?

PR #186, opened 2020-09-06
add an any_ref type that type-erases a reference

ericniebler:


ericniebler:
When reviewing this, hide whitespace changes for a better experience.

PR #191, opened 2020-09-08
support task<void>

### LINKED ISSUE #135: Support task<void>
Also, `task<>` should be an alias for `task<void>`.

ericniebler:
fixes #135

### DIFF
--- a/include/unifex/task.hpp
+++ b/include/unifex/task.hpp
@@ -39,6 +39,7 @@
 
 namespace unifex {
 namespace _task {
+using namespace _util;
 
 template <typename... Types>
 using _is_single_valued_tuple =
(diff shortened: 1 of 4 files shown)


ericniebler:
I used this as an opportunity to factor out of `task<T>::promise_type` the bits that don't depend on `T` in the hopes of bringing down the per-instantiation code size and compile times, so the diff is larger than it would be otherwise.

include/unifex/task.hpp
```diff
@@ -74,112 +75,109 @@ template <typename T>
...
-      reset_value();
-      unifex::activate_union_member(exception_, std::current_exception());
-      state_ = state::exception;
+      this->expected_.reset_value();
+      unifex::activate_union_member(this->expected_.exception_, std::current_exception());
+      this->expected_.state_ = _state::exception;
     }
 
-    template(typename Value)
+    template(typename Value = T)
```
lewissbaker, line 167: The `= T` default shouldn't be here.
The return-type of the coroutine has nothing to do with the type being co_awaited.

include/unifex/task.hpp
```diff
@@ -74,112 +75,109 @@ template <typename T>
...
-      return std::move(value_).get();
-    }
-
-    template <typename Func>
-    friend void
-    tag_invoke(tag_t<visit_continuations>, const type& p, Func&& func) {
-      if (p.info_) {
-        visit_continuations(*p.info_, (Func &&) func);
+      if (this->expected_.state_ == _state::exception) {
+        std::rethrow_exception(std::move(this->expected_.exception_).get());
```
lewissbaker, line 177: Does this need to be enclosed in `#ifdef` for `-fno-exceptions`?

lewissbaker (approved):

include/unifex/task.hpp
ericniebler: Eventually all the `rethrow_exception`s in the code will need to be replaced. The issue is, replaced with _what_? With `-fno-exceptions`, we can never reach this particular `rethrow_exception` until we add support for `co_await co_error(...)`, at which point we'll need to handle the error as we're doing for `unhandled_done`. That's a much bigger change, and should wait for a separate PR.

And until we replace _all_ the `rethrow_exception`s, replacing _this_ one -- when it can never get hit -- doesn't seem to buy us anything.

I'm inclined to leave it alone until we can do it right.

include/unifex/task.hpp
ericniebler: Ah, stray edit. Thanks. The one I added on `return_value` was the intentional one.

PR #195, opened 2020-09-12
shrink the executable footprint of any_unique instances

ericniebler:


include/unifex/this.hpp
```diff
@@ -24,95 +24,127 @@ namespace unifex {
...
+  template <typename, typename T>
   using apply = T;
 
-  template <typename Arg, typename T>
-  static T&& get(Arg&&, T& obj) {
+  template <typename T>
+  static T&& get(_ignore, T& obj) {
     return (T &&) obj;
   }
 };
```
lewissbaker, line 70: I assume these changes are to reduce the number of template instantiations here - which is good.
But do these changes actually have any impact on binary-size?
Are these methods something that we need to sprinkle UNIFEX_ALWAYS_INLINE macros on to force them to be inlined even under -Oz builds?

include/unifex/this.hpp
```diff
@@ -24,95 +24,127 @@ namespace unifex {
...
+  template <typename Arg>
+  static Arg&& get(Arg&& arg, _ignore) {
     return (Arg &&) arg;
   }
 };
 
 template <>
-struct replace_this<this_> {
-  template <typename Arg, typename T>
+struct _replace_this<this_> {
```
lewissbaker, line 62: Hmm, I wonder if we will also need a specialization for `const this_` at some stage...

include/unifex/any_unique.hpp
```diff
@@ -222,65 +208,85 @@ struct _with_type_erased_tag_invoke<
...
+struct _dispatch_cpo<CPO, NoExcept, Ret(Args...)> {
   struct type {
-    friend Ret tag_invoke(
-        _overload::base_cpo_t<CPO> cpo,
-        replace_this_t<Args, Derived>... args) noexcept(NoExcept) {
-      auto& wrapper = extract_this<Args...>{}(args...);
-      auto& wrapped = wrapper.value;
-      return std::move(cpo)(replace_this<Args>::get((Args &&) args, wrapped)...);
+    struct _result {
+      UNIFEX_NO_UNIQUE_ADDRESS _overload::base_cpo_t<CPO> cpo;
```
lewissbaker, line 222: Why do you need to strip back to base_cpo_t here?
Can you just invoke the CPO type inside operator() below?

include/unifex/any_unique.hpp
```diff
@@ -222,65 +208,85 @@ struct _with_type_erased_tag_invoke<
...
-    typename... Args>
-struct _with_forwarding_tag_invoke<
-    Derived,
-    CPO,
-    false,
-    Ret(Args...) noexcept>
-  : _with_forwarding_tag_invoke<Derived, CPO, true, Ret(Args...)> {
+template <typename... CPOs>
+struct _dispatch_fn : _dispatch_cpo<CPOs>::type... {
+  using _dispatch_cpo<CPOs>::type::operator()...;
```
lewissbaker, line 240: This looks like it could be problematic in cases where different overloads of the same CPO are provided here, as you'll have multiple base operator() implementations with the same signature that you're trying to import here - won't this lead to ambiguity?

include/unifex/any_unique.hpp
```diff
@@ -222,65 +208,85 @@ struct _with_type_erased_tag_invoke<
...
+    } else {
+      delete std::addressof(obj);
+    }
   }
-} deallocate {};
+};
+
+template <typename Concrete, typename Allocator, typename... CPOs>
+struct _concrete_impl {
+  struct type final {
```
lewissbaker, line 258: The purpose of this type is to allow us to hold the allocator along with the object.
However, this also means that we'll be instantiating each vtable entry for each combination of Concrete object type and Allocator, where we really only need to instantiate the 'deallocate' entry for this, and should be able to share the vtable entries for the normal CPOs that dispatch to using the object.

I wonder if we could have this type inherit from a base-class, which depends only on Concrete and defines tag_invoke overloads for CPOs, and then only mix in tag_invoke for the deallocate CPO here.

This would require, however, the assumption that the type-erased `void*` for the concrete type reinterpret_cast-able to both the base class and the derived class.

lewissbaker (commented):
@ericniebler Would you be able to add a summary of the changes you've made to the description and describe how each change was able to achieve a binary-size win? I didn't find it immediately obvious from reading the code how some of the changes were able to improve this.

ericniebler:
This PR was a rush job, and not all of this helps much, if at all. The big change was adding a separate code path for use when the user didn't specify an allocator. That accounts for the vast majority of the savings.

I can put up a separate diff with just that change, and be more methodical with the other changes.

include/unifex/any_unique.hpp
ericniebler: Yes, we can probably be smarter about that, but I'm not too concerned about it right now. I anticipate few people to use allocators when constructing `any_unique` instances, and this code isn't hit after my change.

include/unifex/this.hpp
ericniebler: Right, these changes aren't about binary size. They are drive-by cleanups.

include/unifex/any_unique.hpp
```diff
@@ -272,15 +259,54 @@ struct _with_forwarding_tag_invoke<
...
 
-  template(typename T)
-      (requires tag_invocable<deallocate_cpo, T&&>)
+  template <typename T>
   void operator()(T&& obj) const noexcept {
-    tag_invoke(deallocate_cpo{}, (T &&) obj);
+    if constexpr (tag_invocable<_deallocate_cpo, T>) {
+      tag_invoke(_deallocate_cpo{}, (T &&) obj);
+    } else {
+      delete std::addressof(obj);
```
lewissbaker, line 270: When is this alternative code-path going to be hit?

The `_concrete_impl` class below always defines a _deallocate_cpo tag_invoke overload so will always call the allocator version?

include/unifex/type_traits.hpp
```diff
@@ -110,18 +110,18 @@ struct unit {};
 using _unit::unit;
 
 template <bool B>
-struct _conditional {
+struct _if {
```
lewissbaker, line 113: Nice symbol size win ;)

include/unifex/any_unique.hpp
```diff
@@ -334,71 +360,36 @@ class _byval<CPOs...>::type
 
   template <typename Concrete, typename... Args>
   explicit type(std::in_place_type_t<Concrete> tag, Args&&... args)
-    : type(
-          std::allocator_arg,
-          std::allocator<unsigned char>{},
-          tag,
-          (Args &&) args...) {}
+    : impl_(new Concrete((Args&&) args...))
+    , vtable_(vtable_holder_t::template create<Concrete>()) {}
```
lewissbaker, line 364: Ah, I see you're bypassing the allocator version here.

lewissbaker (approved):
Changes all look good, although I'm still left wondering which changes had binary-size impacts.
(other than skipping the allocator).

The rest of the changes are good compile-time changes that reduce template instantiations so worthwhile on that basis.

include/unifex/any_unique.hpp
ericniebler: When you don't construct the `unique_any` with an allocator, I don't create a `_concrete_impl`. I `new` a `T` directly, which doesn't hook the `_deallocate_cpo`.

include/unifex/type_traits.hpp
ericniebler: Thank you very much.

include/unifex/any_unique.hpp
ericniebler: Yep.

ericniebler:
I would have to dig to see if anything besides the allocator changes contributed to LTO'ed code size reduction. I was measuring size of .o files, so some changes were possibly only impacting symbol length. But shorter symbol names and fewer template instantiations are generally a good thing, so I've kept them and not worried too much about it.

include/unifex/any_unique.hpp
lewissbaker: Once this is merged I want to take a look at a godbolt of side-by-side assembly for any_unique constructed with no allocator and for one constructed with `std::allocator` (which should have the same semantics).
Would be enlightening to see where the extra code-size is going.

PR #197, opened 2020-09-29
adds just_error(E)

kirkshoop:
`just_error(E)` produces a sender that calls `set_error(receiver, E)`

### DIFF
--- a/include/unifex/just.hpp
+++ b/include/unifex/just.hpp
@@ -79,7 +79,8 @@ class _sender<Values...>::type {
   static constexpr bool sends_done = false;
 
   template(typename... Values2)
-    (requires (sizeof...(Values2) == sizeof...(Values)))
+    (requires (sizeof...(Values2) == sizeof...(Values)) &&
+      constructible_from<std::tuple<Values...>, Values2...>)
   explicit type(std::in_place_t, Values2&&... values)
     noexcept(std::is_nothrow_constructible_v<std::tuple<Values...>, Values2...>)
     : values_((Values2 &&) values...) {}
@@ -104,8 +105,8 @@ namespace _just_cpo {
   inline const struct just_fn {
     template <typename... Values>
     constexpr auto operator()(Values&&... values) const
-      noexcept(std::is_nothrow_constructible_v<_just::sender<Values...>, Values...>)
-      -> _just::sender<std::decay_t<Values>...> {
+      noexcept(std::is_nothrow_constructible_v<_just::sender<Values...>, std::in_place_t, Values...>)
+      -> _just::sender<Values...> {
       return _just::sender<Values...>{std::in_place, (Values&&)values...};
     }
   } just{};
--- a/include/unifex/just.hpp
+++ b/include/unifex/just.hpp
@@ -79,7 +79,7 @@ class _sender<Values...>::type {
   static constexpr bool sends_done = false;
 
   template(typename... Values2)
-    (requires (sizeof...(Values2) == sizeof...(Values)) &&
+    (requires (sizeof...(Values2) == sizeof...(Values)) AND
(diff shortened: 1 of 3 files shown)


include/unifex/just_error.hpp
```diff
@@ -0,0 +1,108 @@
...
+  static constexpr bool sends_done = false;
+
+  template<typename Error2>
+  explicit type(std::in_place_t, Error2&& error)
+    noexcept(std::is_nothrow_constructible_v<Error, Error2>)
+    : error_((Error2 &&) error) {}
+
+  template(typename This, typename Receiver)
+      (requires same_as<remove_cvref_t<This>, type> AND
+        receiver<Receiver> AND
```
ericniebler, line 80:
```suggestion
        receiver<Receiver, Error> AND
```

ericniebler (changes requested):

include/unifex/just_error.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+template <typename Error>
+class _sender<Error>::type {
+  UNIFEX_NO_UNIQUE_ADDRESS Error error_;
+
+  public:
+  template <
+      template <typename...> class Variant,
+      template <typename...> class Tuple>
+  using value_types = Variant<Tuple<>>;
```
lewissbaker, line 66: This should be `using value_types = Variant<>` to indicate that no set_value() call will be invoked.
This is currently indicating that it _may_ invoke `set_value(receiver)`

include/unifex/just_error.hpp
```diff
@@ -0,0 +1,108 @@
...
+    return blocking_kind::always_inline;
+  }
+};
+} // namespace _just_error
+
+namespace _just_error_cpo {
+  inline const struct just_error_fn {
+    template <typename Error>
+    constexpr auto operator()(Error&& error) const
+      noexcept(std::is_nothrow_constructible_v<_just_error::sender<Error>, Error>)
```
lewissbaker, line 98: This will always result in `noexcept(false)` as the `_just_error::sender<Error>` only has one constructor that takes two args (std::in_place_t and an Error2).

Note, that even if this were changed to `std::is_nothrow_constructible_v<_just_error::sender<Error>, std::in_place_t, Error>` it would still not return the right result as the sender type does not constrain its constructor on the error argument.

This could be simplified to just `std::is_nothrow_constructible_v<std::decay_t<Error>, Error>`

include/unifex/just_error.hpp
```diff
@@ -0,0 +1,108 @@
...
+      -> _just_error::sender<std::decay_t<Error>> {
```
lewissbaker, line 99: This `std::decay_t` usage here is redundant with the usage in the `_just_error::sender` alias.

I generally prefer to avoid doing transformations in the type-alias, otherwise I'm always left wondering at the site of usage whether it's forgotten to apply decay/remove_cvref or not.

include/unifex/just_error.hpp
```diff
@@ -0,0 +1,108 @@
...
+      return _just_error::sender<Error>{std::in_place, (Error&&) error};
```
lewissbaker, line 100: The return object type here is spelled differently from the return-type in the signature.
Can we make them consistent?

include/unifex/just.hpp
```diff
@@ -79,7 +79,8 @@ class _sender<Values...>::type {
   static constexpr bool sends_done = false;
 
   template(typename... Values2)
-    (requires (sizeof...(Values2) == sizeof...(Values)))
+    (requires (sizeof...(Values2) == sizeof...(Values)) &&
```
lewissbaker, line 82: Can you use `AND` here instead of `&&` so that it does the lazy thing with the SFINAE versions of the macros?

lewissbaker (approved):

PR #198, opened 2020-09-29
add just_done

kirkshoop:
`just_done()` produces a sender that calls `set_done(receiver)`

### DIFF
--- a/test/transform_done_test.cpp
+++ b/test/transform_done_test.cpp
@@ -18,6 +18,7 @@
 #include <unifex/timed_single_thread_context.hpp>
 #include <unifex/just.hpp>
 #include <unifex/just_done.hpp>
+#include <unifex/on.hpp>
 #include <unifex/transform.hpp>
 #include <unifex/transform_done.hpp>
 #include <unifex/sequence.hpp>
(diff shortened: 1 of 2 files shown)


test/transform_done_test.cpp
```diff
@@ -57,6 +58,24 @@ TEST(TransformDone, Smoke) {
...
+  timed_single_thread_context context;
+
+  auto scheduler = context.get_scheduler();
+
+  int count = 0;
+
+  sequence(
+    schedule_after(scheduler, 200ms)
+      | transform_done(
+          []{ return just_done(); }),
```
lewissbaker, line 71: This seems like an odd way to test `just_done()` since it's being used to transform the 'done' signal into the 'done' signal.

Have you considered doing something like applying `just_done() | materialize() | sync_wait()` and then checking that the result is `std::optional<std::variant<std::tuple<tag_t<set_done>>>>`?

Or maybe do something similar by using `transform_done()` but this time to have it use `just_done()` as the source instead of the fallback?
```
std::optional<int> result =
  just_done()
  | transform_done([] { return just(42); })
  | sync_wait();
assert(result.has_value());
assert(result.value() == 42);
```

test/transform_done_test.cpp
```diff
@@ -57,6 +59,24 @@ TEST(TransformDone, Smoke) {
...
+  sequence(
+    just_done()
+      | transform_done(
+          []{ return just(); }) 
+      | on(scheduler),
+    lazy([&]{ ++count; }))
+    | sync_wait();
+
+  EXPECT_EQ(count, 1);
+}
```
lewissbaker, line 78: This test doesn't necessarily test that `just_done()` sends the done signal, since it would pass even if it sent `set_value()`.

Maybe something like:
```
{
  std::optional<unit> result = just_done() | sync_wait();
  assert(!result.has_value());
}
{
  std::optional<int> result =
    just_done()
    | transform_done([] { return just(42); })
    | sync_wait();
  assert(result.has_value());
  assert(result.value() == 42);
}
```

PR #199, opened 2020-10-02
fix unused var warnings

kirkshoop:
these were failing on a build that enables warnings as errors

### DIFF
--- a/test/materialize_test.cpp
+++ b/test/materialize_test.cpp
@@ -65,6 +65,7 @@ TEST(Materialize, Failure) {
           | materialize()
           | dematerialize()
           | sync_wait();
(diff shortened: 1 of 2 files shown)


test/materialize_test.cpp
```diff
@@ -65,6 +65,7 @@ TEST(Materialize, Failure) {
           | materialize()
           | dematerialize()
           | sync_wait();
+        EXPECT_FALSE(!!result); // should be unreachable - silences unused warning
```
lewissbaker, line 68: `[[maybe_unused]]` would be a more idiomatic way of doing this.

lewissbaker (approved):

PR #200, opened 2020-10-02
Add noexcept where it's missing in `any_unique`

ericniebler:
Additionally, move the try/catch overhead `std::make_exception_ptr` of out of headers and into the lib.

### DIFF
--- a/include/unifex/any_unique.hpp
+++ b/include/unifex/any_unique.hpp
@@ -30,12 +30,20 @@
 namespace unifex {
 namespace _any_unique {
 
+template <typename CPO>
+using base_cpo_t = _overload::base_cpo_t<CPO>;
+
 template <typename CPO, typename T, bool NoExcept, typename Ret, typename... Args>
 Ret _invoke(
-    _overload::base_cpo_t<CPO> cpo,
+    base_cpo_t<CPO> cpo,
     replace_this_with_void_ptr_t<Args>... args) noexcept(NoExcept) {
+  static_assert(!NoExcept || noexcept(extract_this<Args...>{}(args...)));
   void* thisPointer = extract_this<Args...>{}(args...);
-  return std::move(cpo)(
+  static_assert(!NoExcept || noexcept(Ret(((base_cpo_t<CPO>&&) cpo)(
+      replace_this<Args>::get(
+          (decltype(args)&&) args,
+          *static_cast<T*>(thisPointer))...))));
+  return ((base_cpo_t<CPO>&&) cpo)(
       replace_this<Args>::get(
           (decltype(args)&&) args,
           *static_cast<T*>(thisPointer))...);
@@ -47,8 +55,7 @@ struct vtable_entry;
 template <typename CPO, typename Ret, typename... Args>
 struct vtable_entry<CPO, Ret(Args...) noexcept> {
   using fn_t =
-      Ret(_overload::base_cpo_t<CPO>,
-          replace_this_with_void_ptr_t<Args>...) noexcept;
+      Ret(base_cpo_t<CPO>, replace_this_with_void_ptr_t<Args>...) noexcept;
 
   constexpr fn_t* get() const noexcept {
     return fn_;
@@ -69,8 +76,7 @@ private:
 template <typename CPO, typename Ret, typename... Args>
 struct vtable_entry<CPO, Ret(Args...)> {
   using fn_t =
-      Ret(_overload::base_cpo_t<CPO>,
-          replace_this_with_void_ptr_t<Args>...);
+      Ret(base_cpo_t<CPO>, replace_this_with_void_ptr_t<Args>...);
 
   constexpr fn_t* get() const noexcept {
     return fn_;
@@ -184,13 +190,21 @@ struct _with_type_erased_tag_invoke<
     NoExcept,
     Ret(Args...)> {
   struct type {
-    friend Ret tag_invoke(_overload::base_cpo_t<CPO> cpo, replace_this_t<Args, Derived>... args)
+    friend Ret tag_invoke(base_cpo_t<CPO> cpo, replace_this_t<Args, Derived>... args)
         noexcept(NoExcept) {
+      using cpo_t = base_cpo_t<CPO>;
+      static_assert(
+          !NoExcept || noexcept(extract_this<Args...>{}((decltype(args) &&) args...)));
       auto&& t = extract_this<Args...>{}((decltype(args) &&) args...);
+      static_assert(!NoExcept || noexcept(get_object_address(t)));
       void* objPtr = get_object_address(t);
+      static_assert(!NoExcept || noexcept(get_vtable(t)->template get<CPO>()));
       auto* fnPtr = get_vtable(t)->template get<CPO>();
+      static_assert(!NoExcept || noexcept(fnPtr(
+          (cpo_t&&) cpo,
+          replace_this<Args>::get((decltype(args) &&) args, objPtr)...)));
       return fnPtr(
-          std::move(cpo),
+          (cpo_t&&) cpo,
           replace_this<Args>::get((decltype(args) &&) args, objPtr)...);
     }
   };
@@ -237,10 +251,14 @@ struct _with_forwarding_tag_invoke<
     Ret(Args...)> {
   struct type {
     friend Ret tag_invoke(
-        _overload::base_cpo_t<CPO> cpo,
+        base_cpo_t<CPO> cpo,
         replace_this_t<Args, Derived>... args) noexcept(NoExcept) {
+      static_assert(!NoExcept || noexcept(extract_this<Args...>{}(args...)));
       auto& wrapper = extract_this<Args...>{}(args...);
       auto& wrapped = wrapper.value;
+      static_assert(
+          !NoExcept || noexcept(std::move(cpo)(
+              replace_this<Args>::get((Args &&) args, wrapped)...)));
       return std::move(cpo)(replace_this<Args>::get((Args &&) args, wrapped)...);
     }
   };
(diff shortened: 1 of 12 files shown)


include/unifex/exception.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <unifex/std_concepts.hpp>
+
+#include <unifex/detail/prologue.hpp>
+
+// Microsoft's STL has a fast implementation of std::make_exception_ptr.
+// So does libc++ when targetting Microsoft's ABI. On other stdlibs,
+// move as much exception machinery into a cpp file as possible to avoid
+// template code bloat.
+#if !defined(_LIBCPP_ABI_MICROSOFT) && !defined(_MSVC_STL_VERSION)
+#define UNIFEX_HAS_FAST_MAKE_EXCEPTION_PTR() 0
```
lewissbaker, line 31: The precedent for naming and form of this style of "is this thing available" macros is.
```
#define UNIFEX_HAVE_FILESYSTEM 1
```

Can we change this to:
```
#define UNIFEX_HAVE_FAST_MAKE_EXCEPTION_PTR 0
```

include/unifex/exception.hpp
```diff
@@ -0,0 +1,76 @@
...
+// If std::make_exeption_ptr() is slow, then move it into a cpp
+// file to generate less code elsewhere.
+struct _ref {
+  template (typename Obj)
+    (requires (!same_as<remove_cvref_t<Obj>, _ref>))
+  _ref(Obj&& obj) noexcept
+    : p_((void*) std::addressof(obj))
+    , throw_([](void* p) {
+        throw (Obj&&) *(std::add_pointer_t<Obj>) p;
+      })
```
lewissbaker, line 49: I wonder, would using a static function template rather than a lambda here give the linker a better chance at deduplicating the symbols compared to a lambda.

Also, is there any benefit in declaring these as `[[noreturn]]`?
I don't think so, because it's type-erased through a function-pointer and function-pointers can't be declared as `[[noreturn]]`, but maybe I'm missing something?

include/unifex/exception.hpp
```diff
@@ -0,0 +1,76 @@
...
+#if !UNIFEX_HAS_FAST_MAKE_EXCEPTION_PTR()
+// If std::make_exeption_ptr() is slow, then move it into a cpp
+// file to generate less code elsewhere.
+struct _ref {
+  template (typename Obj)
+    (requires (!same_as<remove_cvref_t<Obj>, _ref>))
+  _ref(Obj&& obj) noexcept
+    : p_((void*) std::addressof(obj))
+    , throw_([](void* p) {
+        throw (Obj&&) *(std::add_pointer_t<Obj>) p;
```
lewissbaker, line 48: This probably needs some `#ifdef` around it to support `-fno-exceptions`..

source/exception.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace unifex {
+namespace _except_ptr {
+#if !UNIFEX_HAS_FAST_MAKE_EXCEPTION_PTR()
+[[noreturn]] inline void _ref::rethrow() const {
+  throw_(p_);
+  UNIFEX_ASSUME_UNREACHABLE;
+}
+
+[[nodiscard]] std::exception_ptr _fn::operator()(_ref const e) const noexcept {
```
lewissbaker, line 28: Is the `[[nodiscard]]` attribute needed on the definition or only on the declaration?

include/unifex/exception.hpp
lewissbaker: Also, is this macro _really_ about it being "runtime fast" or is it more about reducing binary size?

ericniebler: My preference would be to change all the config macros to function-like macros like I'm doing here. [It's safer.](https://www.fluentcpp.com/2019/05/28/better-macros-better-flags/)

The issue is that sometimes we might do something like this:

```c++
#define SOME_MACRO 0
```

and then erroneously use it like

```c++
#ifdef SOME_MACRO
```

If it's a function-like macro, that misuse is immediately flagged.

include/unifex/exception.hpp
ericniebler: There's a bunch of places in the code where we need to do something special for `-fno-exceptions`. It's unclear to me what that thing should be here, and answering that question is not the point of this PR. Unifex doesn't support `-fno-exceptions` yet. TBD.

include/unifex/exception.hpp
ericniebler: I could replace the lambda with a static function template. Are you asking whether `rethrow()` should be declared `[[noreturn]]`?

source/exception.cpp
ericniebler: Don't know. I'll try and see.

include/unifex/any_unique.hpp
```diff
@@ -186,11 +192,22 @@ struct _with_type_erased_tag_invoke<
...
+      static_assert(!NoExcept || noexcept(fnPtr(
+          (cpo_t) (cpo_t&&) cpo,
+          replace_this<Args>::get((decltype(args) &&) args, objPtr)...)));
+      // The strange cast of the cpo to save the compiler from having to reserve
+      // space in the image for the cpo in case we pass it by reference through
+      // this function pointer:
       return fnPtr(
-          std::move(cpo),
+          (cpo_t) (cpo_t&&) cpo,
           replace_this<Args>::get((decltype(args) &&) args, objPtr)...);
```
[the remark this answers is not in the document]
lewissbaker: Are there places where we are passing the CPO by reference?
If so, we should change them to passing by-value.

I try to always pass CPO objects by-value precisely to avoid ODR-using the CPO object.

include/unifex/exception.hpp
lewissbaker: I was asking whether the lambda body should be declared as `[[nothrow]]`- I'm not sure if this is currently possible...
But making it a static function might make it possible.

e.g
```
private:
  template<typename Obj>
  [[noreturn]] void throw_impl(void* p) {
    throw static_cast<Obj&&>(*static_cast<std::add_pointer_t<Obj>>(p));
  }
public:
  template<typename Obj>
  ref(Obj&& obj) noexcept
  : p_(static_cast<void*>(std::addressof(obj)))
  , throw_(&throw_impl<Obj>) {}

  ...
};
```

But it seems you can't declare a function signature type as `[[noreturn]]` so the callsite is not going to be able to deduce this anyway (which you've handled with the UNIFEX_ASSUME_UNREACHABLE() macro).

include/unifex/exception.hpp
lewissbaker: I thought @ispeters had done a pass through the code to make `-fno-exceptions` work, at least for a subset of the code...

lewissbaker:
Were you still working on this @ericniebler?

ericniebler:
I will get back to this, I promise.

include/unifex/any_unique.hpp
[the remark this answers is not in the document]
ericniebler: There are no such places currently. I experimented with passing it by reference to see its impact on generated code. Since `operator()` ostensibly has an extra hidden `this` argument, it seemed to me that passing the CPO by reference might make it possible for the dispatch function into a simple tailcall. But really, the tailcall doesn't really matter because the compiler generally inlines the application of the CPO anyway. I can remove this.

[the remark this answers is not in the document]
lewissbaker: I generally thought compilers did a pretty good job of completely eliminating overhead of passing empty objects by-value. e.g. https://godbolt.org/z/bK9hqs

If you pass an empty object by-reference then you end up consuming an extra register to pass the address through.

include/unifex/exception.hpp
ericniebler: Yes, attributes aren't part of the type system, so putting one here would have no effect.

include/unifex/exception.hpp
ericniebler: Yeah, he did, but that work is incomplete.

include/unifex/any_unique.hpp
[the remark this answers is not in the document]
ericniebler: You misunderstand the point of my experiment. If the only thing a function (A) does is call another function (B) with the same arguments, then the compiler rewrites A as a `jmp` to B. My experiment here was in trying to jigger things so that the thunk gets turned into a `jmp` by putting the same arguments on the stack that will be used inside the thunk to do the dispatch. I hoped to save space that way.

That's not what was happening, so I can remove this part of the experiment.

include/unifex/any_unique.hpp
```diff
@@ -186,11 +192,22 @@ struct _with_type_erased_tag_invoke<
   struct type {
     friend Ret tag_invoke(_overload::base_cpo_t<CPO> cpo, replace_this_t<Args, Derived>... args)
```
lewissbaker, line 193: use `base_cpo_t` here too?

include/unifex/any_unique.hpp
```diff
@@ -186,11 +192,22 @@ struct _with_type_erased_tag_invoke<
...
+      static_assert(!NoExcept || noexcept(get_object_address(t)));
       void* objPtr = get_object_address(t);
+      static_assert(!NoExcept || noexcept(get_vtable(t)->template get<CPO>()));
       auto* fnPtr = get_vtable(t)->template get<CPO>();
+      static_assert(!NoExcept || noexcept(fnPtr(
+          (cpo_t&&) cpo,
+          replace_this<Args>::get((decltype(args) &&) args, objPtr)...)));
+      // The strange cast of the cpo to save the compiler from having to reserve
+      // space in the image for the cpo in case we pass it by reference through
+      // this function pointer:
```
lewissbaker, line 208: Remove this comment now?

lewissbaker (approved):

PR #203, opened 2020-10-12
remove unused variables_CXX_FOO_STDLIB and _CXX_FOO_STDLIB_LIB

andreasbuhr:
Usage of these variables was removed in March 2020
in e4b315870ff16d21a8cc50d745036dbb9552832c .


cmake/FindCoroutines.cmake
```diff
@@ -221,29 +217,21 @@ if(CXX_COROUTINES_HAVE_COROUTINES)
     ]] code @ONLY)
 
     # Try to compile a simple coroutines program without any compiler flags
-    set(CMAKE_REQUIRED_FLAGS "${_CXX_COROUTINES_STD17} ${_CXX_COROUTINES_STDLIB}")
-    set(CMAKE_REQUIRED_LIBRARIES "${_CXX_COROUTINES_STDLIB_LIB}")
+    set(CMAKE_REQUIRED_FLAGS "${_CXX_COROUTINES_STD17}")
```
lewissbaker, line 220: I wonder, do we still need this as well?
Should we be requiring the CMake config the user passes on the command-line to specify flags that enable coroutines rather than forcing them to be added here?

lewissbaker (approved):
Other than the question about whether this should be configuring vs detecting, the rest looks ok to me.

cmake/FindCoroutines.cmake
andreasbuhr: Correct me if I am wrong, but I think yes, we do need it. When using FindCoroutines.cmake, the user does not need to specify -std=c++17, we do it for her in https://github.com/facebookexperimental/libunifex/blob/5ce7f7eda5e573cee39932d4aa5fc9c86c69ab56/cmake/FindCoroutines.cmake#L234 But if the user does not specify c++17, we still need c++17 for the check_cxx_source_compiles tests, for example in https://github.com/facebookexperimental/libunifex/blob/5ce7f7eda5e573cee39932d4aa5fc9c86c69ab56/cmake/FindCoroutines.cmake#L228
Anyway, we do not interfere with the 'REAL' CMAKE_REQUIRED_FLAGS as we do cmake_push_check_state() and cmake_pop_check_state() in https://github.com/facebookexperimental/libunifex/blob/5ce7f7eda5e573cee39932d4aa5fc9c86c69ab56/cmake/FindCoroutines.cmake#L107 and https://github.com/facebookexperimental/libunifex/blob/5ce7f7eda5e573cee39932d4aa5fc9c86c69ab56/cmake/FindCoroutines.cmake#L247

lewissbaker: At the moment libunifex as a whole requires C++17 or later, so I would expect the user to have already configured their compiler for C++17 or C++20.

Regardless, cleaning up this can be left to another PR.

lewissbaker (approved):

cmake/FindCoroutines.cmake
andreasbuhr: I just created a pull request to clean this up:
https://github.com/facebookexperimental/libunifex/pull/204

PR #210, opened 2020-11-24
refactor reduce_stream to avoid constraint recursion; add missing constraints

ericniebler:


include/unifex/reduce_stream.hpp
```diff
@@ -309,22 +315,23 @@ struct _sender<StreamSender, State, ReducerFunc>::type {
...
+  using next_receiver = next_receiver<StreamSender, State, ReducerFunc, Receiver>;
+  template <typename Receiver>
+  using error_cleanup_receiver = error_cleanup_receiver<StreamSender, State, ReducerFunc, Receiver>;
   template <typename Receiver>
-  operation<Receiver> connect(Receiver&& receiver) const& {
+  using done_cleanup_receiver = done_cleanup_receiver<StreamSender, State, ReducerFunc, Receiver>;
+
+  template (typename Self, typename Receiver)
+      (requires same_as<remove_cvref_t<Self>, type> AND
+          //sender_to<next_sender_t<StreamSender>, next_receiver<Receiver>> AND
```
lewissbaker, line 327: Why is this commented out?
Can you leave a note in the code to explain why this constraint is not being checked?

include/unifex/reduce_stream.hpp
```diff
@@ -309,22 +315,23 @@ struct _sender<StreamSender, State, ReducerFunc>::type {
...
   template <typename Receiver>
-  operation<Receiver> connect(Receiver&& receiver) && {
-    return operation<Receiver>{
-        (StreamSender &&) stream_,
-        (State &&) initialState_,
-        (ReducerFunc &&) reducer_,
-        (Receiver &&) receiver};
-  }
-
+  using next_receiver = next_receiver<StreamSender, State, ReducerFunc, Receiver>;
```
lewissbaker, line 319: I thought it was ill-formed to use a name within class scope and have that name later change to refer to something else. Is this the one exception, where the only prior use of a name is in the RHS of a type-alias of the same name?

examples/fp_delegation.cpp
```diff
@@ -123,22 +123,27 @@ class delegating_sender {
...
-    using op = delegating_operation<
-        remove_cvref_t<decltype(unifex::connect(
-          unifex::schedule(unifex::get_scheduler(std::as_const(receiver))),
-          (Receiver&&)receiver))>,
-        LC>;
     if(context_->reserve()) {
       auto local_op = [&receiver, context = context_]() mutable {
         return LC{unifex::connect(
-          unifex::schedule(context->single_thread_context_.get_scheduler()),
+          unifex::schedule(unifex::get_scheduler(context->single_thread_context_)),
```
lewissbaker, line 146: This seems to be the only use of `get_scheduler(context)` which relies on this dispatching to the member `.get_scheduler()` method.

This change should probably be moved out to a separate PR so it can be considered on its own. It doesn't seem to be integral to the main thrust of this PR, which is to fix reduce_stream constraints.

If we do decide that this is the direction we want to go (ie. to allow using `unifex::get_scheduler()` to obtain the scheduler from an execution context) then the alternative to dispatching to member function would be to replace the `.get_scheduler()` methods on the execution context objects with a tag_invoke-based customisation of the existing `unifex::get_scheduler()` CPO and update all calls to use the CPO instead.

examples/fp_delegation.cpp
```diff
@@ -123,22 +123,27 @@ class delegating_sender {
...
+  using operation =
+      delegating_operation<
+          connect_result_t<
+              schedule_result_t<get_scheduler_result_t<Receiver>>,
+              Receiver>,
+          LocalContextType<connect_result_t<SchedSender, Receiver>>>;
+
   template <typename Receiver>
-  auto connect(Receiver&& receiver) {
+  operation<Receiver> connect(Receiver&& receiver) {
```
lewissbaker, line 138: It seems that attempting to use `auto` as the return-type of `connect()` always runs into trouble with recursive instantiation of some sort.

lewissbaker (changes requested):
Generally looks good.
A few minor suggestions and it would be good to separate out the `get_scheduler()` member function change to a separate PR.

include/unifex/scheduler_concepts.hpp
```diff
@@ -156,19 +159,59 @@ UNIFEX_CONCEPT //
...
+    template (typename SchedulerProvider)
+        (requires (!tag_invocable<_fn, SchedulerProvider>) AND
+            _with_member_get_schedule<SchedulerProvider>)
+    auto operator()(SchedulerProvider&& context) const noexcept
+        -> _result_t<SchedulerProvider> {
+      static_assert(
+          noexcept(static_cast<SchedulerProvider&&>(context).get_scheduler()));
+      static_assert(
+          scheduler<_member_get_scheduler_result_t<SchedulerProvider>>);
+      return static_cast<SchedulerProvider&&>(context).get_scheduler();
```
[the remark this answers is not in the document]
ericniebler: The reason is a little weak. I needed to name the type returned by `<context>.get_scheduler()` in fp_delegation.cpp, so I added the `get_scheduler_result_t` alias. Then I discovered that the `get_scheduler` CPO didn't dispatch to the `.get_scheduler()` member and was surprised, so I "fixed" it. I get the sense that you're not a fan.

include/unifex/reduce_stream.hpp
ericniebler: I commented it out because it is causing an error in never_stream_cancellation.cpp with concepts emulation. I forgot I commented it out. The problem remains and needs investigation. I'll dig some more.

include/unifex/reduce_stream.hpp
ericniebler: I don't _think_ it's ill-formed, but I know some compilers warn, so I'll rename it.

examples/fp_delegation.cpp
ericniebler: I can factor it out.

examples/fp_delegation.cpp
ericniebler: Yes. Every time we've done that it's resulted in hours of hair pulling tracking down instantiation problems. Let's never do that. And we really need to get in the habit of constraining `connect` at the very least.

ericniebler:
Bummer, I can't reproduce the MSVC error in my VM. I have no idea what the problem could be. :-(

include/unifex/reduce_stream.hpp
```diff
@@ -38,18 +38,33 @@
...
+struct _op {
+  struct type;
+};
+template <typename StreamSender, typename State, typename ReducerFunc, typename Receiver>
+using operation =
+    typename _op<
+        remove_cvref_t<StreamSender>,
+        remove_cvref_t<State>,
+        remove_cvref_t<ReducerFunc>,
+        remove_cvref_t<Receiver>>::type;
```
lewissbaker, line 51: I've recently been trying to move away from putting transformations like remove_cv_ref_t in the type-alias here as I find myself constantly wondering at the usage site of these sort of aliases whether it's forgotten to apply remove_cvref_t or not.

I would like to be consistent in our usage pattern for this kind of class as we use this ADL-isolation pattern everywhere. We can tackle that in another PR though.

ericniebler: Yes, I know I had promised to fix all that, but I've never had time. :-(

include/unifex/scheduler_concepts.hpp
```diff
@@ -180,22 +190,22 @@ struct _schedule::sender {
 
   static constexpr bool sends_done = true;
 
-  template(
-    typename Receiver,
-    typename Scheduler =
-      std::decay_t<callable_result_t<decltype(get_scheduler), const Receiver&>>,
-    typename ScheduleSender = callable_result_t<decltype(schedule), Scheduler&>)
```
lewissbaker, line 187: Removing this line was the key change needed to fix msvc in my experiments.
Not sure why it is particularly problematic...

include/unifex/scheduler_concepts.hpp
```diff
@@ -180,22 +190,22 @@ struct _schedule::sender {
...
+  template(typename Receiver)
     (requires receiver<Receiver>)
   friend auto tag_invoke(tag_t<connect>, sender, Receiver &&r)
-      -> connect_result_t<ScheduleSender, Receiver> {
+      -> connect_result_t<
+            schedule_result_t<std::decay_t<
+                get_scheduler_result_t<const remove_cvref_t<Receiver>&>>&>,
+            Receiver> {
     auto scheduler = get_scheduler(std::as_const(r));
     return connect(schedule(scheduler), (Receiver &&) r);
```
lewissbaker, line 201: We could add a `std::move(scheduler)` here and avoid the need for `std::decay_t` in the return-value type calculation.

include/unifex/sender_concepts.hpp
```diff
@@ -144,6 +133,23 @@ namespace detail {
...
+        typename (typename sender_traits<S>::_unspecialized)
+      ));
+  template <typename S>
+  UNIFEX_CONCEPT         //
+    _has_sender_traits = //
+      (!UNIFEX_FRAGMENT(detail::_not_has_sender_traits, S));
+#else
+  template <typename S>
+  inline constexpr bool _has_sender_traits =
+      !std::is_base_of_v<_no_sender_traits, sender_traits<S>>;
```
lewissbaker, line 150: Do you know why this change was necessary for MSVC?
Maybe add a comment to that effect - at least indicating which version of MSVC was problematic so we know when to revisit.

lewissbaker (approved):

include/unifex/scheduler_concepts.hpp
ericniebler: Yeah, I don't know why. There was nothing wrong with the original code. MSVC doesn't make sense sometimes.

include/unifex/sender_concepts.hpp
ericniebler: I don't know why it was necessary. I'll add a comment.

PR #212, opened 2020-11-28
Add bulk_via

StepanPieshkin:
bulk_via returns a ManySender that produces the results from predecessor on the execution context of specified scheduler, every result scheduled individually.

The example use case I am trying to fulfill is: we are parsing some data, which should be done sequentially, but once we have a parsed an independent chunk of data - we want to send it for processing immediately (processing can be done in parallel):

```c++            
            unifex::bulk_transform(
                unifex::bulk_via(
                    thread_pool_scheduler,
                    unifex::bulk_transform(
                        unifex::bulk_schedule(thread_pool_scheduler, count),
                        [](std::size_t index) noexcept { /* Parse data sequentially*/ return handleToParsedObject; },
                        unifex::seq
                    )
                ),
                [&](std::size_t handleToParsedObject) noexcept { /* Do post-processing in parallel*/ },
                unifex::par_unseq
            )
```

### DIFF
--- a/include/unifex/bulk_via.hpp
+++ b/include/unifex/bulk_via.hpp
@@ -335,7 +335,7 @@ public:
   }
 
   template(typename... Values)(
-      requires !std::is_void_v<Values...>) void set_next(Values &&... values) &
+      requires (!std::is_void_v<Values...>)) void set_next(Values &&... values) &
       noexcept(is_nothrow_next_receiver_v<Receiver, Values...>) {
     op_->element_start();
     UNIFEX_TRY {
--- a/include/unifex/bulk_via.hpp
+++ b/include/unifex/bulk_via.hpp
@@ -1,0 +1,601 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


LeeHowes:
Your intent here is that when set_next is called rather than propagating that inline, it schedules a task onto the underlying executor to run the downstream set_next calls. Is that right?

I can see this working, i'm not sure it's the right way to structure it, though. We could have a general repost algorithm that enqueues to the scheduler obtained by get_scheduler on the receiver. The entire bulk set of operations here is incomplete. I'm a little concerned about long-term maintenance of the whole lot of it as it appears to not be the direction we're going for bulk in C++23.

StepanPieshkin:
> Your intent here is that when set_next is called rather than propagating that inline, it schedules a task onto the underlying executor to run the downstream set_next calls. Is that right?

Correct, same for set_error and set_done - it re-schedule's every call on specified context.

> We could have a general repost algorithm that enqueues to the scheduler obtained by get_scheduler on the receiver.

Doesn't it need a synchronization anyway to ensure that set_value (error/done) called last?

> I'm a little concerned about long-term maintenance of the whole lot of it as it appears to not be the direction we're going for bulk in C++23.

Does it mean that ManySender/ManyReceiver concepts going to be abandoned/redesigned?

Anyway, if you think that this algorithm may be usable, but needs some polishing - let me know =).
If not, I would appreciate if you describe a correct way to express the use case from description.

PR #215, opened 2020-12-09
Minor fixes for MSVC

lewissbaker:
- Remove lvalue-qualification of connect on some schedule senders.
- Simplify CPO forwarding in with_query_value() using a single tag_invoke() overload instead
  of separate overloads for lvalue/const-lvalue/rvalue receiver.

Both of these had been causing problems in recent MSVC builds which seems to
do overload resolution differently from GCC/Clang.


kirkshoop (approved):
Looks good to me

ericniebler (approved):

PR #217, opened 2020-12-14
Add published version of P2175R0 - the cancellation paper

lewissbaker:
This paper has now been published as P2175R0.
https://isocpp.org/files/papers/P2175R0.html

### DIFF
--- a/doc/std/D2175.md
+++ b/doc/std/D2175.md
@@ -3,15 +3,15 @@
 
 | Author:   | Lewis Baker <lbaker@fb.com> |
 | Document: | D2175R0                     |
-| Date:     | 2020-12-14                  |
+| Date:     | 2020-12-15                  |
 | Audience: | SG1                         |
 
 # Abstract
 
 This paper proposes a general-purpose, composable mechanism for requesting cancellation of
-in-flight sender-based async operations as an extension to P0443R13.
+in-flight sender-based async operations as an extension to P0443R14.
 
-The paper [P0443R13](https://wg21.link/P0443R13) proposes adding the sender/receiver concepts
+The paper [P0443R14](https://wg21.link/P0443R14) proposes adding the sender/receiver concepts
 which provide a generic interface for asynchronous operations. As part of this interface,
 when an async operation completes it completes with one of three possible kinds of signals,
 delivered by invoking one of the `set_value`, `set_error` or `set_done` customisation-points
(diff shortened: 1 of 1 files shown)


doc/std/D2175.md
```diff
@@ -0,0 +1,3182 @@
...
+goes out of scope then the program has undefined behaviour. Further, if we create
+a detached operation (work that can _never_ be joined) then this makes it very
+difficult, and in some cases impossible, to know when it will be safe to destroy
+resources passed by-reference to that operation.
+
+Fire and forget interfaces, such as the proposed `std::execution::execute()`
+from P0443R14, often require the programmer to manually ensure that they do
+some extra work at the end of the operation to signal completion so that the
+work can be joined.
+
```
kirkshoop, line 472: I was looking for something stronger to draw attention to the ad-hoc and non-uniform mechanisms that are built to signal completion of work when using fire-and-forget interfaces like execute

doc/std/D2175.md
```diff
@@ -0,0 +1,3182 @@
...
+}
+```
+
+The `stoppable_token` concept checks for the basic interface of a "stop token"
+which is copyable and allows polling to see if stop has been requested and also
+whether a stop request is possible.
+
+It also provides an associated nested template-type-alias, `T::callback_type<CB>`,
+that identifies the stop-callback type to use to register a callback to be executed
+if a stop-request is ever made on a `stoppable_token` of type, `T`.
```
kirkshoop, line 1431: did we consider replacing this nested type alias with an algorithm?

a 'tag_invoke' based CPO named `make_callback()` and a trait to derive the return type would make it possible to support different callback types based on the arguments (like the invocable argument)

This also means that the existing `stop_token` does not need a change to its public interface in order to match the `stoppable_token` concept

doc/std/D2175.md
```diff
@@ -0,0 +1,3182 @@
...
+the `std::stoppable_token` concept that has this more restrictive usage model compatible
+with usage in sender-based algorithms that adhere to the structured concurrency guarantee.
+
+This allows sender algorithms introducing new cancellation scopes to use a stop-token based
+cancellation-model without the need for heap-allocations or atomic reference counting.
+This should allow more efficient implementations compared to what is possible with the
+more general `std::stop_token` interface.
+
+It is worth noting, however, that implementations are still free to use `std::stop_token`
+if desired, as it is also a valid implementation of
```
kirkshoop, line 2632: tagging an incomplete sentence

doc/std/D2175.md
lewissbaker: Yes, I had been considering something like this as well.
I'm not entirely sure why I ended up going with the type-alias, but it was probably because that's what we've implemented in libunifex.

I had also been wondering how the factory-function would accommodate being able to specify the callback type to store separate from the argument used to initialise it.
e.g. I can do: `ST::callback_type<C>{st, X{}}` and this will initialise C with the rvalue-reference to X.

I guess it could be `make_callback<C>(st, X{})` but you'd also want to be able to do something like `make_callback(st, C{})`.

It's probably worth exploring further.

PR #225, opened 2021-02-08
Introduce async_manual_reset_event

ispeters:
This borrows pretty heavily from `folly::coro::Baton`.

`unifex::async_manual_reset_event` is not itself a _Sender_ because it isn't
movable, but it returns an unstoppable sender from `.async_wait()`.

### DIFF
--- a/include/unifex/unstoppable_baton.hpp
+++ b/include/unifex/unstoppable_baton.hpp
@@ -17,6 +17,8 @@
 
 #include <unifex/config.hpp>
 #include <unifex/receiver_concepts.hpp>
+#include <unifex/scheduler_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
 
 #include <atomic>
 #include <type_traits>
@@ -55,7 +57,7 @@ struct _sender {
   template (typename Receiver)
     (requires receiver_of<Receiver>)
   operation<std::decay_t<Receiver>> connect(Receiver&& r) const noexcept(
-      std::is_nothrow_constructible_v<std::decay_t<Receiver>, Receiver&&>) {
+      std::is_nothrow_constructible_v<std::decay_t<Receiver>, Receiver>) {
     return operation<std::decay_t<Receiver>>{*baton_, (Receiver&&)r};
   }
 
@@ -148,12 +150,9 @@ struct _operation<Receiver>::type : private _op_base {
   static void set_value_impl(_op_base* base) noexcept {
     auto self = static_cast<type*>(base);
 
-    UNIFEX_TRY {
-      unifex::set_value(std::move(self->receiver_));
-    }
-    UNIFEX_CATCH(...) {
-      unifex::set_error(std::move(self->receiver_), std::current_exception());
-    }
+    auto op = connect(schedule(), std::move(self->receiver_));
+
+    unifex::start(op);
   }
 };
 
--- a/include/unifex/unstoppable_baton.hpp
+++ b/include/unifex/unstoppable_baton.hpp
@@ -1,0 +1,166 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <unifex/config.hpp>
+#include <unifex/receiver_concepts.hpp>
+
+#include <atomic>
+#include <type_traits>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+
+namespace _ub {
+
+struct _op_base;
+
+template <typename Receiver>
+struct _operation {
+  struct type;
+};
+
+template <typename Receiver>
+using operation = typename _operation<Receiver>::type;
+
+struct unstoppable_baton;
+
+struct _sender {
+  template <template <class...> class Variant,
+            template <class...> class Tuple>
+  using value_types = Variant<Tuple<>>;
+
+  template <template <class...> class Variant>
+  using error_types = Variant<std::exception_ptr>;
+
+  static constexpr bool sends_done = false;
+
+  explicit _sender(unstoppable_baton& baton) noexcept
+    : baton_(&baton) {}
+
+  template (typename Receiver)
+    (requires receiver_of<Receiver>)
+  operation<std::decay_t<Receiver>> connect(Receiver&& r) const noexcept(
+      std::is_nothrow_constructible_v<std::decay_t<Receiver>, Receiver&&>) {
+    return operation<std::decay_t<Receiver>>{*baton_, (Receiver&&)r};
+  }
+
+ private:
+  unstoppable_baton* baton_;
+};
+
+struct unstoppable_baton {
+  unstoppable_baton() noexcept
+    : unstoppable_baton(false) {}
+
+  explicit unstoppable_baton(bool startSignalled) noexcept
+    : state_(startSignalled ? this : nullptr) {}
+
+  void post() noexcept;
+
+  bool ready() const noexcept {
+    return state_.load(std::memory_order_acquire) ==
+        static_cast<const void*>(this);
+  }
+
+  void reset() noexcept {
+    // transition from signalled (i.e. state_ == this) to not-signalled
+    // (i.e. state_ == nullptr).
+    void* oldState = this;
+
+    // We can ignore the the result.  We're using _strong so it won't fail
+    // spuriously; if it fails, it means it wasn't previously in the signalled
+    // state so resetting is a no-op.
+    (void)state_.compare_exchange_strong(
+        oldState, nullptr, std::memory_order_acq_rel, std::memory_order_relaxed);
+  }
+
+  [[nodiscard]] _sender wait() noexcept {
+    return _sender{*this};
+  }
+
+ private:
+  std::atomic<void*> state_{};
+
+  friend struct _op_base;
+
+  static void start_or_wait(_op_base& op, unstoppable_baton& baton) noexcept;
+};
+
+struct _op_base {
+  // note: next_ is intentionally left indeterminate until the operation is
+  //       pushed on the baton's stack of waiting operations
+  //
+  // note: next_ is the first member so that list operations don't have to
+  //       offset into this struct; hopefully that leads to smaller code
+  _op_base* next_;
+  unstoppable_baton* baton_;
+  void (*setValue_)(_op_base*) noexcept;
+
+  explicit _op_base(unstoppable_baton& baton, void (*setValue)(_op_base*) noexcept)
+    : baton_(&baton), setValue_(setValue) {}
+
+  ~_op_base() = default;
+
+  _op_base(_op_base&&) = delete;
+  _op_base& operator=(_op_base&&) = delete;
+
+  void set_value() noexcept {
+    setValue_(this);
+  }
+
+  void start() noexcept {
+    unstoppable_baton::start_or_wait(*this, *baton_);
+  }
+};
+
+template <typename Receiver>
+struct _operation<Receiver>::type : private _op_base {
+  explicit type(unstoppable_baton& baton, Receiver r)
+      noexcept(std::is_nothrow_move_constructible_v<Receiver>)
+    : _op_base(baton, &set_value_impl),
+      receiver_(std::move(r)) {}
+
+  ~type() = default;
+
+  type(type&&) = delete;
+  type& operator=(type&&) = delete;
+
+  using _op_base::start;
+
+ private:
+  UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+
+  static void set_value_impl(_op_base* base) noexcept {
+    auto self = static_cast<type*>(base);
+
+    UNIFEX_TRY {
+      unifex::set_value(std::move(self->receiver_));
+    }
+    UNIFEX_CATCH(...) {
+      unifex::set_error(std::move(self->receiver_), std::current_exception());
+    }
+  }
+};
+
+} // namespace _ub
+
+using _ub::unstoppable_baton;
+
+} // namespace unifex
+
+#include <unifex/detail/epilogue.hpp>
--- a/source/unstoppable_baton.cpp
+++ b/source/unstoppable_baton.cpp
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <unifex/unstoppable_baton.hpp>
+
+namespace unifex::_ub {
+
+void unstoppable_baton::post() noexcept {
+  void* const signalledState = this;
+
+  // replace the stack of waiting operations with a sentinel indicating we've
+  // been signalled
+  void* top = state_.exchange(signalledState, std::memory_order_acq_rel);
+
+  if (top == signalledState) {
+    // we were already signalled so there are no waiting operations
+    return;
+  }
+
+  // We are the first thread to set the state to signalled; iteratively pop
+  // the stack and complete each operation.
+  auto op = static_cast<_op_base*>(top);
+  while (op != nullptr) {
+    std::exchange(op, op->next_)->set_value();
+  }
+}
+
+void unstoppable_baton::start_or_wait(_op_base& op, unstoppable_baton& baton) noexcept {
+  // Try to push op onto the stack of waiting ops.
+  void* const signalledState = &baton;
+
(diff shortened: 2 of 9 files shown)


include/unifex/unstoppable_baton.hpp
```diff
@@ -0,0 +1,166 @@
...
+    void* oldState = this;
+
+    // We can ignore the the result.  We're using _strong so it won't fail
+    // spuriously; if it fails, it means it wasn't previously in the signalled
+    // state so resetting is a no-op.
+    (void)state_.compare_exchange_strong(
+        oldState, nullptr, std::memory_order_acq_rel, std::memory_order_relaxed);
+  }
+
+  [[nodiscard]] _sender wait() noexcept {
```
kirkshoop, line 92: It would be nice to not use the word wait in non-blocking code. I would like 'wait' to become a code-smell. `sync_wait(b.wait())` is also awkward.

options:

```
sender on_ready()
sync_wait(b.on_ready())
```
[on_ is not my favorite prefix.. w.on_mouse_moves() or w.mouse_moves()?]

```
bool is_ready()
sender ready()
sync_wait(b.ready())
```

ispeters: I'm happy to change; your ideal of making `wait` a code smell sounds nice.  I'm uncertain about your suggested names, though.  Predicates in the STL seem not to start with `is_` (e.g. `vector::empty()`), so `bool is_ready()` and `sender ready()` feels suboptimal.

include/unifex/unstoppable_baton.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+struct _op_base;
+
+template <typename Receiver>
+struct _operation {
+  struct type;
+};
+
+template <typename Receiver>
+using operation = typename _operation<Receiver>::type;
```
ispeters, line 38: Some of the code below would be simpler if this template alias was defined as
```
template <typename Receiver>
using operation = typename _operation<std::decay_t<Receiver>>::type;
```

Is there a preference one way or the other?

include/unifex/unstoppable_baton.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+  static constexpr bool sends_done = false;
+
+  explicit _sender(unstoppable_baton& baton) noexcept
+    : baton_(&baton) {}
+
+  template (typename Receiver)
+    (requires receiver_of<Receiver>)
+  operation<std::decay_t<Receiver>> connect(Receiver&& r) const noexcept(
+      std::is_nothrow_constructible_v<std::decay_t<Receiver>, Receiver&&>) {
```
ispeters, line 58: I'm skeptical I've done this right because forwarding references confuse me.

Also, now that I think about it, I also need to add to this condition that the `operation` constructor is no-throw, because it's not guaranteed that one follows from the other.

source/unstoppable_baton.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    // note: on the first iteration, this line transitions op.next_ from
+    //       indeterminate to a well-defined value
+    op.next_ = static_cast<_op_base*>(top);
+  } while (!baton.state_.compare_exchange_weak(
+      top,
+      static_cast<void*>(&op),
+      std::memory_order_release,
+      std::memory_order_acquire));
+}
```
ispeters, line 62: This is a static function that takes the baton *second* because the caller will have `op` in the first argument position before calling.  If this were a member function, I'd expect the compiler to have to emit instructions to juggle registers before calling to match the calling convention.  I would assume that register renaming would make the juggling zero runtime cost, but, assuming the instructions are emitted, it's not zero binary size cost.

include/unifex/unstoppable_baton.hpp
```diff
@@ -0,0 +1,166 @@
...
+  void reset() noexcept {
+    // transition from signalled (i.e. state_ == this) to not-signalled
+    // (i.e. state_ == nullptr).
+    void* oldState = this;
+
+    // We can ignore the the result.  We're using _strong so it won't fail
+    // spuriously; if it fails, it means it wasn't previously in the signalled
+    // state so resetting is a no-op.
+    (void)state_.compare_exchange_strong(
+        oldState, nullptr, std::memory_order_acq_rel, std::memory_order_relaxed);
```
MarkCMann, line 89: From the perspective of the caller, there's no indication that the operation either succeeded or failed so the caller can't make any assumptions about whether this method synchronizes or not -- thus making the asymmetry here confusing. Also, the release and acquire semantics don't seem to be needed because the new value (`nullptr`) doesn't have any dependent data to synchronize.

lewissbaker: The use case this was trying to satisfy was implementing a dirty-flag where one or more producers might add work to be done and then call post() on the baton to wake up the consumer, then the consumer would wake-up and immediately call reset() to clear the dirty-flag before processing changes so that it can be sure that it won't miss any subsequent wake-ups that might come up while processing the previous changes.

```
unstoppable_baton baton;
T state; // assume internally synchronised

void produce() {
  state.enqueue(item);  // 'release' semantics - synchronises with 'dequeue_all()'
  baton.post();
}

task consume() {
  while (true) {
    co_await baton.wait();
    baton.reset();
    auto items = state.dequeue_all(); // 'acquire' semantics here.
    process(items);
  }
}
```

Now let's say we have T1, T2 both calling produce() and T3 calling consume().
We want to make sure that we don't have any missed wake-ups of consume().

Consider what could happen if reset() has relaxed memory order.
I was worried that the write inside reset() might be reordered to occur after the state.dequeue_all() call (since they are occurring on unrelated objects), leaving the following program order.
```
T1: state.enqueue(item)
T1: baton.post()
T3: baton.wait() - sees already signalled
T3: state.dequeue_all() - sees item from T1
T2: state.enqueue(item)
T2: baton.post() - sees already signalled, does nothing
T3: baton.reset() - clears flag set by T2 from T
T3: ... processes item from T1 and continues around loop
T3: baton.wait() - blocks here even though there are unprocessed items in the queue
```

The intention behind making the reset() operation 'acquire' was to ensure that the subequent dequeue_all() operation is not reordered before the reset() from the perspective of T2. But it's possible this logic is flawed.

We need to ensure that either:
- T2's enqueue() appears in TMO of state before T3's dequeue() - i.e. that T3 will process T2's item
- T3's reset() appears in TMO of baton before T3's post() - i.e. that T2's subsequent post() will signal the baton so that T3's subsequent wait() will see it

It's possible that we may actually need to strengthen the memory order here to seq_cst to satisfy this, although I'm not entirely sure...

Alternatively, this use-case might be better served by a different synchronisation primitive like an `auto_reset_event`.

include/unifex/unstoppable_baton.hpp
ericniebler: I like it one way. Lewis likes it the other. :-P

include/unifex/unstoppable_baton.hpp
ericniebler, line 58: The `&&` here is superfluous. Either `Receiver` is an lvalue reference type, in which case the `&&` collapses away, or else it isn't reference qualified, in which case `Receiver` and `Receiver&&` are semantically indistinguishable.

```suggestion
      std::is_nothrow_constructible_v<std::decay_t<Receiver>, Receiver>) {
```

include/unifex/unstoppable_baton.hpp
```diff
@@ -0,0 +1,166 @@
...
+  operation<std::decay_t<Receiver>> connect(Receiver&& r) const noexcept(
+      std::is_nothrow_constructible_v<std::decay_t<Receiver>, Receiver&&>) {
+    return operation<std::decay_t<Receiver>>{*baton_, (Receiver&&)r};
+  }
+
+ private:
+  unstoppable_baton* baton_;
+};
+
+struct unstoppable_baton {
```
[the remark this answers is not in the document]
ericniebler: Seconded. I never much cared for the name "baton".

source/unstoppable_baton.cpp
```diff
@@ -0,0 +1,64 @@
...
+  if (top == signalledState) {
+    // we were already signalled so there are no waiting operations
+    return;
+  }
+
+  // We are the first thread to set the state to signalled; iteratively pop
+  // the stack and complete each operation.
+  auto op = static_cast<_op_base*>(top);
+  while (op != nullptr) {
+    std::exchange(op, op->next_)->set_value();
```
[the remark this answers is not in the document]
ericniebler: That seems like a good idea. In the case when you actually want to complete inline, you could just specify the `inline_scheduler`.

include/unifex/unstoppable_baton.hpp
ispeters: Sounds like I should follow the established convention, then.  I think the convention is not to embed `decay_t` in this alias, right?

include/unifex/unstoppable_baton.hpp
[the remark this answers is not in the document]
ispeters: It seems key to me that this type's sender doesn't send done.  Do we want `unstoppable_async_manual_reset_event`?  Or do we think there will never be an `async_manual_reset_event` that _does_ send done?

include/unifex/unstoppable_baton.hpp
```diff
@@ -0,0 +1,166 @@
+/*
+ * Copyright 2019-present Facebook, Inc.
```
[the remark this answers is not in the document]
ispeters: Is this a question for Facebook's lawyers?

source/unstoppable_baton.cpp
```diff
@@ -0,0 +1,64 @@
...
+void unstoppable_baton::start_or_wait(_op_base& op, unstoppable_baton& baton) noexcept {
+  // Try to push op onto the stack of waiting ops.
+  void* const signalledState = &baton;
+
+  void* top = baton.state_.load(std::memory_order_acquire);
+
+  do {
+    if (top == signalledState) {
+      // Already in the signalled state; don't push it.
+      op.set_value();
```
ispeters, line 50: @lewissbaker and @ericniebler, should this call to `set_value()` be scheduled on the receiver's scheduler, too, to match the behaviour you propose in `post()`?

include/unifex/unstoppable_baton.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+  using _op_base::start;
+
+ private:
+  UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+
+  static void set_value_impl(_op_base* base) noexcept {
+    auto self = static_cast<type*>(base);
+
+    auto op = connect(schedule(), std::move(self->receiver_));
```
ispeters, line 153: Composing `schedule()` in like this means that, if `receiver_` has a stoppable stop token, then the scheduler may respond to it and `set_done`.  Given the baton is unstoppable, this feels awkwardly inconsistent.  Should I also compose in a `with_query_value` that ensures the scheduler only receives an unstoppable token?

include/unifex/unstoppable_baton.hpp
ericniebler: I think mostly the aliases do the de-qualification, since I did most of the ADL isolation work.

include/unifex/unstoppable_baton.hpp
MarkCMann: I think that the use case that you provided makes sense for the acquire semantics here. From my perspective this is a good argument to strengthen the memory ordering for `reset`. 

To expand on your example, if there are two consumers each resetting the baton, then only one of them will have acquire semantics -- any other consumer would have relaxed semantics which means they wouldn't synchronize and could receive stale results.

include/unifex/unstoppable_baton.hpp
ericniebler: That would be my preference. @lewissbaker ?

source/unstoppable_baton.cpp
ericniebler: IMO yes.

include/unifex/unstoppable_baton.hpp
ispeters: I settled on `async_manual_reset_event`, `set`, `reset`, and `async_wait`.  I'm a little uncomfortable losing the "unstoppable" nature of the type since you _could_ write a stoppable version at some cost increment, and I'm a little concerned @kirkshoop won't be happy with `async_wait`, so feel free to suggest new names.

include/unifex/unstoppable_baton.hpp
ispeters: I was trying to find a pattern to follow and ran into the construction of the operation state in [`let_with`](https://github.com/facebookexperimental/libunifex/blob/master/include/unifex/let_with.hpp#L82).  Because that operation state can be instantiated with some `Receiver&` type, [this constructor](https://github.com/facebookexperimental/libunifex/blob/master/include/unifex/let_with.hpp#L96) takes either an lvalue or rvalue reference.  In contrast, my operation state takes its receiver by value and `move`s it into place.

It feels like there should be a policy I follow here and it's not obvious to me what the policy should be.  Is there already a consensus?  The approach in `let_with` potentially leads to more distinct template instantiations because it could be instantiated with both `Receiver` and `Receiver&`, but it probably results in a more efficient constructor.

ispeters:
I've taken this as far as I can without further commentary from the reviewers.

The things I still need are:
 * an ack on the naming changes;
 * a resolution to the synchronization debate between @MarkCMann and @lewissbaker; and
 * an answer from @ericniebler to [my question about references](https://github.com/facebookexperimental/libunifex/pull/225#discussion_r575513241).

include/unifex/unstoppable_baton.hpp
ericniebler: Personally, I think taking by value and moving it into `connect` is perfectly fine.

lewissbaker: > I think mostly the aliases do the de-qualification, since I did most of the ADL isolation work.

I tended to prefer to leave the decay out of the alias so that I can tell its doing the right thing at point of use.
If I the decay is in the alias then I always had to scroll up to see the definition of the alias to check which template args implicitly had what transformation performed to them.

Having the aliases be straight aliases I feel separates the concerns better. The aliases are there to simplify use of the ADL-isolation wrappers and that only.

lewissbaker: > Personally, I think taking by value and moving it into connect is perfectly fine.

Yes, I think that's probably fine. The vast majority of receivers end up just being a pointer or two anyway which should be pretty cheap to copy and/or pretty inlinable.

include/unifex/unstoppable_baton.hpp
lewissbaker: You don't need to write `Receiver&&` there, just having `is_nothrow_constructible_v<std::decay_t<Receiver>, Receiver>` should give the same result.

source/unstoppable_baton.cpp
lewissbaker: Have you veriifed that the compiler does not emit the register juggling instructions when calling this?
If so, maybe record the intent/rationale in a comment (either here or at the declaration)?

include/unifex/unstoppable_baton.hpp
lewissbaker: > From my perspective this is a good argument to strengthen the memory ordering for `reset`

Do you mean strengthening to seq_cst? Or do you think acq_rel is strong enough?

> if there are two consumers each resetting the baton

This pattern really only works correctly for a single consumer processing the dirty flag/event and resetting it.

For multiple consumers an `async_auto_reset_event` would be more appropriate.

include/unifex/unstoppable_baton.hpp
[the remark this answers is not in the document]
lewissbaker: I think folly ended up removing all of the copyright years.
https://github.com/facebook/folly/blob/master/folly/experimental/coro/Task.h#L2

include/unifex/unstoppable_baton.hpp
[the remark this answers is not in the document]
lewissbaker: I think it's probably better to leave out the unstoppable prefix for now.
If we do end up implementing a stoppable version then we can rework the names then...

test/unstoppable_baton_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+      .WillOnce(Invoke([](std::exception_ptr eptr) noexcept {
+        try {
+          std::rethrow_exception(eptr);
+        } catch (const std::runtime_error& e) {
+          EXPECT_STREQ(e.what(), "from set_value()");
+        }
+      }));
+
+  start(op);
+}
```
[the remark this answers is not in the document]
lewissbaker: I wonder if the default scheduler should be a `trampoline_scheduler` rather than an `inline_scheduler`.
I'm sure @kirkshoop has some thoughts on this ;)

include/unifex/unstoppable_baton.hpp
lewissbaker: I think that semantic seems fine for now.

It's probably worth encapsulating the logic for an adapter that prevents stop-requests from propagating to a child sender into an `unstoppable(s)` algorithm. It would just wrap the parent receiver in something that hooks `get_stop_token()` to return an `unstoppable_token`.

source/unstoppable_baton.cpp
ispeters: I hadn't done it, but now I have: https://godbolt.org/z/nej31v.

At `-O3`, the differences seem negligible because of the effects of inlining, but at `-Oz`, `unifex::_amre::_op_base::start()` is actually emitted, and it's one instruction shorter in the case of the static method.

Interestingly, the implementation of `start_or_wait` is also slightly shorter at `-Oz` when it's a static function (change the `#if 0` to `#if 1` at the Godbolt to see the differences).  Fooling around a little more, it seems I should put `next_` and `setValue_` first and second (in either order) to shrink `start_or_wait` because ARM can `ldp` the first two fields into a pair of registers.

ericniebler (approved):

include/unifex/unstoppable_baton.hpp
MarkCMann: > Do you mean strengthening to seq_cst? Or do you think acq_rel is strong enough?

`acq_rel` should be strong enough here

> This pattern really only works correctly for a single consumer processing the dirty flag/event and resetting it.
> For multiple consumers an async_auto_reset_event would be more appropriate.

I might be overthinking this then -- I'm not sure of the intended use case of this class so if the single consumer case makes sense to optimize for then let's go with `acq_rel` semantics.

ispeters: Addressed in 6b39655.

PR #226, opened 2021-02-09
Support building with RTTI and exceptions disabled

ericniebler:
I have tested this locally by building and running all of the examples and tests with `-fno-rtti -fno-exceptions`. This should absolutely have test coverage in CI, but I'm deferring that for later.

### DIFF
--- a/examples/any_unique.cpp
+++ b/examples/any_unique.cpp
@@ -25,6 +25,7 @@
 template <typename T>
 using is_type_index = std::is_same<std::type_index, T>;
 
+#if __cpp_rtti
 inline constexpr struct get_typeid_cpo {
   using type_erased_signature_t =
       std::type_index(const unifex::this_&) noexcept;
@@ -44,6 +45,7 @@ inline constexpr struct get_typeid_cpo {
     return tag_invoke(get_typeid_cpo{}, x);
   }
 } get_typeid;
+#endif
 
 struct destructor {
   explicit destructor(bool& x) : ref_(x) {}
@@ -87,6 +89,7 @@ class counting_memory_resource : public memory_resource {
 #endif
 
 int main() {
+#if __cpp_rtti
   using A = unifex::any_unique_t<get_typeid>;
   using B = unifex::any_unique_t<>;
   {
@@ -108,6 +111,7 @@ int main() {
     }
     assert(hasDestructorRun);
   }
+#endif
 #if !UNIFEX_NO_MEMORY_RESOURCE
   {
     counting_memory_resource res{new_delete_resource()};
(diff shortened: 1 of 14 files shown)


test/any_unique_test.cpp
```diff
@@ -26,6 +26,7 @@
 #include <gtest/gtest.h>
 
 namespace {
+#if __cpp_rtti
 inline constexpr struct get_typeid_cpo {
   using type_erased_signature_t =
       unifex::type_index(const unifex::this_&) noexcept;
```
ispeters, line 32: I introduced this type in PR #154; its definition is _supposed_ to depend on whether `__cpp_rtti` is defined or not.  I wouldn't expect you'd need to compile out this CPO.  If you do, maybe there's a deeper change that's needed?

ericniebler: Ah, you're right. I got a little overenthusiastic. Will remove.

examples/any_unique.cpp
```diff
@@ -25,6 +25,7 @@
 template <typename T>
 using is_type_index = std::is_same<std::type_index, T>;
 
+#if __cpp_rtti
```
ispeters, line 28: Two thoughts:
 1. I think it'd be good to either use `__cpp_rtti` everywhere, or use `UNIFEX_NO_RTTI` everywhere; and
 2. I wonder whether it'd be worth switching from `std::type_index` to `unifex::type_index` so you can leave this code in the build.

source/inplace_stop_token.cpp
```diff
@@ -27,7 +27,11 @@ inplace_stop_source::~inplace_stop_source() {
   assert((state_.load(std::memory_order_relaxed) & locked_flag) == 0);
 #ifndef NDEBUG
   for (auto* cb = callbacks_; cb != nullptr; cb = cb->next_) {
+#if __cpp_rtti
```
ispeters, line 30: Same comment as above re: consistent use of `UNIFEX_NO_RTTI`.

source/inplace_stop_token.cpp
```diff
@@ -27,7 +27,11 @@ inplace_stop_source::~inplace_stop_source() {
...
     printf("dangling inplace_stop_callback: %s\n", typeid(*cb).name());
+#else
+    printf("dangling inplace_stop_callback\n");
```
ispeters, line 33: This is a significantly less-useful error message than what you'd get with RTTI.  What do you think of having a debug-only member on the callback that uses `unifex::type_id` to ensure there's always a string available?

ispeters (approved):
LGTM

ispeters:
Looks good!

PR #228, opened 2021-02-11
change detection of coroutines from config time to compile time

ericniebler:


include/unifex/connect_awaitable.hpp
```diff
@@ -92,12 +92,13 @@ class _sender_task<Receiver>::type {
...
-    auto await_transform(Value&& value)
-        noexcept(is_nothrow_callable_v<decltype(unifex::await_transform), promise_type&, Value>)
-        -> callable_result_t<decltype(unifex::await_transform), promise_type&, Value> {
-      return unifex::await_transform(*this, (Value&&)value);
+    template <typename Value>
+    auto await_transform(Value&& value) -> decltype(auto) {
+      if constexpr (callable<decltype(unifex::await_transform), promise_type&, Value>) {
+        return unifex::await_transform(*this, (Value&&)value);
+      } else {
+        return Value((Value &&) value);
```
lewissbaker, line 100: Why are you forcing a copy of the argument here?
You should be able to return a `Value&&` safely.

include/unifex/connect_awaitable.hpp
```diff
@@ -192,10 +193,7 @@ namespace _await_cpo {
                   if constexpr (valueCount == 0) {
                     unifex::set_value(std::move(receiver));
                   } else if constexpr (valueCount == 1) {
-                    using value_type =
-                        typename sender_value_types_t<Awaitable, single_type, single_type>
-                            ::type::type;
-                    unifex::set_value(std::move(receiver), static_cast<value_type&&>(result));
+                    unifex::set_value(std::move(receiver), static_cast<result_type&&>(result));
```
lewissbaker, line 196: Was this causing a problem with msvc or are you just simplifying?

test/awaitable_senders_test.cpp
```diff
@@ -66,7 +66,7 @@ TEST(awaitable_senders, task_cancellation) {
       [&]() -> task<int> {
         co_await schedule_after(sched, 500ms);
         ADD_FAILURE();
-        std::terminate();
+        co_return -1;
```
lewissbaker, line 69: MSVC was complaining about executing running off the end?

lewissbaker:
I think with these changes that you can now also remove the logic in `unifex_flags.cmake` that sets the UNIFEX_NO_COROUTINES and associated vars.

PR #230, opened 2021-02-17
Update copyright headers

ispeters:
It seems we've got new guidance for what a copyright header should look
like.  To improve the odds that copy-paste programming gets it right,
I've updated all the copyright notices to use the new style.


ericniebler:
Thanks.

PR #236, opened 2021-03-02
Introduce single_async_manual_reset_event

ispeters:
This diff introduces `single_async_manual_reset_event`.

The semantics are very similar to `async_manual_reset_event` but:
 - there may only be one waiter at a time; trying to add a second will
   terminate the process
 - waiters may cancel their wait

The method names are the same as `async_manual_reset_event`:
 - async_wait() to get a _Sender_ upon which to wait
 - set() to signal waiter to wake up
 - reset() to return to the unsignalled state
 - ready() to query whether the object is currently signalled

I've chosen to `std::terminate()` when I detect an incorrect state; I
don't know whether this is acceptable.  We could throw an exception,
instead, I suppose, but the sorts of errors I'm detecting are
unrecoverable programmer errors so I've erred on the side of failing
fast.

### DIFF
--- a/include/unifex/single_async_manual_reset_event.hpp
+++ b/include/unifex/single_async_manual_reset_event.hpp
@@ -32,7 +32,7 @@
 
 namespace unifex {
 
-namespace _unnamed {
+namespace _samre {
 
 struct _op_base;
 
@@ -44,7 +44,7 @@ struct _operation {
 template <typename Receiver>
 using operation = typename _operation<Receiver>::type;
 
-struct unnamed_primitive;
+struct single_async_manual_reset_event;
 
 struct _sender {
   template <template <class...> class Variant,
@@ -56,29 +56,29 @@ struct _sender {
 
   static constexpr bool sends_done = true;
 
-  explicit _sender(unnamed_primitive& evt) noexcept
+  explicit _sender(single_async_manual_reset_event& evt) noexcept
     : evt_(&evt) {}
 
   template (typename Receiver)
     (requires receiver_of<Receiver>)
   operation<remove_cvref_t<Receiver>> connect(Receiver&& r) const noexcept(
-      noexcept(operation<remove_cvref_t<Receiver>>{std::declval<unnamed_primitive&>(), (Receiver&&)r})) {
+      noexcept(operation<remove_cvref_t<Receiver>>{std::declval<single_async_manual_reset_event&>(), (Receiver&&)r})) {
     return operation<remove_cvref_t<Receiver>>{*evt_, (Receiver&&)r};
   }
 
  private:
-  unnamed_primitive* evt_;
+  single_async_manual_reset_event* evt_;
 };
 
 inline std::uintptr_t to_addr(const void* ptr) noexcept {
   return reinterpret_cast<std::uintptr_t>(ptr);
 }
 
-struct unnamed_primitive {
-  unnamed_primitive() noexcept
-    : unnamed_primitive(false) {}
+struct single_async_manual_reset_event {
+  single_async_manual_reset_event() noexcept
+    : single_async_manual_reset_event(false) {}
 
-  explicit unnamed_primitive(bool startSignalled) noexcept
+  explicit single_async_manual_reset_event(bool startSignalled) noexcept
     : state_(to_addr(startSignalled ? this : nullptr)) {}
 
   void set() noexcept;
@@ -123,9 +123,9 @@ struct unnamed_primitive {
 
 struct _op_base {
   void (*completeImpl_)(_op_base*) noexcept;
-  unnamed_primitive* evt_;
+  single_async_manual_reset_event* evt_;
 
-  explicit _op_base(unnamed_primitive& evt, void (*completeImpl)(_op_base*) noexcept) noexcept
+  explicit _op_base(single_async_manual_reset_event& evt, void (*completeImpl)(_op_base*) noexcept) noexcept
     : evt_(&evt), completeImpl_(completeImpl) {}
 
   ~_op_base() = default;
@@ -191,7 +191,7 @@ struct _receiver<Receiver>::type {
 
 template <typename Receiver>
 struct _operation<Receiver>::type : private _op_base {
-  explicit type(unnamed_primitive& evt, Receiver r)
+  explicit type(single_async_manual_reset_event& evt, Receiver r)
       // TODO: noexceptness is incomplete
       noexcept(std::is_nothrow_move_constructible<Receiver>::value)
     : _op_base(evt, &complete_impl),
@@ -254,9 +254,9 @@ struct _operation<Receiver>::type : private _op_base {
   }
 };
 
-} // namespace _unnamed
+} // namespace _samre
 
-using _unnamed::unnamed_primitive;
+using _samre::single_async_manual_reset_event;
 
 } // namespace unifex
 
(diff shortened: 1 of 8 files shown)


include/unifex/unnamed_primitive.hpp
```diff
@@ -0,0 +1,260 @@
...
+      const type& r,
+      Func&& func) {
+    std::invoke(func, r.receiver_);
+  }
+};
+
+template <typename Receiver>
+struct _operation<Receiver>::type : private _op_base {
+  explicit type(unnamed_primitive& evt, Receiver r)
+      // TODO: noexceptness is incomplete
```
ispeters, line 195: Oops.

source/unnamed_primitive.cpp
```diff
@@ -0,0 +1,145 @@
...
+  //  1. it was nullptr; we must be getting cancelled while we're still in
+  //     start() and start_or_wait hasn't updated state_, yet.
+  //
+  //  2. it was signalled; we could be in one of two sub-states:
+  //    a. we're still in start() and state_ was signalled before the call to
+  //       start_or_wait, or
+  //    b. start() completed by registering us to wait and cancellation lost a
+  //       race to signalling.
+  //
+  //  3. it was "this"; we were waiting for a signal but cancellation happened
```
ispeters, line 60:
```suggestion
  //  3. it was opAddr; we were waiting for a signal but cancellation happened
```

ispeters:
@ericniebler suggested I include an example use case for this new thing.  I think it would be a building block for something I've sketched out as `SingleQueue`, but it could do with a better name, and I'm not completely clear on the missing details:
```
template <typename T>
struct SingleQueue {
  SingleQueue() noexcept = default;

  ~SingleQueue() {
    if (canConsume_.ready()) {
      value_.destruct();
    }
  }

  auto set_next(T&& value) {
    return transform(canProduce_.async_wait(), [this, &value] {
      canProduce_.reset();

      value_.construct(std::move(value));

      canConsume_.set();
    });
  }

  auto next() {
    return transform(canConsume_.async_wait(), [this] {
      canConsume_.reset();

      scope_guard guard = [this]() noexcept {
        value_.destruct();
        canProduce_.set();
      };

      return std::move(value_).get();
    });
  }

  auto cleanup() {
    // ???
    return just();
  }

 private:
  single_async_manual_reset_event canProduce{true};
  single_async_manual_reset_event canConsume;
  manual_lifetime<T> value_;
};
```
The intent behind `SingleQueue` is that it's a stream that can bridge from a listener-style API to a _Sender_-based one.

You can't really write the above `SingleQueue` with `async_manual_reset_event` because you need to be able to cancel.

source/single_async_manual_reset_event.cpp
```diff
@@ -0,0 +1,182 @@
...
+    // leave it in the no-waiter state
+    clear_cancelled_state(state_, cancelledState);
+
+    complete_operation(opAddr);
+    return;
+  }
+
+  if (expectedState == to_addr(nullptr)) {
+    // state 1
+    if (state_.compare_exchange_strong(
```
lewissbaker, line 129: I think the order of this compare-exchange and the previous compare-exchange might need to be reversed, or at least you need to account for the possibility that this compare-exchange failed because the state transition from nullptr -> opAddr on another thread in between the previous compare-exchange and this compare-exchange and thus need to retry the opAddr -> not-signalled transition again.

The state is potentially going to transition from: not-signalled -> opAddr -> signalled concurrently with this cancel call and so you probably need to try to transition from not-signalled -> cancelled and if that fails then try opAddr -> not-signalled, and if it reaches 'signalled' then do nothing as the 'set()' call is going to be responsible for calling start().

source/single_async_manual_reset_event.cpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+
+void single_async_manual_reset_event::start_or_wait(_op_base* op) noexcept {
+  const std::uintptr_t opAddr = to_addr(op);
+  const std::uintptr_t signalledState = to_addr(this);
+  const std::uintptr_t cancelledState = signalledState + 1;
+
+  std::uintptr_t expectedState = to_addr(nullptr);
```
lewissbaker, line 153: Consider doing a load(acquire) here and doing a short-circuit that immediately calls complete_operation() if the state is signalled, or that clears the state back to not-signalled if it was cancelled and immediately calls complete_operation(). This way we don't end up with an unnecessary RMW operation in the case that the event is already signalled.

Also, with some tweaks you might also be able to avoid registering the stop-callback in start() if the event is already in the signalled state when you enter start().

include/unifex/single_async_manual_reset_event.hpp
```diff
@@ -0,0 +1,263 @@
...
+struct single_async_manual_reset_event {
+  single_async_manual_reset_event() noexcept
+    : single_async_manual_reset_event(false) {}
+
+  explicit single_async_manual_reset_event(bool startSignalled) noexcept
+    : state_(to_addr(startSignalled ? this : nullptr)) {}
+
+  void set() noexcept;
+
+  bool ready() const noexcept {
```
lewissbaker, line 86: Declare `[[nodiscard]]`?

source/single_async_manual_reset_event.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+    // leave it in the no-waiter state
+    clear_cancelled_state(state_, cancelledState);
+
+    complete_operation(opAddr);
+    return;
+  }
+  else {
+    // someone is already waiting
+    std::terminate();
```
lewissbaker, line 178: Maybe `UNIFEX_ASSUME_UNREACHABLE;` instead?

source/single_async_manual_reset_event.cpp
```diff
@@ -0,0 +1,182 @@
...
+  else if (expectedState == signalledState) {
+    // already signalled; don't wait
+    complete_operation(opAddr);
+    return;
+  }
+  else if (expectedState == cancelledState) {
+    // cancelled before starting; clean up then complete
+
+    // leave it in the no-waiter state
+    clear_cancelled_state(state_, cancelledState);
```
lewissbaker, line 171: This is problematic as it might race with another thread calling `set()` which could cause the "signalled" state of the event to be lost.

test/single_async_manual_reset_event_test.cpp
```diff
@@ -0,0 +1,421 @@
...
+    EXPECT_CALL(receiverImpl, set_value()).Times(0);
+    EXPECT_CALL(receiverImpl, set_error(_)).Times(0);
+    EXPECT_CALL(receiverImpl, set_done()).Times(0);
+
+    start(op);
+  }
+
+  EXPECT_CALL(receiverImpl, set_value()).Times(1);
+  EXPECT_CALL(receiverImpl, set_error(_)).Times(0);
+  EXPECT_CALL(receiverImpl, set_done()).Times(0);
```
lewissbaker, line 151: I forget - is the `EXPECT_CALL(...).Times(1)` checking that "There have so far been 1 call to this method" or "There will eventually be exactly 1 call to this method by the time the mock object destructs"?
The usage seems like the former, but my recollection of mocking frameworks was that many of the methods behave like the latter.

lewissbaker (changes requested):
Thanks for putting up this PR. This will be a useful primitive for applications.

I think some of the state-transition logic around cancellation still needs to be tweaked a bit to eliminate some race conditions, but otherwise the general approach looks good.

In particular, I think it's important to make sure that the event object itself only logically has unsignalled and signalled states and that the cancelled state is only ever transitory while a waiter is attaching itself.

Also, don't forget to add documentation to the `doc/api_reference.md` file for use of this synchronisation primitive.

PR #239, opened 2021-03-12
add stop, stop_if_requested, and co_await get_stop_token to fetch a coro's stop token

ericniebler:


include/unifex/get_stop_token.hpp
```diff
@@ -16,15 +16,41 @@
...
+        return false; // don't suspend
+      }
+      inplace_stop_token await_resume() const noexcept {
+        return stoken_;
+      }
+    };
+    friend awaiter_ operator co_await(_fn) noexcept {
+      return {};
+    }
+#endif
```
lewissbaker: I think in time this should be generalised to allow the `get_stop_token` CPO to be used as a sender that will just return the result of calling `get_stop_token()` on the receiver it is connected to.
This could be made a general pattern which could be encapsulated, either in a base-class or a higher-order algorithm.
e.g. `query(get_stop_token)` could be a syntax for producing such a sender.

However, I think for this to work in coroutines it needs to be a `typed_sender` which currently doesn't let the return-type depend on the receiver type. For that we need to move the type-traits for senders into the operation-state so that will have to wait.

However, this implementation assumes that the only coroutine that might await this will have a stop-token type of `inplace_stop_token` which is not necessarily always going to be the case.

This would be better implemented by customising the `await_transform()` CPO, which provides the promise object/type, and then returning an awaiter whose `await_resume()` method returns the appropriate stop-token type deduced from calling the CPO on the promise.

lewissbaker: Also, I'm not sure about just having the CPO itself be directly awaitable.

What was the thinking between having `co_await get_stop_token;` vs `co_await get_stop_token();`?

include/unifex/task.hpp
```diff
@@ -226,6 +227,21 @@ struct _awaiter {
...
+  coro::coroutine_handle<> await_suspend(coro::coroutine_handle<Promise> coro) const noexcept {
+    if (get_stop_token(coro.promise()).stop_requested()) {
+      return coro.promise().unhandled_done();
+    }
+    return coro; // don't suspend
+  }
+  void await_resume() const noexcept {
+  }
+};
+
```
lewissbaker, line 244: Could this algorithm be implemented as a sender that could then be used within other composed operations rather than just being usable by coroutines?

include/unifex/task.hpp
```diff
@@ -282,6 +298,15 @@ struct _task<T>::type {
 template <typename T>
 using task = typename _task::_task<T>::type;
 
+// Await this to cancel and unwind if stop has been requested:
+inline constexpr _task::_stop_if_requested stop_if_requested {};
+
+// Await this to cancel unconditionally:
+inline constexpr auto stop = just_done();
```
lewissbaker, line 305: This is creating a separate dialect for coroutines.
Why `co_await stop` and not `co_await just_done();`?

lewissbaker (changes requested):

include/unifex/get_stop_token.hpp
```diff
@@ -16,15 +16,44 @@
...
+#if !UNIFEX_NO_COROUTINES
+    template <typename StopToken>
+    struct awaiter_ {
+      StopToken stoken_;
+      bool await_ready() const noexcept {
+        return false;
+      }
+      template <typename Promise>
+      bool await_suspend(coro::coroutine_handle<Promise> coro) noexcept {
+        stoken_ = _fn{}(coro.promise());
```
lewissbaker, line 41: You don't need this line any more. Retrieval of the stop-token is done in `await_transform()` customisation below.

include/unifex/get_stop_token.hpp
```diff
@@ -16,15 +16,44 @@
...
+        return false; // don't suspend
+      }
+      StopToken await_resume() const noexcept {
+        return stoken_;
```
lewissbaker, line 45: This could use `std::move(stoken_)`.

Alternatively, you could have the `awaiter` hold a reference to the promise and then just `return get_stop_token(promise_);` here instead.

include/unifex/task.hpp
ericniebler: Possibly! I'll play with it.

include/unifex/task.hpp
ericniebler: I'm ambivalent. On the one hand, I'm in favor of keeping the library small, and not creating dialects. On the other hand, with "done" and "stop" we already have two ways of saying the same thing. We have `set_done`, `just_done`, and `transform_done`, but we also have `inplace_stop_[token|source]`, `request_stop`, and `stop_requested`. I expect the "stop"-flavored names to be used extensively in coroutines (though some in sender/receiver), and the "done"-flavored names to be used primarily in sender/receiver.

When you have a codebase filled with lots of coroutines and little to no sender/receiver, `co_await just_done()` doesn't roll of the tongue.

All that said, I'm happy to leave it out and define it outside libunifex for the convenience of the clients of `unifex::task<>`.

include/unifex/get_stop_token.hpp
ericniebler: I think parameterizing on the stop token will lead to fewer template instantiations.

include/unifex/get_stop_token.hpp
ericniebler: > What was the thinking between having `co_await get_stop_token;` vs `co_await get_stop_token();`?

I don't care which. Seems like you would prefer `co_await get_stop_token()`, is that right?

include/unifex/get_stop_token.hpp
```diff
@@ -16,15 +16,44 @@
...
+        return (StopToken&&) stoken_;
+      }
+    };
+    template <typename StopToken>
+    awaiter_(StopToken) -> awaiter_<StopToken>;
+
+    template <typename Promise>
+    friend auto tag_invoke(tag_t<await_transform>, Promise& promise, _fn get_stop_token) noexcept {
+      return awaiter_{get_stop_token(promise)};
+    }
```
ispeters, line 54: I feel like you could do this more generically in the implementation of `await_transform`.  Every awaitable _Receiver_ query CPO is going to have the same shape.  Couldn't you add a case to [the default implementation](https://github.com/facebookexperimental/libunifex/blob/master/include/unifex/await_transform.hpp#L207) that looks like
```
else if constexpr (is_receiver_query_cpo_v<Value> && is_invocable<Value, Promise>) {
  return awaiter_{value(promise)};
}
```

You might need to tweak the `awaiter_` type a little if there's such a thing as a _Receiver_ query CPO that returns something immovable and uncopyable (maybe have `await_resume` return a reference to the value?), but it ought to otherwise work for every promise/CPO pair that the given promise supports.

include/unifex/get_stop_token.hpp
```diff
@@ -16,15 +16,44 @@
...
+
 #include <unifex/detail/prologue.hpp>
 
 namespace unifex {
 namespace _get_stop_token {
   inline const struct _fn {
+  private:
+#if !UNIFEX_NO_COROUTINES
+    template <typename StopToken>
+    struct awaiter_ {
```
ispeters, line 37: I don't know how important the naming convention is, here, but "private" Unifex types seem to _start_ with underscores rather than end with them...

test/task_cancel_test.cpp
```diff
@@ -51,11 +51,33 @@ task<int> bar() {
...
 task<void> void_test() {
   co_await stop;
   co_return;
 }
 
+bool didNotStopWhenNotAskedTo = false;
+
+task<int> test_stop_if_requested(inplace_stop_source& stopSource) {
+  co_await stop_if_requested; // shouldn't stop
+  didNotStopWhenNotAskedTo = true;
```
ispeters, line 74: The double negatives didn't leave me not confused.

How about `continuedWhenStopWasNotYetRequested`?

test/task_cancel_test.cpp
```diff
@@ -51,11 +51,33 @@ task<int> bar() {
...
+
+task<inplace_stop_token> get_token_outer() {
+  auto a = co_await get_stop_token;
+  auto b = co_await get_token_inner();
+  EXPECT_EQ(a, b);
+  co_return b;
+}
+
 task<void> void_test() {
   co_await stop;
```
ispeters, line 66: Looks like you've got a local `stop` defined; perhaps you should delete it so that this is using the thing you've added.

Also, I think `co_await stop;` and `co_await stop_if_requested;` look _weird_; it would be way more normal-looking to my eyes if you `co_await`ed the result of calling them: `co_await stop();` and `co_await stop_if_requested();`.  Did you have a reason for making them values instead of callables?

ispeters (approved):
LGTM

include/unifex/task.hpp
ericniebler: It occurs to me that if it's implemented as a sender algorithm, we don't get to benefit from symmetric transfer. I'm not sure whether that matters here or not.

test/task_cancel_test.cpp
ericniebler: > Did you have a reason for making them values instead of callables?

Slightly less typing for me and everybody else? That's it, really; no other reason.

include/unifex/get_stop_token.hpp
ericniebler: If I go with Lewis' suggestion and make it `co_await get_stop_token()`, then `Value` is whatever `get_stop_token()` returns. I would need to make that also a receiver query CPO for your code to work, and it would need to behave exactly like `get_stop_token`. Which makes me tempted to just make `get_stop_token()` return `get_stop_token`.

ericniebler: > You might need to tweak the awaiter_ type a little if there's such a thing as a Receiver query CPO that returns something immovable and uncopyable

I don't _think_ we have a receiver query that returns an immovable thing by value.

test/task_cancel_test.cpp
ispeters: It occurs to me that maybe I should get over the feeling that `co_await <value>;` is weird: it's the fact that it's being `co_await`ed that matters so the weirdness makes a different thing look different.  This might be important.  It feels related to "`std::move` doesn't move"; if it were `co_await stop();` then we'd have to say "`stop()` doesn't stop".  I suspect that no one would expect
```
task<void> foo() {
  stop;
}
```
to do anything, but they might expect
```
task<void> foo() {
  stop();
}
```
to do something.

include/unifex/get_stop_token.hpp
```diff
@@ -16,30 +16,67 @@
...
+    _awaiter(StopToken) -> _awaiter<StopToken>;
+
+    struct _awaitable {
+      template <typename Promise>
+      friend auto tag_invoke(tag_t<await_transform>, Promise& promise, _awaitable) noexcept {
+        return _awaiter{_fn{}(promise)};
+      }
+    };
+#endif
+  public:
```
ispeters, line 58: The `_awaitable`-returning `operator()` needs to be inside the `#if`; maybe this?

```suggestion
#if !UNIFEX_NO_COROUTINES
  private:
    template <typename StopToken>
    struct _awaiter {
      StopToken stoken_;
      bool await_ready() const noexcept {
        return true;
      }
      void await_suspend(coro::coroutine_handle<>) const noexcept {
      }
      StopToken await_resume() noexcept {
        return (StopToken&&) stoken_;
      }
    };
    template <typename StopToken>
    _awaiter(StopToken) -> _awaiter<StopToken>;

    struct _awaitable {
      template <typename Promise>
      friend auto tag_invoke(tag_t<await_transform>, Promise& promise, _awaitable) noexcept {
        return _awaiter{_fn{}(promise)};
      }
    };
    
   public:
    // `co_await get_stop_token()` to fetch a coroutine's current stop token.
    [[nodiscard]] constexpr _awaitable operator()() const noexcept {
      return {};
    }
#endif
```

PR #242, opened 2021-03-18
Enable CMake install and fix minor warnings

maikel:
This PR 

- fixes a CMake warning when calling `find_package(LibUring COMPONENTS)`
- add calls to CMakes `install( ... )` commands. I can't test this locally in case of liburing / coroutines, sorry.
- adds a default CMAKE_BUILD_TYPE based, some logic taken from #75 and  https://blog.kitware.com/cmake-and-the-default-build-type/
- change using gtest only if BUILD_TESTING is ON

This PR enables me to install this library such that  

```cmake
find_package(unifex REQUIRED)
add_library(MyLib INTERFACE)
# ...
target_link_libraries(AsyncMPI INTERFACE unifex::unifex)
```

works for me. 

I'm not a CMake expert myself, but maybe we can do this together!


ericniebler:
I admit this one is hard for me to review because I don't really understand cmake very well. I guess I can merge it and see who complains.

maikel:
I think there are definitively questionable parts, such as how to handle io_uring in the installation routines. On the other hand this PR also includes less questionable things such as just removing a CMake warning due to case sensitivity or disabling gtest if one only wants to build the examples but not the test.

Is it ok for you, if I split this PR into two or three PRs that are address those orthogonal issues? It might be easier to go about this.

Looking at the GitHub actions, I am also not sure if those need to be adjusted? Is the projects CMakeLists.txt even used? It seems to being overwritten by a hard-coded one.  I could also add actions, that test the installation process.

Thank you for your consideration.

ericniebler:
> Is it ok for you, if I split this PR into two or three PRs that are address those orthogonal issues? It might be easier to go about this.

By all means.

> Looking at the GitHub actions, I am also not sure if those need to be adjusted? Is the projects CMakeLists.txt even used?

Yes, it's used. Here in the libunifex-ci.yaml is where libunifex is getting configured with the project CMakeLists.txt:

```cmake
        execute_process(
          COMMAND ${{ steps.cmake_and_ninja.outputs.cmake_dir }}/cmake
            -S .
            -B build
            -D CMAKE_BUILD_TYPE=${{ matrix.config.build_type }}
            -G Ninja
            -D CMAKE_MAKE_PROGRAM=ninja
            -D CMAKE_VERBOSE_MAKEFILE=ON
            ${{ matrix.config.cmake_args }}
          RESULT_VARIABLE result
        )
```

> I could also add actions, that test the installation process.

I mean, that would be AWESOME if you're up for it.

PR #249, opened 2021-04-04
Get exception ptr CPO

Garcia6l20:
Create new  `get_exception_ptr` CPO allowing users to control how any kind of error type will be internally converted to an `std::exception_ptr`.


examples/linux/io_epoll_test.cpp
```diff
@@ -51,6 +56,18 @@ using namespace unifex;
...
 
+#if !UNIFEX_NO_COROUTINES
+namespace unifex
+{
+  // get_exception_ptr customization point for std::error_code
+  std::exception_ptr
+  tag_invoke(tag_t<unifex::get_exception_ptr>, std::error_code&& ec) noexcept {
+    return std::make_exception_ptr(
+        std::system_error{std::forward<std::error_code>(ec)});
+  }
```
ericniebler, line 67: This doesn't really belong here. Personally, I would bake it into the `get_exception_ptr` CPO itself since this is likely to be the behavior everybody is going to want to `std::error_code`.

examples/linux/io_epoll_test.cpp
```diff
@@ -155,6 +172,43 @@ int main() {
...
+              << ns << "ns, "
+              << reps << "ops\n";
+          std::cout
+              << "stats - "
+              << reads << "reads, "
+              << ns/reps << "ns-per-op, "
+              << reps/ms << "ops-per-ms\n";
+          stopWrite.request_stop();
+        }());
+#else
```
ericniebler, line 211: This is an awful lot of code duplication, even if it's "just a test". Would you be willing to factor out the bits that are common?

examples/linux/io_uring_test.cpp
```diff
@@ -40,6 +45,18 @@ using namespace unifex;
...
+namespace unifex
+{
+  // get_exception_ptr customization point for std::error_code
+  std::exception_ptr
+  tag_invoke(tag_t<unifex::get_exception_ptr>, std::error_code&& ec) noexcept {
+    return std::make_exception_ptr(
+        std::system_error{std::forward<std::error_code>(ec)});
+  }
+}  // namespace unifex
+#endif
```
ericniebler, line 58: Once this gets moved into the CPO proper, remove this.

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex
+{
+  inline const struct _get_exception_ptr_cpo {
+    template <typename ErrorCode>
+    constexpr auto operator()(ErrorCode&& error) const noexcept
+        -> tag_invoke_result_t<_get_exception_ptr_cpo, ErrorCode> {
+      return unifex::tag_invoke(*this, (ErrorCode &&) error);
+    }
```
ericniebler, line 32: Probably add an overload so that if its argument is already an `exception_ptr` it just returns it unmodified. I would also expect this to have a default implementation that accepts anything and passes it to `std::make_exception_ptr`. I'm curious if others share that opinion.

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <unifex/tag_invoke.hpp>
+#include <unifex/type_traits.hpp>
+
+#include <memory>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex
+{
+  inline const struct _get_exception_ptr_cpo {
```
ericniebler, line 27: I would rename this to simply `_fn` and move it into a nested `unifex::_get_exception_ptr` namespace. That way, we get to keep the `unifex` namespace unpolluted for ADL purposes.

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,40 @@
...
+    template <typename ErrorCode>
+    constexpr auto operator()(ErrorCode&& error) const noexcept
+        -> tag_invoke_result_t<_get_exception_ptr_cpo, ErrorCode> {
+      return unifex::tag_invoke(*this, (ErrorCode &&) error);
+    }
+  } get_exception_ptr{};
+
+  template <typename T>
+  constexpr bool is_exception_ptr_convertible_v =
+      is_callable_v<_get_exception_ptr_cpo, remove_cvref_t<T>>;
```
ericniebler, line 37: Once the contract has been widened, this will no longer be necessary.

include/unifex/await_transform.hpp
```diff
@@ -112,6 +113,14 @@ struct _awaitable_base<Promise, Value>::type {
...
       continuation_.resume();
     }
+
+    template (class Error)
+      (requires is_exception_ptr_convertible_v<Error>)
+    void set_error(Error &&error) && noexcept {
+      unifex::activate_union_member(result_->exception_, get_exception_ptr(std::forward<Error>(error)));
+      result_->state_ = _state::exception;
+      continuation_.resume();
+    }
```
ericniebler, line 123: If you widen the contract of `get_exception_ptr` as described below, then you only need one (unconstrained) `set_error` here that unconditionally passes the argument through `get_exception_ptr`.

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+namespace unifex
+{
+  inline const struct _get_exception_ptr_cpo {
+    template <typename ErrorCode>
+    constexpr auto operator()(ErrorCode&& error) const noexcept
+        -> tag_invoke_result_t<_get_exception_ptr_cpo, ErrorCode> {
+      return unifex::tag_invoke(*this, (ErrorCode &&) error);
+    }
+  } get_exception_ptr{};
```
ericniebler, line 33: And I would consider naming this `as_exception_ptr`.

ericniebler (changes requested):

include/unifex/get_exception_ptr.hpp
Garcia6l20: I also realized that there is an already existing `make_exception_ptr` CPO, that is commented to be a faster may to create an exception ptr than `std::make_exception_ptr`.
Maybe it could also be used for that purpose instead of creating a new CPO ?

examples/linux/io_epoll_test.cpp
Garcia6l20: Yes, that was just to show usage with coroutines.

I rolled back *io_uring_test.cpp* and *epoll_test.cpp* and created a new *file_coroutines_test.cpp* to show usage.
However It requires io_uring test since *file_concepts* are not implemented neither for *epoll_context* or *low_latency_iocp_context*.

include/unifex/get_exception_ptr.hpp
ericniebler: I don't think we want to change the meaning of `make_exception_ptr` or make it a customization point, but perhaps after you add `as_exception_ptr`, we can fold the behavior of `make_exception_ptr` into it and remove `make_exception_ptr`.

CMakeLists.txt
```diff
@@ -23,6 +23,9 @@ if(BUILD_TESTING OR UNIFEX_BUILD_EXAMPLES)
   include(CTest)
 
   target_compile_features(gtest PUBLIC cxx_std_17)
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL GNU)
+    target_compile_options(gtest PRIVATE "-Wno-error=maybe-uninitialized")
+  endif()
```
ericniebler, line 28: Why?

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+      // forward std::exception_ptr
+      std::exception_ptr operator()(std::exception_ptr &&eptr) const noexcept {
+        return std::forward<std::exception_ptr>(eptr);
+      }
+
+      // convert std::exception based types to std::exception_ptr
+      template(typename Exception)(
+          requires std::is_base_of_v<std::exception, Exception>)
+          std::exception_ptr
+          operator()(Exception &&ex) const noexcept {
```
ericniebler, line 41: Formatting.

```suggestion
      template(typename Exception)
          (requires std::is_base_of_v<std::exception, Exception>)
      std::exception_ptr operator()(Exception &&ex) const noexcept {
```

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+        return make_exception_ptr(std::forward<Exception>(ex));
+      }
+
+      // use customization point
+      // to resolve ErrorCode -> std::exception_ptr conversion
+      template(typename ErrorCode)
+          (requires is_tag_invocable_v<_fn, ErrorCode>)
```
ericniebler, line 48:
```suggestion
          (requires tag_invocable<_fn, ErrorCode>)
```

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex
+{
+  namespace _get_exception_ptr
+  {
+    inline const struct _fn {
+      // forward std::exception_ptr
+      std::exception_ptr operator()(std::exception_ptr &&eptr) const noexcept {
```
ericniebler, line 33: This will only accept rvalue `exception_ptr`s, which isn't what you want. I would just take the `exception_ptr` by value.

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      // use customization point
+      // to resolve ErrorCode -> std::exception_ptr conversion
+      template(typename ErrorCode)
+          (requires is_tag_invocable_v<_fn, ErrorCode>)
+      std::exception_ptr operator()(ErrorCode &&error) const noexcept {
+        return tag_invoke(*this, std::forward<ErrorCode>(error));
+      }
+    } get_exception_ptr{};
```
ericniebler, line 52:
```suggestion
    } as_exception_ptr{};
```

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+    // default std::error_code -> std::exception_ptr conversion
+    std::exception_ptr tag_invoke(tag_t<get_exception_ptr>, std::error_code &&error) noexcept {
+      return make_exception_ptr(
+          std::system_error{std::forward<std::error_code>(error)});
+    }
+
+  }  // namespace _get_exception_ptr_cpo
```
ericniebler, line 60: Fix the comment here.

include/unifex/get_stop_token.hpp
```diff
@@ -43,8 +43,11 @@ namespace _get_stop_token {
           return (StopToken&&) stoken_;
         }
       };
+#if !defined(__GNUC__) || (__GNUC__ < 10)
+      // GCC 10 (and 11) bug ? deduction guide must be declared at namespace scope
       template <typename StopToken>
       _awaiter(StopToken) -> _awaiter<StopToken>;
+#endif
```
ericniebler, line 50: Bummer. Maybe we should move the `_awaiter` out to namespace scope, then.

include/unifex/get_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+          operator()(Exception &&ex) const noexcept {
+        return make_exception_ptr(std::forward<Exception>(ex));
+      }
+
+      // use customization point
+      // to resolve ErrorCode -> std::exception_ptr conversion
+      template(typename ErrorCode)
+          (requires is_tag_invocable_v<_fn, ErrorCode>)
+      std::exception_ptr operator()(ErrorCode &&error) const noexcept {
+        return tag_invoke(*this, std::forward<ErrorCode>(error));
```
ericniebler, line 50: I think we should assert here that the selected `tag_invoke` overload is `noexcept` (although that is a tough contract to meet since most exceptions don't have `noexcept` constructors). I see this `operator()` is `noexcept` so we're getting termination if any `tag_invoke` overload for `get_exception_ptr` exits with an exception. Which, meh, happens in synchronous code when you try to throw an exception with another one active. Dunno. I'm undecided whether to push this trouble off on our users or not.

ericniebler (changes requested):

include/unifex/get_exception_ptr.hpp
Garcia6l20: Yes, clang-format is a bit confused with this template macro.

CMakeLists.txt
Garcia6l20: GoogleTest raises this gcc warning (with gcc >= 10 only). And as all warnings are treated as errors, I had to disable this to be treated as an error.
This was unexpected in this PR, I'll remove this.

include/unifex/get_stop_token.hpp
Garcia6l20: Should not be in this PR, sorry, I'll remove.
I'll probably propose another PR one for some fixes I made for GCC >= 10 support.

include/unifex/get_exception_ptr.hpp
Garcia6l20: My opinion is that we should assume that an `ErrorCode` to ` exception_ptr` must be `noexcept`, thus, a static_assert might be helpful (or a new `nothow_tag_invokable` concept).

include/unifex/as_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex
+{
+  namespace _as_exception_ptr
+  {
+    inline const struct _fn {
+      // forward std::exception_ptr
+      std::exception_ptr operator()(std::exception_ptr eptr) const noexcept {
+        return eptr;
```
Garcia6l20, line 34: Just to be sure...
Copy elision is guaranteed ? Or should I explicitly use `std::move` , I don't think so.

include/unifex/any_sender_of.hpp
```diff
@@ -70,6 +71,11 @@ struct _rec_ref<CPOs, Values...>::type
     : _rec_ref_base<CPOs>::template type<Values...>(*op)
     , stoken_(st) {}
 
+  template<typename Error>
+  void set_error(Error &&error) const noexcept {
```
lewissbaker, line 75: Having to add this sort of thing here makes me wonder whether it should be built in to the `set_error()` CPO itself...

e.g. try to call `set_error(r, e)` and if that fails then if you can call `set_error(r, exception_ptr)` and you can call `as_exception_ptr(e)` then do that, otherwise fail to find an overload.

examples/linux/file_coroutines_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+  inplace_stop_source stopSource;
+  std::thread t{[&] {
+    ctx.run(stopSource.get_token());
+  }};
+  scope_guard stopOnExit = [&]() noexcept {
+    stopSource.request_stop();
+    t.join();
+  };
+  sync_wait([&]() -> task<void> {
+    auto this_file = open_file_read_only(sched, filesystem::path{__FILE__});
```
lewissbaker, line 41: I'm a little nervous about using `__FILE__` here as the path may be relative (depending on the compiler and paths passed on the command-line) and the test may actually be run from a different directory / PWD.

Perhaps just write a dummy test file first and then try to read/copy it?

include/unifex/tag_invoke.hpp
```diff
@@ -104,6 +104,10 @@ namespace unifex {
       (sizeof(_tag_invoke::try_tag_invoke<CPO, Args...>(0)) ==
        sizeof(_tag_invoke::yes_type));
 
+  template <typename CPO, typename... Args>
+  UNIFEX_CONCEPT nothrow_tag_invocable =
+      (tag_invocable<CPO, Args...> && is_nothrow_tag_invocable_v<CPO, Args...>);
+
```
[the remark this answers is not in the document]
ericniebler: A concept has both syntax and semantics. This doesn't really have any semantics, so it feels more type trait-y to me.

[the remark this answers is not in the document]
Garcia6l20: In fact I've been a bit confused, why tag_invocable is not just:
```cpp
template <typename CPO, typename... Args>
UNIFEX_CONCEPT tag_invocable = is_tag_invocable_v<CPO, Args...>;
```
So, I was wondering if I was missing something, in doubt I used both requirements.

include/unifex/any_sender_of.hpp
Garcia6l20: This is what I did in the last update.
Now, `set_error()` handles this directly.
If the receiver accepts `std::exception_ptr` as error type and the given `Error` is convertible using `as_exception_ptr`, then it'll use it.
It also results in rollback of all `xxx::set_error<Error>(...)` previous modifications.

examples/linux/file_coroutines_test.cpp
Garcia6l20: Replaced with simple writes.

include/unifex/as_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+        return make_exception_ptr(std::forward<Exception>(ex));
+      }
+
+      // use customization point
+      // to resolve ErrorCode -> std::exception_ptr conversion
+      template(typename ErrorCode)
+          (requires nothrow_tag_invocable<_fn, ErrorCode>)
+      std::exception_ptr operator()(ErrorCode&& error) const noexcept {
+        return tag_invoke(*this, std::forward<ErrorCode>(error));
+      }
```
ericniebler, line 50: These last two overloads are potentially ambiguous: what if there is a `tag_invoke` overload for a type derived from `std::exception`? In that case, you should prefer the `tag_invoke` overload. Make the change I suggest above.

include/unifex/as_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+  {
+    inline constexpr struct _fn {
+      // forward std::exception_ptr
+      std::exception_ptr operator()(std::exception_ptr eptr) const noexcept {
+        return eptr;
+      }
+
+      // convert std::exception based types to std::exception_ptr
+      template(typename Exception)
+          (requires std::is_base_of_v<std::exception, Exception>)
```
ericniebler, line 39:
```suggestion
          (requires (!tag_invocable<_fn, Exception>) AND
              derived_from<Exception, std::exception>)
```

include/unifex/as_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+      std::exception_ptr operator()(Exception&& ex) const noexcept {
+        return make_exception_ptr(std::forward<Exception>(ex));
+      }
+
+      // use customization point
+      // to resolve ErrorCode -> std::exception_ptr conversion
+      template(typename ErrorCode)
+          (requires nothrow_tag_invocable<_fn, ErrorCode>)
```
ericniebler, line 47: Suggest you constrain on `tag_invocable` and then `static_assert` in the function body that the overload is `noexcept`.

```suggestion
          (requires tag_invocable<_fn, ErrorCode>)
```

include/unifex/as_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+        return tag_invoke(*this, std::forward<ErrorCode>(error));
+      }
+    } as_exception_ptr{};
+
+    // default std::error_code -> std::exception_ptr conversion
+    inline std::exception_ptr
+    tag_invoke(tag_t<as_exception_ptr>, std::error_code&& error) noexcept {
+      return make_exception_ptr(
+          std::system_error{std::forward<std::error_code>(error)});
+    }
```
ericniebler, line 58: I would prefer if this were an overload in `_fn` rather than a `tag_invoke` overload at namespace scope.

include/unifex/as_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+      std::exception_ptr operator()(ErrorCode&& error) const noexcept {
+        return tag_invoke(*this, std::forward<ErrorCode>(error));
+      }
+    } as_exception_ptr{};
+
+    // default std::error_code -> std::exception_ptr conversion
+    inline std::exception_ptr
+    tag_invoke(tag_t<as_exception_ptr>, std::error_code&& error) noexcept {
+      return make_exception_ptr(
+          std::system_error{std::forward<std::error_code>(error)});
```
ericniebler, line 57:
```suggestion
          std::system_error{(std::error_code&&) error});
```

include/unifex/as_exception_ptr.hpp
```diff
@@ -0,0 +1,65 @@
...
+      std::exception_ptr operator()(Exception&& ex) const noexcept {
+        return make_exception_ptr(std::forward<Exception>(ex));
+      }
+
+      // use customization point
+      // to resolve ErrorCode -> std::exception_ptr conversion
+      template(typename ErrorCode)
+          (requires nothrow_tag_invocable<_fn, ErrorCode>)
+      std::exception_ptr operator()(ErrorCode&& error) const noexcept {
+        return tag_invoke(*this, std::forward<ErrorCode>(error));
```
ericniebler, line 49: Compile times are bad enough without creating unnecessary `std::forward` instantiations. Elsewhere, we forward with C-style casts. Yes, it's poor form. You can use a `static_cast` if you prefer.

```suggestion
        return tag_invoke(*this, (ErrorCode&&) error);
```

include/unifex/receiver_concepts.hpp
```diff
@@ -116,14 +117,25 @@ namespace _rec_cpo {
           _set_error_fn{}, (Receiver &&) r, (Error&&) error);
     }
     template(typename Receiver, typename Error)
-      (requires (!tag_invocable<_set_error_fn, Receiver, Error>))
+      (requires (!tag_invocable<_set_error_fn, Receiver, Error>) &&
```
ericniebler, line 120: The `AND` macro lets us do short-circuiting in the concepts emulation layer.

```suggestion
      (requires (!tag_invocable<_set_error_fn, Receiver, Error>) AND
```

include/unifex/receiver_concepts.hpp
```diff
@@ -116,14 +117,25 @@ namespace _rec_cpo {
...
+         (!tag_invocable<decltype(as_exception_ptr), Error>))
     auto operator()(Receiver&& r, Error&& error) const noexcept
         -> _result_t<Receiver, Error> {
       static_assert(
           noexcept(static_cast<Receiver&&>(r).set_error((Error &&) error)),
           "receiver.set_error() method must be nothrow invocable");
       return static_cast<Receiver&&>(r).set_error((Error&&) error);
     }
+    template(typename Receiver, typename Error)
+      (requires (!tag_invocable<_set_error_fn, Receiver, Error>) &&
```
ericniebler, line 130:
```suggestion
      (requires (!tag_invocable<_set_error_fn, Receiver, Error>) AND
```

include/unifex/receiver_concepts.hpp
```diff
@@ -116,14 +117,25 @@ namespace _rec_cpo {
...
+         tag_invocable<decltype(as_exception_ptr), Error>)
+    auto operator()(Receiver&& r, Error&& error) const noexcept
+    -> _result_t<Receiver, std::exception_ptr> {
```
ericniebler, line 133: You need to update the `_result_t` alias to account for this overload. Look how it is done elsewhere with `if constexpr` in other customization points.

ericniebler (changes requested):

include/unifex/receiver_concepts.hpp
```diff
@@ -116,14 +117,25 @@ namespace _rec_cpo {
...
     auto operator()(Receiver&& r, Error&& error) const noexcept
         -> _result_t<Receiver, Error> {
       static_assert(
           noexcept(static_cast<Receiver&&>(r).set_error((Error &&) error)),
           "receiver.set_error() method must be nothrow invocable");
       return static_cast<Receiver&&>(r).set_error((Error&&) error);
     }
+    template(typename Receiver, typename Error)
+      (requires (!tag_invocable<_set_error_fn, Receiver, Error>) &&
+         tag_invocable<decltype(as_exception_ptr), Error>)
```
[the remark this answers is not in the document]
Garcia6l20: Yes, I missed it.
I added a test case to validate this.

include/unifex/receiver_concepts.hpp
Garcia6l20: I added a `_select` function (reproduced from `connect` CPO), I hope it is fine.

include/unifex/receiver_concepts.hpp
```diff
@@ -115,15 +124,32 @@ namespace _rec_cpo {
       return unifex::tag_invoke(
           _set_error_fn{}, (Receiver &&) r, (Error&&) error);
     }
+
+    // direct call
     template(typename Receiver, typename Error)
-      (requires (!tag_invocable<_set_error_fn, Receiver, Error>))
+      (requires (!tag_invocable<_set_error_fn, Receiver, Error>) AND
+         std::invocable<decltype(&Receiver::set_error), Receiver, Error>)
```
ericniebler, line 131: > ```c++
> std::invocable<decltype(&Receiver::set_error), Receiver, Error>
> ```

This is not the right way to check for a `.set_error(Error)` member function. Consider that `Receiver::set_error` could be a template. In that case, `&Receiver::set_error` is ill-formed. Check how such constraints are checked elsewhere; you'll need to test the validity of the expression `rec.set_error(err)` in a SFINAE context (or with a concept) to get a correct answer.

include/unifex/receiver_concepts.hpp
```diff
@@ -115,15 +124,32 @@ namespace _rec_cpo {
...
           noexcept(static_cast<Receiver&&>(r).set_error((Error &&) error)),
           "receiver.set_error() method must be nothrow invocable");
       return static_cast<Receiver&&>(r).set_error((Error&&) error);
     }
+
+    // call through as_exception_ptr
+    template(typename Receiver, typename Error)
+      (requires (!tag_invocable<_set_error_fn, Receiver, Error>) AND
+        (!std::invocable<decltype(&Receiver::set_error), Receiver, Error>) AND
+         std::invocable<decltype(&Receiver::set_error), Receiver, std::exception_ptr> AND
```
ericniebler, line 144: Same thing here about testing for a `.set_error(e)` member function.

ericniebler (changes requested):

PR #250, opened 2021-04-10
add any_scheduler

ericniebler:


ericniebler:
Ways I intend to eventually improve `any_scheduler` eventually:
* Implement the small object optimization in `any_unique`.
* Define an `any_scheduler_ref` CPO. The `task<>` promise can take advantage of the structured nature of coroutines and only store an `any_scheduler_ref` that refers to an `any_scheduler` that lives in `sender_task<>::promise_type`. (The promise type will need a separate customization of `get_scheduler` to return this scheduler reference.)
* Integrate `any_scheduler` with `with_receiver_queries` from `any_sender_of.hpp`, so that we can punch custom receiver queries though to a type-erased scheduler's sender.

I expect these can be done in separate PRs.

include/unifex/any_scheduler.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+  template <typename T>
+  type_index operator()(const T& x) const noexcept {
+    if constexpr (tag_invocable<get_type_id_fn, const T&>) {
+      return tag_invoke(*this, x);
+    } else {
+      return type_id<T>();
+    }
+  }
+} get_type_id{};
```
lewissbaker, line 62: Naming of this should be `get_typeid` to match the keyword?
Or alternatively `get_type_index`?

include/unifex/any_scheduler.hpp
```diff
@@ -0,0 +1,121 @@
...
+      return type_id<T>() == get_type_id(u.impl_) &&
+          t == *static_cast<const T*>(get_object_address(u.impl_));
+    }
+  }
+} equal_to{};
+
+using any_scheduler_impl =
+  any_unique_t<
+    overload<any_sender_of<>(this_ const&)>(schedule),
+    copy_as<any_scheduler>,
```
lewissbaker, line 79: I wonder if we need a new copyable type-erased wrapper here instead - one that actually hooks up the copy-constructor to require the type-erased object to be copyable rather than trying to build it around the `any_unique` type and it's move-only semantics

include/unifex/any_scheduler.hpp
```diff
@@ -0,0 +1,121 @@
...
+    (requires (!same_as<Scheduler, any_scheduler>) AND scheduler<Scheduler>)
+  /* implicit */ any_scheduler(Scheduler sched)
+    : impl_((Scheduler&&) sched) {}
+
+  any_scheduler(any_scheduler&&) noexcept = default;
+  any_scheduler(const any_scheduler& that)
+    : impl_(copy_as<any_scheduler>(that.impl_).impl_) {}
+
+  any_scheduler& operator=(any_scheduler&&) noexcept = default;
+  any_scheduler& operator=(const any_scheduler& that) {
```
lewissbaker, line 94: The fact that this can potentially throw potentially answers the question of whether or not executors/schedulers should be nothrow copyable I guess. The current executors wording requires executors to be nothrow copy constructible but scheduler concept does not require that.

It does kind of make me wonder whether this should be using some kind of reference-counting for copyability rather than allocating a whole new heap-allocated object, though.

Interestingly, though, none of the concepts require assignability so I assume you're adding assignability here to satisfy some need specific to the any_scheduler?

include/unifex/any_scheduler.hpp
```diff
@@ -0,0 +1,121 @@
...
+    return unifex::schedule(impl_);
+  }
+
+private:
+  friend equal_to_fn;
+  friend bool operator==(const any_scheduler& left, const any_scheduler& right) {
+    return equal_to(left.impl_, right);
+  }
+  friend bool operator!=(const any_scheduler& left, const any_scheduler& right) {
+    return !equal_to(left.impl_, right);
```
lewissbaker, line 109: Simplify to `return !(left == right);`?

include/unifex/any_unique.hpp
```diff
@@ -395,6 +395,21 @@ class _byval<CPOs...>::type
     , vtable_(other.vtable_) {}
 
   UNIFEX_ALWAYS_INLINE ~type() {
+    unsafe_deallocate();
+  }
+
+  type& operator=(type&& other) noexcept {
+    unsafe_deallocate();
+    impl_ = std::exchange(other.impl_, nullptr);
+    vtable_ = std::move(other.vtable_);
```
lewissbaker, line 404: If we implemented `swap()` for `any_unique` then this could just be:
```
type& operator=(type t) noexcept {
  swap(t);
  return *this;
}
```
and would also then be self-assignment safe.

test/any_scheduler_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <unifex/single_thread_context.hpp>
+#include <unifex/transform.hpp>
+#include <unifex/sync_wait.hpp>
+
+#include <gtest/gtest.h>
+
+using namespace unifex;
+
+TEST(AnySchedulerTest, SatisfiedConcept) {
+  EXPECT_TRUE(scheduler<any_scheduler>);
```
lewissbaker, line 29: This could be a static_assert

test/any_sender_of_test.cpp
```diff
@@ -20,6 +20,7 @@
 #include <unifex/ready_done_sender.hpp>
 #include <unifex/transform.hpp>
 #include <unifex/inline_scheduler.hpp>
+#include <unifex/any_scheduler.hpp>
```
lewissbaker, line 23: Is this `#include` needed here?

lewissbaker (changes requested):
I do worry about the use of `any_unique` to back the type-erased scheduler to be used in task. The fact that it heap-allocates on every copy will potentially mean that we'll get a heap-allocation every time someone calls `get_scheduler()` and gets a copy of the scheduler. I generally assume (and I think we are trying to educate people that) copying a scheduler is cheap and incurring a heap-allocation kind of violates that assumption.

It might be time to bite the bullet and implement a handful of other type-erasing wrappers.
I think a general any_scheduler might be better as a wrapper around a reference-counting `any_shared`, or perhaps even some kind of hybrid that either copies the value if it fits in a small-object-buffer or reference-counts if it needs to be heap-allocated.

Another option that might be viable for structured approaches is to use an `any_ref` type-erasing wrapper that just captures a pointer/reference to the concrete scheduler which is valid for the lifetime of the task. The adaptation from the parent scheduler could be done in a similar way to how we do things with `inplace_stop_token` for tasks.

Another thing to consider is having the `any_ref` type's `schedule()` implementation simply return a proxy that doesn't actually call `schedule()` on the underlying scheduler but instead defers calling `schedule()` until `connect()` time so that we don't need to type-erase the schedule-sender and can instead just type-erase the operation-state (possibly storing it inline with use of a small-object buffer big enough for most `schedule()` implementations.

include/unifex/any_scheduler.hpp
```diff
@@ -0,0 +1,121 @@
...
+  }
+} get_type_id{};
+
+inline constexpr struct equal_to_fn {
+  template <typename T, typename U>
+  bool operator()(const T& t, const U& u) const noexcept {
+    if constexpr (tag_invocable<equal_to_fn, const T&, const U&>) {
+      return tag_invoke(*this, t, u);
+    } else {
+      return type_id<T>() == get_type_id(u.impl_) &&
```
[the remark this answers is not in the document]
ericniebler: These CPOs are implementation details of `any_scheduler`.

include/unifex/any_scheduler.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+inline constexpr struct get_type_id_fn {
+  using type_erased_signature_t = type_index(const this_&) noexcept;
+
+  template <typename T>
+  type_index operator()(const T& x) const noexcept {
+    if constexpr (tag_invocable<get_type_id_fn, const T&>) {
+      return tag_invoke(*this, x);
+    } else {
+      return type_id<T>();
```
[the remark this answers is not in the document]
ericniebler: No, we use `unifex::type_index` to support `-fno-rtti` builds.

include/unifex/any_scheduler.hpp
ericniebler: OK.

include/unifex/any_scheduler.hpp
ericniebler: But why implement two types when one suffices? The type-erasing constructor of `any_unique` should SFINAE away if any of the CPOs cannot be applied to the type being type-erased. Together with a generic `copy_as` that is itself constrained is sufficient to meet the needs of a copyable type-erasing wrapper.

include/unifex/any_scheduler.hpp
ericniebler: I saw no reason to not provide assignability. The principle of least surprise.

test/any_scheduler_test.cpp
ericniebler: The test fails either way, and I prefer the gtest diagnostic.

test/any_sender_of_test.cpp
ericniebler: Yes.

ericniebler:
@lewissbaker Can you tell me why you don't think `any_unique` should provide the small object optimization? When would you prefer to use a type-erasing wrapper that _doesn't_ store small objects _in situ_ when it can?

test/any_sender_of_test.cpp
lewissbaker: Where is it needed? There was no other change to this file to add a usage of `any_scheduler` that I could see...
Is there some transitive dependency of the code under test below that requires `any_scheduler`?

lewissbaker: Ah, I see there seems to be some kind of cyclic dependency between `any_scheduler` and `any_sender_of`...

lewissbaker:
> Can you tell me why you don't think `any_unique` should provide the small object optimization? When would you prefer to use a type-erasing wrapper that doesn't store small objects in situ when it can?

I guess I have a few things that cause me hesitation:
- unless the small-object buffer is a pointer in size then it is going to grow the size of the any_unique wrapper object which may be undesirable in some cases
- it will force an extra indirection for the heap-allocated case - we'll need to pass a pointer to the small-object buffer to the type-erased function which will then need to load the pointer to the object from that location rather than loading the pointer directly
  this may or may not be much of a performance impact...
- we'd ideally like to maintain the noexcept-movability which would mean you couldn't store small objects inline that have throwing move constructors anyway.
- move constructors and move assignment operators become a virtual call instead of an inlinable move of the pointers
- some vague concerns about the address of the underlying object changing during a move breaking some use-cases
- some vague sense that it should have the same ownership/lifetime semantics as `unique_ptr` due to its naming

There is a much richer range of type-erasing wrappers that I can imagine, each supporting some combination of the cross-product of:
- copyability: non-moveable, move-only, copyable, nothrow copyable
- ownership: non-owning (ie. reference-semantics), unique ownership, shared ownership
- storage: heap-allocated, heap-allocated with small-objects inline, always inline
- other operators: equality comparison, partial ordering comparisons, assignment
- support for casting to narrower interfaces (subset of available CPOs)
- storage of vtable (inline vs pointer to static vtable)

The current `any_unique` implementation represents one useful point in that space.
The move-only + unique-ownership + heap-allocated with small-object buffer combination is another valid choice in that space but should perhaps have a different name...

I think for your use-case for any_scheduler you are probably wanting the combination: copyable +heap allocated with SOO + unique ownership if inline or shared ownership if heap-allocated + equality comparison.

include/unifex/any_scheduler.hpp
```diff
@@ -0,0 +1,185 @@
...
+          }
+        };
+    }
+
+  private:
+    friend type;
+    _sender(any_scheduler_impl<CPOs...> const& impl) noexcept
+      : impl_(impl)
+    {}
+    any_scheduler_impl<CPOs...> const& impl_;
```
ericniebler, line 160: This is technically a lifetime issue. There is no requirement that the scheduler remain alive until the result of `schedule` is connected. This needs a bit of thought. Ideally, we would be able to either:
* hold a strong reference to the impl when it is dynamically-allocated, or
* store a copy of the impl when it is stored in situ (one we implement that).

include/unifex/any_unique.hpp
```diff
@@ -450,13 +450,21 @@ class _byref<CPOs...>::type
...
+    , impl_((void*) std::addressof(impl)) {}
 
   void swap(type& other) noexcept {
     std::swap(vtable_, other.vtable_);
     std::swap(impl_, other.impl_);
   }
 
+  // Two any_ref's compare equal IFF they refer to the same object (shallow).
+  friend bool operator==(type const& left, type const& right) noexcept {
+    return left.impl_ == right.impl_;
```
lewissbaker, line 462: Do you also need to compare the vtables?
I'm just wondering about a case where you have an any_ref constructed from a ref-to-base and an any_ref constructed from ref-to-derived. These may have vtables that dispatch to different implementations and have different behaviour.

But then how do you compare vtables?
There is no guarantee that the function-pointers will be the same across different translation units.
I guess the vtable could store the type-id it was generated from as the first element and you could compare that. Might be a rabbit hole, though.

Do we really _need_ equality comparisons here?

ericniebler: Copy and assign is shallow, so equality comparison should be shallow also. That's part of regularity. And I provide it because it's hard to implement as a CPO, but cheap and easy for the type to provide it directly.

PR #254, opened 2021-04-13
Use gtest iff BUILD_TESTING=ON

maikel:
This PR disables gtest if one wants to build the examples but not the tests. 
In my case, I have issues compiling gtest with GCC 11.0.1 on Fedora 34 but this enables me to "test" the examples.


maikel:
Do we need `include(CTest)` ? Only if using `BUILD_TESTING` or for `UNIFEX_BUILD_EXAMPLES`, too?

ericniebler:
> Do we need `include(CTest)` ? Only if using `BUILD_TESTING` or for `UNIFEX_BUILD_EXAMPLES`, too?

If it works without it, then leave it out, I guess?

CMakeLists.txt
```diff
@@ -17,16 +17,18 @@ include(unifex_flags)
 add_subdirectory(source)
 
 if(BUILD_TESTING OR UNIFEX_BUILD_EXAMPLES)
+  # We enable testing in both, either if doing tests of build examples
```
ericniebler, line 20: Can you add to this comment _why_ we need to `enable_testing()` if we're not `BUILD_TESTING`?

Also, there's a small typo:
```suggestion
  # We enable testing in both, either if doing tests or build examples
```

maikel:
I think there is still a silent change in behvaour and there is no default BUILD_TESTING option anymore, which came from CTest. I need to read upon CTest and ask the cmake folk about how to use it correctly. The Github Actions currently do no execute the tests.

ericniebler:
Is this PR still a draft? I notices you took "WIP" out of the title.

maikel:
Yes this PR is still a draft. I just realized that there was no need for a special title to get into draft mode. I will hit ready for review if I've managed to put some time into it. 

I changed my job and I am still awaiting an answer from my employer if it is legal for me to contribute here, which is not clear to me even outside of my working time.

maikel:
From https://cmake.org/cmake/help/latest/module/CTest.html: (Up-to my own highlighted words)

> Include this module in the **top CMakeLists.txt file of a project** to enable testing with CTest and dashboard submissions to CDash

@ericniebler Would it be okay, to only build tests or examples if libunifex is a top-level project? Or would you prefer a `UNIFEX_BUILD_TESTING` option? I ask because CTest automatically defines a `BUILD_TESTING` option, which seems to be an analogue to C++ global names.

ericniebler:
> Would it be okay, to only build tests or examples if libunifex is a top-level project?

I think that's ok.

CMakeLists.txt
```diff
@@ -10,23 +10,31 @@ project(libunifex LANGUAGES CXX
...
 add_subdirectory(source)
 
-if(BUILD_TESTING OR UNIFEX_BUILD_EXAMPLES)
-  enable_testing()
-
-  include(gtest)
-  include(CTest)
+if(NOT PROJECT_IS_TOP_LEVEL)
+  return()
+endif()
```
ericniebler, line 22: This isn't quite what I was expecting. If I understand correctly, with this change, unifex will not build _at all_ unless it's the top level project. I don't want that. It's fine if examples and tests aren't built if it's not a top level project, but I still want to support building unifex in a subdirectory.

ericniebler (changes requested):

CMakeLists.txt
maikel: The library target unifex is still consumable by using `add_subdirectory` from another top level project. Note, that the library target is defined in call to `add_subdirectory(sources/)` prior to the early return. I will add a comment and remove the early return in favor of a longer if statement such as in`if (PROJECT_IS_TOP_LEVEL AND (BUILD_TESTING OR UNIFEX_BUILD_EXAMPLES))` etc...

ericniebler (approved):

ericniebler:
Merged. Thanks for all your work on this.

PR #256, opened 2021-04-14
add async_scope::spawn_call_on to simplify usage

ericniebler:
Also, swap the argument order to `unifex::on` (take the scheduler first), and disable pipe syntax for `on`.

### DIFF
--- a/include/unifex/async_scope.hpp
+++ b/include/unifex/async_scope.hpp
@@ -101,9 +101,9 @@ struct _receiver<Sender>::type final : _receiver_base {
 
 struct async_scope {
 private:
-  template <typename Sender, typename Scheduler>
+  template <typename Scheduler, typename Sender>
   using _on_result_t =
-    decltype(on(UNIFEX_DECLVAL(Sender&&), UNIFEX_DECLVAL(Scheduler&&)));
+    decltype(on(UNIFEX_DECLVAL(Scheduler&&), UNIFEX_DECLVAL(Sender&&)));
 
 public:
   async_scope() noexcept = default;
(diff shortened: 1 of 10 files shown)


include/unifex/async_scope.hpp
```diff
@@ -170,6 +170,14 @@ struct async_scope {
     }
   }
 
+  template (typename Fun, typename Scheduler)
+    (requires callable<Fun> && scheduler<Scheduler>)
```
lewissbaker, line 174: Use `AND` instead of `&&` for lazy evaluation.

include/unifex/async_scope.hpp
```diff
@@ -170,6 +170,14 @@ struct async_scope {
...
+  void spawn_with(Fun&& fun, Scheduler&& scheduler) {
+    spawn(
+      transform(schedule(), (Fun&&) fun),
+      (Scheduler&&) scheduler);
+  }
+
```
lewissbaker, line 180: I wonder whether this should constrain the function to be nothrow_invocable_v since it just terminates on exception/error anyway. Might be useful to require the callers to explicitly annotate their lambda/function to indicate that this is the case rather than it being hidden behaviour?
Thoughts?

include/unifex/async_scope.hpp
```diff
@@ -170,6 +170,14 @@ struct async_scope {
     }
   }
 
+  template (typename Fun, typename Scheduler)
+    (requires callable<Fun> && scheduler<Scheduler>)
+  void spawn_with(Fun&& fun, Scheduler&& scheduler) {
```
lewissbaker, line 175: I'm not in love with the name `spawn_with`.
The `with` part is presumably referring to the "scheduler" but this is not something that is different from the sender-form of `spawn()`.

Maybe something like `spawn_call(fn, scheduler)` or `spawn_call_on(fn, scheduler)` would better indicate that this is for callables/invocables?

ericniebler: What does the `_with` in `let_with` refer to? I thought it was referring to the lambda, which is why I used it here.

LeeHowes: Yes, let_with refers to the lambda. The meaning intended was more "let_with_the_value_from_the_lambda". let didn't itself refer to the function. Spawn feels like it does already refer to the function, though, so maybe that's why this seems different.

include/unifex/async_scope.hpp
```diff
@@ -170,6 +170,14 @@ struct async_scope {
...
+    spawn(
+      transform(schedule(), (Fun&&) fun),
+      (Scheduler&&) scheduler);
```
ispeters, line 178: Since you're invoking `spawn(sender, scheduler)` and not `spawn(sender)`, isn't it the case that you should either use `just()` instead of `schedule()` here, or use `schedule(scheduler)` but invoke the unary `spawn`?
```suggestion
    spawn(
      transform(just(), (Fun&&) fun),
      (Scheduler&&) scheduler);
```
```suggestion
    spawn(
      transform(schedule((Scheduler&&) scheduler), (Fun&&) fun));
```

include/unifex/async_scope.hpp
ispeters, line 175: What do you think of taking the callable second?  Code that looks like this reads better, I think:
```c++
scope.spawn_with(scheduler, []() noexcept {
  // something very long that makes the
  // lambda body break lines
});
```
while this is harder to read:
```c++
scope.spawn_with([]() noexcept {
  // something very long that makes the
  // lambda body break lines
}, scheduler);
```

ericniebler: My beef with this suggestion is that `on` takes the scheduler second, and `spawn_on` also takes the scheduler second. I would dislike the inconsistency more that I dislike putting the scheduler after the lambda.

LeeHowes: Maybe on should change, though. The only good reason to have the Sender first is for piping, but on(sched, blah | blah | blah) might actually be easier to read, knowing where something is going before what it is you are sending there.

ericniebler: I agree that we shouldn't support pipe syntax with on. It's seriously confusing.

I'd like to hear from @lewissbaker and @kirkshoop before I make that change, though.

ericniebler: For another data point, `via` takes the scheduler first. At the very least, we're being inconsistent here.

lewissbaker: via() should really be taking a scheduler rather than the schedule-sender.
We have finally() which does the equivalent of a via() but takes a sender instead of a scheduler.

PR #260, opened 2021-04-21
Fix transform_done _frcvr not sending value

### LINKED ISSUE #258: Issues regarding `transform_done`
My understanding of `transform_done` is that it preserves the input sender's value and error, and produces the result of the given sender-returning function. Thus
```cpp
just_done()
| transform_done([] { return just(42); })
| sync_wait()
```
should produce 42 (wrapped in an `optional<int>`), but currently, it produces 0 for some reasons.

Is my understanding of the `transform_done` algorithm off, or is it an issue in the implementation?

Also a related topic, how could I use `transform_done` together with `transform` so that I can implement something like transforming a sender that sends `T` or done into a sender that sends `optional<T>`?

Thanks.

Chlorie:
Turns out it was just a little mistake.

Resolves #258

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build/
 .vscode/
 .vs/
 /CMakeSettings.json
+/Folder.DotSettings.user
--- a/include/unifex/transform_done.hpp
+++ b/include/unifex/transform_done.hpp
@@ -162,7 +162,7 @@ public:
   void set_value(Values&&... values) noexcept(
       is_nothrow_receiver_of_v<Receiver, Values...>) {
     UNIFEX_ASSERT(op_ != nullptr);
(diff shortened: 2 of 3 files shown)


Chlorie:
The tests for the `transform_done` algorithm don't cover the case where the sender actually sends some values other than the default values, maybe that's why this issue went undiscovered.

ericniebler (changes requested):
Nice catch. Think you  could add a test case that covers this?

Chlorie:
Added a simple test.

Visual Studio always adds a `Folder.DotSettings.user` file each time I run the tests and it's annoying, so I added that file into `.gitignore`. I guess this is ok?

lewissbaker (approved):
lgtm. Thanks for the fix!

PR #261, opened 2021-04-21
spawn work on correct context, add unary spawn overload

ericniebler:
`spawn`-ing work with a scheduler was not actually launching the work on that scheduler! Fix the bug, and also add a unary overload of `spawn` that doesn't take a scheduler operation. The binary form is trivially implementable in terms of `spawn` and `on`.

### DIFF
--- a/include/unifex/single_thread_context.hpp
+++ b/include/unifex/single_thread_context.hpp
@@ -39,7 +39,7 @@ public:
     return loop_.get_scheduler();
   }
 
-  std::thread::id get_id() const noexcept {
+  std::thread::id get_thread_id() const noexcept {
     return thread_.get_id();
   }
 };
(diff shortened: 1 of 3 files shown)


ispeters:
I mentioned offline that I think it's worth adding a test that _would_ have failed with the previous, buggy implementation of `spawn` to confirm that it _now_ passes.  Other than that, looks good to me.

include/unifex/single_thread_context.hpp
```diff
@@ -38,6 +38,10 @@ class context {
   auto get_scheduler() noexcept {
     return loop_.get_scheduler();
   }
+
+  std::thread::id get_id() const noexcept {
```
ispeters, line 42: Should this be `get_thread_id`?
```suggestion
  std::thread::id get_thread_id() const noexcept {
```

ispeters (approved):
Nice!

PR #264, opened 2021-04-22
add just_with() and defer() sender factories

ericniebler:


doc/api_reference.md
```diff
@@ -154,13 +154,29 @@ Returns a sender that completes synchronously by calling `set_done()`.
...
 Returns a sender that completes synchronously by calling `set_error()` with `e`.
 
+### `just_with(callable)`
+
+Returns a sender that completes synchronously by calling `set_value()` with
+the result of invoking the callable with no arguments. If the callable returns
+`void`, then the sender completes synchronously by calling `set_value()` with
+no arguments.
+
+`just_with(callable)` is synonymous with `transform(just(), callable)`.
```
LeeHowes, line 164: Can just_with support non-moveable types? let_with can do that because it stores the returned value in the operation state.

LeeHowes: Or at least, presumably it can't if it is implemented as transform, but should it be able to?

ispeters: I think `transform` can pass immovable types to `set_value` when they're returned from the given callable: https://github.com/ispeters/libunifex/blob/master/include/unifex/transform.hpp#L83.

ispeters:
Your new documentation doesn't mention anything about what happens if the lambdas passed to either `just_with` or `defer` throw an exception, which feels like a miss.

Looks good to me otherwise.

PR #268, opened 2021-04-27
make any_unique movable and swappable; fix under-constrained any_unique ctor

ericniebler:


include/unifex/any_unique.hpp
```diff
@@ -433,9 +453,18 @@ class _byref<CPOs...>::type
     : vtable_(vtable_holder_t::template create<Concrete>())
     , impl_(std::addressof(impl)) {}
 
+  void swap(type& other) noexcept {
+    std::swap(vtable_, other.vtable_);
+    std::swap(impl_, other.impl_);
```
ispeters, line 458: I suppose you're not doing the two-step since you're not in a generic context and you know you want the `std` version?

ispeters (approved):
Seems like a used-to-be-broken test would be a good addition.

Looks good to me.

include/unifex/any_unique.hpp
ericniebler: Yes.

ericniebler:
> Seems like a used-to-be-broken test would be a good addition.

That's why I added the `static_assert` that `any_unique` satisfies the `movable` concept. That would have failed previously.

PR #271, opened 2021-04-28
Fix missing config.hpp include on unifex_fwd

mizvekov:
unifex_fwd uses UNIFEX_NO_COROUTINES, so needs to include config.hpp.

Signed-off-by: Matheus Izvekov <mizvekov@gmail.com>


ericniebler:
Thanks!

PR #272, opened 2021-04-29
unsubscribe from task's inplace stop token adaptor on unhandled done

ericniebler:


ericniebler:
> Is the new behaviour testable?

In fact, yes. I added a test and confirmed that it failed before the fix.

ericniebler:
I feel good enough about the change to merge it without waiting. I mean, it's certainly an improvement. If anybody disagrees, holla and I'll fix differently.

Issue #18, opened 2019-11-13
Merge via() and typed_via()


lewissbaker:
The typed_via() algorithm should be incorporated as an overload of the via() algorithm rather than having a separate name.

The implementation should switch based on whether or not the sender is a typed-sender or not.

PR #273, opened 2021-04-30
change via() to take a scheduler instead of a successor sender

ericniebler:


lewissbaker:
Note that probably what you want to use is the `typed_via()` algorithm which does not incur a heap-allocation to store the intermediate result and operation-state of the predecessor.

There is a long-standing issue #18 to merge these two algorithms.
Which should be easier now that they have the same interface.

PR #277, opened 2021-05-07
co_invoke review feedback from @lewissbaker

ericniebler:


include/unifex/async_invoke.hpp
```diff
@@ -17,49 +17,43 @@
...
           Args...> {
       return tag_invoke(
           *this,
-          tag<std::invoke_result_t<Fn, Args...>, Fn, Args...>{},
+          type_identity<std::invoke_result_t<Fn, Args...>>{},
           (Fn&&) fn,
           (Args&&) args...);
     }
-} co_invoke{};
+} async_invoke{};
```
lewissbaker, line 53: It's not necessarily "async", though is it?
I can imagine we might customise this for `generator` to also allow capturing args by value?

Maybe `copy_invoke` or `by_value_invoke`?
Or `no_dangling_refs_invoke`? ;)

ericniebler: Is this just a flavor of `let`?

test/async_invoke_test.cpp
```diff
@@ -29,39 +29,61 @@
...
+  std::optional<int> result = sync_wait(std::move(t));
+  ASSERT_TRUE(!!result);
+  EXPECT_EQ(*result, 100);
+}
+
+TEST(CoInvoke, WithLvalueArgumentsWithByRefCaptures) {
+  int i = 42;
+  task<int> t = [&]() {
+    int arg = 58;
+    return async_invoke([&](int j) -> task<int> { co_return i + j; }, arg);
```
lewissbaker, line 85: Can you also add a test for passing the lambda itself as an lvalue?
e.g. a case where a lambda declared outside of a loop is used multiple times with different args?

ericniebler:
I will add the requested test case. We're still stuck on the naming issue, though. Do you have any suggestions, @lewissbaker ?

include/unifex/async_invoke.hpp
lewissbaker: Yes, I think you can kind of do the same thing with `let`.

```
co_invoke(f, args...)
```
is roughly equivalent to
```
let(just(args...), f)
```

PR #281, opened 2021-05-14
at_coroutine_exit for registering an async cleanup action to run automatically at coroutine exit

ericniebler:


test/at_coroutine_exit_test.cpp
```diff
@@ -0,0 +1,239 @@
...
+    ++result;
+    co_await at_coroutine_exit([this]() -> task<void> { result *= 2; co_return; });
+    co_await stop();
+    ++result;
+  }
+
+  task<void> test_two_cleanup_actions_with_stop() {
+    ++result;
+    co_await at_coroutine_exit([this]() -> task<void> { result *= 2; co_return; });
+    co_await at_coroutine_exit([this]() -> task<void> { result += result; co_return; });
```
LeeHowes, line 59: Do you want to have *=result like above so that order matters here?

include/unifex/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,283 @@
...
+  ~_cleanup_task() {
+    UNIFEX_ASSERT(!continuation_);
+  }
+
+  bool await_ready() const noexcept {
+    return false;
+  }
+
+  template <typename Promise>
+  bool await_suspend(coro::coroutine_handle<Promise> parent) noexcept {
```
LeeHowes, line 249: If I understand correctly, each time you co_await a cleanup task it inserts itself in the continuation chain.

The cleanup actions run in *reverse* order, like destructors? I think that is right from the code, and from the behaviour of test_two_cleanup_actions, but the docblock doesn't make that clear.

include/unifex/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,283 @@
...
+private:
+  continuation_handle<promise_type> continuation_;
+};
+
+namespace _at_coroutine_exit {
+  inline constexpr struct _fn {
+  private:
+    template <typename Action, typename... Ts>
+    static _cleanup_task<Ts...> at_coroutine_exit(Action action, Ts... ts) {
+      co_await std::move(action)(std::move(ts)...);
```
lewissbaker, line 268: Should we be checking that the result of the `co_await` expression is `void` here?
Or are you just relying on `[[nodiscard]]` warnings if values are discarded?

lewissbaker (approved):
Generally looks good!
A few minor comments/nits.

include/unifex/continuations.hpp
```diff
@@ -0,0 +1,330 @@
...
+      });
+}
+
+template <typename Promise>
+coro::coroutine_handle<> _done_callback_for(void* address) noexcept {
+  return coro::coroutine_handle<Promise>::from_address(address).promise().unhandled_done();
+}
+
+template <typename Promise>
+inline static constexpr _continuation_handle_vtable _vtable_for {
```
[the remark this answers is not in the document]
ericniebler: Oops, yes. Nice spot.

include/unifex/continuations.hpp
```diff
@@ -0,0 +1,330 @@
...
+
+template <>
+struct continuation_handle<void> {
+  continuation_handle() = default;
+
+  template (typename Promise)
+    (requires (!same_as<Promise, void>))
+  /*implicit*/ continuation_handle(coro::coroutine_handle<Promise> continuation) noexcept
+    : continuation_handle(0, (coro::coroutine_handle<Promise>&&) continuation)
+  {}
```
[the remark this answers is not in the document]
ericniebler: It works around a problem caused by the concepts emulation macros. When using the macros, it's not possible to forward-declare a constrained function and then define it later. So this defines a constrained thunk _in situ_ which just forwards to an unconstrained function defined out-of-line.

All of that deserves a comment in the code, at the very least.

include/unifex/at_coroutine_exit.hpp
ericniebler: This is just my opinion, but it feels overly nannying to enforce that the function returns a void sender.

PR #286, opened 2021-05-20
Update googletest to fix some compile errors under clang trunk

lewissbaker:
Upgrade googletest to fix a warning/error under clang-trunk that was complaining
about use of an implicit copy constructor when the copy-assignment operator had
been deleted.

Also removed a spurious `#define M(ARG)` macro in `concept_macros.hpp` that was causing problems with the latest googletest which was using `M` as a template parameter.


ericniebler:
Great, thanks @lewissbaker !

PR #287, opened 2021-05-24
Fix race condition on notifyingThreadId_ in inplace_stop_source::remove_callback

ericniebler:


ispeters:
LGTM

source/inplace_stop_token.cpp
```diff
@@ -150,11 +150,12 @@ void inplace_stop_source::remove_callback(
     }
     unlock(oldState);
   } else {
+    auto notifyingThreadId = notifyingThreadId_;
     unlock(oldState);
 
     // Callback has either already been executed or is
     // currently executing on another thread.
-    if (std::this_thread::get_id() == notifyingThreadId_) {
+    if (std::this_thread::get_id() == notifyingThreadId) {
```
lewissbaker, line 158: I'm reasonably sure that the old behaviour was ok.

We would only get into this branch if the callback has been removed by a thread calling `request_stop()`.
The thread that called `request_stop()` will have assigned its thread-id to `notifyingThreadId_` before dequeuing this callback and subsequently releasing the lock. The unlock operation then synchronises with this thread's lock operation when it sees that the callback has been dequeued already. Thus the write to `notifyingThreadId_` should "happen before" the read from `notifyingThreadId_` here.

Having said that, we hold the lock already so there's not really any harm in doing the read while holding the lock...

PR #290, opened 2021-06-05
task<> scheduler affinity

ericniebler:
This PR makes `unifex::task<>` scheduler-aware. It inherits and propagates the current scheduler just as it does with stop tokens. When awaiting arbitrary awaitables and senders, it wraps the awaitable/sender in `typed_via` to thunk back onto the current scheduler; that is, unless the awaitable/scheduler responds to a `blocking` query with a compile-time `blocking_kind::always_inline`.

`co_await schedule(sched)` is magical. It transitions the coroutine onto the execution context of `sched` and registers a cleanup action to put the coroutine back onto the original scheduler before resuming the parent coroutine. When awaiting another `task<>`, we assume that we will be resumed on the correct context and do not insert an additional thunk.

Still TODO:
- [x] Lots of tests
- [x] Fix MSVC build
- [ ] Docs

Incidentally, I am aware of a situation that breaks `task<>` scheduler affinity: if you register an async cleanup action with `at_coroutine_exit` that changes execution context, the original context is not automatically restored. For now, my recommendation is: don't do that.

### DIFF
--- a/include/unifex/at_coroutine_exit.hpp
+++ b/include/unifex/at_coroutine_exit.hpp
@@ -23,6 +23,7 @@
 #include <unifex/stop_token_concepts.hpp>
 #include <unifex/inline_scheduler.hpp>
 #include <unifex/any_scheduler.hpp>
+#include <unifex/blocking.hpp>
 
 #if UNIFEX_NO_COROUTINES
 # error "Coroutine support is required to use this header"
@@ -275,6 +276,10 @@ struct [[nodiscard]] _cleanup_task {
 
   std::tuple<Ts&...> await_resume() noexcept {
     return std::exchange(continuation_, {}).promise().args_;
+  }
+
+  friend auto tag_invoke(tag_t<blocking>, const _cleanup_task&) noexcept {
+    return blocking_kind::always_inline;
   }
 
 private:
--- a/include/unifex/at_coroutine_exit.hpp
+++ b/include/unifex/at_coroutine_exit.hpp
@@ -278,7 +278,7 @@ struct [[nodiscard]] _cleanup_task {
     return std::exchange(continuation_, {}).promise().args_;
   }
 
-  friend auto tag_invoke(tag_t<blocking>, const _cleanup_task&) noexcept {
+  friend constexpr auto tag_invoke(tag_t<blocking>, const _cleanup_task&) noexcept {
     return blocking_kind::always_inline;
   }
 
--- a/include/unifex/at_coroutine_exit.hpp
+++ b/include/unifex/at_coroutine_exit.hpp
@@ -19,7 +19,9 @@
 #include <unifex/tag_invoke.hpp>
 #include <unifex/await_transform.hpp>
 #include <unifex/continuations.hpp>
+#include <unifex/scheduler_concepts.hpp>
 #include <unifex/stop_token_concepts.hpp>
+#include <unifex/inline_scheduler.hpp>
 
 #if UNIFEX_NO_COROUTINES
 # error "Coroutine support is required to use this header"
@@ -110,8 +112,15 @@ struct _cleanup_promise_base {
     visit_continuations(p.continuation_, (Func &&) func);
   }
 
-  friend unstoppable_token tag_invoke(tag_t<get_stop_token>, const _cleanup_promise_base&) noexcept {
+  friend unstoppable_token
+  tag_invoke(tag_t<get_stop_token>, const _cleanup_promise_base&) noexcept {
     return unstoppable_token{};
+  }
+
+  // BUGBUG should cleanup actions inherit the current scheduler from the parent coroutine?
+  friend inline_scheduler
+  tag_invoke(tag_t<get_scheduler>, const _cleanup_promise_base&) noexcept {
+    return inline_scheduler{};
   }
 
   continuation_handle<> continuation_{};
@@ -139,6 +148,15 @@ struct _die_on_done_rec {
     [[noreturn]] void set_done() && noexcept {
       UNIFEX_ASSERT(!"A cleanup action tried to cancel. Calling terminate...");
       std::terminate();
+    }
+
+    template(typename CPO)
+      (requires is_receiver_query_cpo_v<CPO> AND
+                is_callable_v<CPO, const Receiver&>)
+    friend auto tag_invoke(CPO cpo, const type& p)
+        noexcept(is_nothrow_callable_v<CPO, const Receiver&>)
+        -> callable_result_t<CPO, const Receiver&> {
+      return cpo(p.rec_);
     }
   };
 };
--- a/include/unifex/at_coroutine_exit.hpp
+++ b/include/unifex/at_coroutine_exit.hpp
@@ -22,6 +22,7 @@
 #include <unifex/scheduler_concepts.hpp>
 #include <unifex/stop_token_concepts.hpp>
 #include <unifex/inline_scheduler.hpp>
+#include <unifex/any_scheduler.hpp>
 
 #if UNIFEX_NO_COROUTINES
 # error "Coroutine support is required to use this header"
@@ -117,13 +118,14 @@ struct _cleanup_promise_base {
     return unstoppable_token{};
   }
 
-  // BUGBUG should cleanup actions inherit the current scheduler from the parent coroutine?
-  friend inline_scheduler
-  tag_invoke(tag_t<get_scheduler>, const _cleanup_promise_base&) noexcept {
-    return inline_scheduler{};
+  friend any_scheduler_ref
+  tag_invoke(tag_t<get_scheduler>, const _cleanup_promise_base& p) noexcept {
+    return p.sched_;
   }
 
+  inline static constexpr inline_scheduler _default_scheduler{};
   continuation_handle<> continuation_{};
+  any_scheduler_ref sched_{_default_scheduler};
   bool isUnhandledDone_{false};
 };
 
@@ -267,6 +269,7 @@ struct [[nodiscard]] _cleanup_task {
   bool await_suspend(coro::coroutine_handle<Promise> parent) noexcept {
     continuation_.promise().continuation_ =
         exchange_continuation(parent.promise(), continuation_);
+    continuation_.promise().sched_ = get_scheduler(parent.promise());
     return false;
   }
 
--- a/include/unifex/at_coroutine_exit.hpp
+++ b/include/unifex/at_coroutine_exit.hpp
@@ -193,7 +193,7 @@ struct _die_on_done {
           _die_on_done_rec_t<Receiver>{(Receiver&&) rec});
     }
 
-    Sender sender_;
+    UNIFEX_NO_UNIQUE_ADDRESS Sender sender_;
   };
 };
 
@@ -245,7 +245,7 @@ struct _cleanup_promise : _cleanup_promise_base {
     return unifex::await_transform(*this, _die_on_done_fn{}((Value&&) value));
   }
 
-  std::tuple<Ts&...> args_;
+  UNIFEX_NO_UNIQUE_ADDRESS std::tuple<Ts&...> args_;
 };
 
 template <typename... Ts>
@@ -296,7 +296,7 @@ namespace _at_coroutine_exit {
   public:
     template (typename Action, typename... Ts)
       (requires callable<std::decay_t<Action>, std::decay_t<Ts>...>)
-    _cleanup_task<Ts...> operator()(Action&& action, Ts&&... ts) const {
+    _cleanup_task<std::decay_t<Ts>...> operator()(Action&& action, Ts&&... ts) const {
       return _fn::at_coroutine_exit((Action&&) action, (Ts&&) ts...);
     }
   } at_coroutine_exit{};
--- a/include/unifex/await_transform.hpp
+++ b/include/unifex/await_transform.hpp
@@ -27,6 +27,7 @@
 #include <unifex/sender_concepts.hpp>
 #include <unifex/type_traits.hpp>
 #include <unifex/manual_lifetime.hpp>
+#include <unifex/typed_via.hpp>
 
 #include <exception>
 #include <optional>
@@ -85,13 +86,13 @@ struct _awaitable_base<Promise, Value>::type {
   struct _rec {
   public:
     explicit _rec(_expected<Value>* result, coro::coroutine_handle<Promise> continuation) noexcept
-    : result_(result)
-    , continuation_(continuation)
+      : result_(result)
+      , continuation_(continuation)
     {}
 
     _rec(_rec&& r) noexcept
-    : result_(std::exchange(r.result_, nullptr))
-    , continuation_(std::exchange(r.continuation_, nullptr))
+      : result_(std::exchange(r.result_, nullptr))
+      , continuation_(std::exchange(r.continuation_, nullptr))
     {}
 
     template(class... Us)
@@ -201,14 +202,19 @@ inline constexpr struct _fn {
     // to avoid instantiating 'unifex::sender<Value>' concept check in
     // the case that _awaitable<Value> evaluates to true.
     if constexpr (detail::_awaitable<Value>) {
+      // TODO: We need to wrap awaitables in typed_via also
       return (Value&&) value;
     } else if constexpr (unifex::sender<Value>) {
-      if constexpr (unifex::sender_to<Value, _receiver_t<Promise, Value>>) {
+      // Wrap the sender in a via so that we always transition back to the proper
+      // execution context.
+      using ViaSender = typed_via_result_t<Value, get_scheduler_result_t<Promise&>>;
+      if constexpr (unifex::sender_to<ViaSender, _receiver_t<Promise, ViaSender>>) {
         auto h = coro::coroutine_handle<Promise>::from_promise(promise);
-        return _as_awaitable<Promise, Value>{(Value&&) value, h};
+        auto viaSender = typed_via((Value&&) value, get_scheduler(promise));
+        return _as_awaitable<Promise, ViaSender>{(ViaSender&&) viaSender, h};
       } else {
         static_assert(
-          unifex::sender_to<Value, _receiver_t<Promise, Value>>,
+          unifex::sender_to<ViaSender, _receiver_t<Promise, ViaSender>>,
           "This sender is not awaitable in this coroutine type.");
         return (Value&&) value;
       }
--- a/include/unifex/await_transform.hpp
+++ b/include/unifex/await_transform.hpp
@@ -190,6 +190,9 @@ struct _fn {
   auto operator()(Promise& promise, Value&& value) const
     noexcept(is_nothrow_tag_invocable_v<_fn, Promise&, Value>)
     -> tag_invoke_result_t<_fn, Promise&, Value> {
+    static_assert(detail::_awaitable<tag_invoke_result_t<_fn, Promise&, Value>>,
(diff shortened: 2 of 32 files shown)


include/unifex/finally.hpp
```diff
@@ -696,36 +716,50 @@ namespace unifex
                 static_cast<Receiver&&>(r)};
       }
 
+      friend constexpr auto tag_invoke(tag_t<unifex::blocking>, const type& self) noexcept {
+        if constexpr (
```
ispeters, line 720: This is non-obvious. Why do you have this `if constexpr`?

ispeters (commented):
Some initial thoughts from my phone; hopefully reviewing in the Android app actually works.

include/unifex/at_coroutine_exit.hpp
```diff
@@ -249,13 +270,18 @@ struct [[nodiscard]] _cleanup_task {
...
         exchange_continuation(parent.promise(), continuation_);
+    continuation_.promise().sched_ = get_scheduler(parent.promise());
     return false;
   }
 
   std::tuple<Ts&...> await_resume() noexcept {
     return std::exchange(continuation_, {}).promise().args_;
   }
 
+  friend auto tag_invoke(tag_t<blocking>, const _cleanup_task&) noexcept {
```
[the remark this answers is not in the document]
ericniebler: Yes, but `cblocking` will work anyway because it only checks return type, not return value.

include/unifex/finally.hpp
ericniebler: In the two branches, the return types are different. When the condition is true, we have access to the blocking property at compile time and can return a compile-time constant. (`blocking_kind::constant` is an alias for an instantiation of `std::integral_constant`.) When the condition is false, we fall back to a runtime computation and return the result as a `blocking_kind`, which is a wrapper around an ordinary `enum`.

include/unifex/task.hpp
```diff
@@ -172,20 +173,21 @@ struct _promise {
...
+      if constexpr (derived_from<remove_cvref_t<Value>, _task_base> ||
+          blocking_kind::always_inline == cblocking<Value>()) {
+        // We are co_await-ing a unifex::task or something that completes inline. We don't
+        // need an additional transition.
         return unifex::await_transform(*this, (Value&&) value);
       } else if constexpr (tag_invocable<tag_t<unifex::await_transform>, type&, Value>
           || detail::_awaitable<Value>) {
         // Either await_transform has been customized or Value is an awaitable. Either
         // way, we can dispatch to the await_transform CPO, then insert a transition back
         // to the correct execution context.
```
ispeters: Is it possible for the result of a customized `await_transform` to communicate that it completes inline?

include/unifex/task.hpp
```diff
@@ -195,23 +209,31 @@ struct _awaiter {
...
+      sched_.destruct();
+      auto thisCoro = coro::coroutine_handle<ThisPromise>::from_address(
+          (void*) std::exchange(--coro_, 0));
+      scope_guard destroyOnExit{[&]() noexcept { thisCoro.destroy(); }};
+      return thisCoro.promise().result();
     }
 
   private:
-    coro::coroutine_handle<ThisPromise> coro_;
+    std::uintptr_t coro_; // Stored as an integer so we can use the low bit as a dirty bit
```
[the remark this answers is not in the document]
ericniebler: > If you get into the falsey | 1 state, I think you'll try to `.destroy()` an invalid `coroutine_handle<>`.

I don't think it's possible to get into this state.

> are trying to make it defined behaviour to construct and destruct an awaiter without awaiting it (why?)

Because it is possible to call `get_awaiter(task)` and ignore the result. Unlikely, but possible.

> need to destroy the `coroutine_handle<>` if it's truthy even if the dirty bit is false, and destroy the other state when the dirty bit is true.

Good catch.

include/unifex/await_transform.hpp
```diff
@@ -201,14 +202,19 @@ inline constexpr struct _fn {
...
     } else if constexpr (unifex::sender<Value>) {
-      if constexpr (unifex::sender_to<Value, _receiver_t<Promise, Value>>) {
+      // Wrap the sender in a via so that we always transition back to the proper
+      // execution context.
+      using ViaSender = typed_via_result_t<Value, get_scheduler_result_t<Promise&>>;
+      if constexpr (unifex::sender_to<ViaSender, _receiver_t<Promise, ViaSender>>) {
         auto h = coro::coroutine_handle<Promise>::from_promise(promise);
-        return _as_awaitable<Promise, Value>{(Value&&) value, h};
+        auto viaSender = typed_via((Value&&) value, get_scheduler(promise));
+        return _as_awaitable<Promise, ViaSender>{(ViaSender&&) viaSender, h};
```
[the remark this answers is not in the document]
ericniebler: Not sure what you mean. Care to elaborate?

include/unifex/task.hpp
ericniebler: I suppose so. I should account for that.

include/unifex/at_coroutine_exit.hpp
```diff
@@ -140,6 +149,15 @@ struct _die_on_done_rec {
...
     }
+
+    template(typename CPO)
+      (requires is_receiver_query_cpo_v<CPO> AND
+                is_callable_v<CPO, const Receiver&>)
+    friend auto tag_invoke(CPO cpo, const type& p)
+        noexcept(is_nothrow_callable_v<CPO, const Receiver&>)
+        -> callable_result_t<CPO, const Receiver&> {
+      return cpo(p.rec_);
+    }
```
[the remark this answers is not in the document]
ericniebler: I don't disagree, but I think the CRTP base thing can be overdone. I find more templates and more (source-level) indirection can make the code hard to read and error messages worse. If it's just one function, I will tend to just write the function rather than go the CRTP route.

include/unifex/task.hpp
```diff
@@ -175,41 +175,52 @@ struct _promise {
...
     }
 
+    template <typename Awaitable>
+    decltype(auto) transform_awaitable_(Awaitable&& awaitable) {
+      if constexpr (blocking_kind::always_inline == cblocking<Awaitable>()) {
+        return (Awaitable&&) awaitable;
+      } else {
+        return unifex::await_transform(
+            *this,
+            typed_via(as_sender((Awaitable&&) awaitable), this->sched_));
```
ispeters, line 205: This obviously works, and, given the time pressure, we should probably go ahead with it, but is there any unnecessary overhead in converting an awaitable to a sender and back again?

include/unifex/task.hpp
```diff
@@ -290,27 +306,56 @@ struct _awaiter {
...
       return thisCoro.promise().result();
     }
 
   private:
+    using scheduler_t = get_scheduler_result_t<OtherPromise&>;
+    using stop_token_t = stop_token_type_t<OtherPromise>;
+    using needs_scheduler_t =
+        std::bool_constant<!same_as<scheduler_t, any_scheduler_ref>>;
+    using needs_stop_token_t =
+        std::bool_constant<!same_as<stop_token_t, inplace_stop_token>>;
```
ispeters, line 341: Should these be `remove_cvref_t`ed?
```suggestion
        std::bool_constant<!same_as<remove_cvref_t<scheduler_t>, any_scheduler_ref>>;
    using needs_stop_token_t =
        std::bool_constant<!same_as<remove_cvref_t<stop_token_t>, inplace_stop_token>>;
```

include/unifex/task.hpp
ericniebler: Agreed. I would like to write a `via` for awaitables, but this works for now.

include/unifex/sender_for.hpp
```diff
@@ -0,0 +1,116 @@
...
+        return sender_for<CPO, remove_cvref_t<Sender>, _context<remove_cvref_t<Queries>...>>{
+            (Sender&&) snd, _context<remove_cvref_t<Queries>...>{(Queries&&) queries...}};
+      }
+    };
+  } // namespace _sf
+
+  using _sf::sender_for;
+
+  template <const auto& CPO>
+  inline constexpr _sf::_make_sender_for<CPO> make_sender_for {};
```
lewissbaker, line 102: I gather that the use of this `make_sender_for` helper is to wrap a sender in something that tags that sender as having been produced by a particular CPO so that you can check whether a given `co_await` expression is syntactically of the form `co_await some_cpo(...)`.

In particular, so that you can do special processing for `co_await schedule(s);` to change the current scheduler.

But this seems to come at the cost of losing some of the type information about the original sender when it is wrapped.
While the wrapper type does forward through some customisations on the sender, it only does so for customisations where the sender is the first argument, which may miss some customisations.

I wonder if there are other ways that we could tackle this requirement?

One would be defining a new tag-like function specifically for changing the current scheduler for the current coroutine.
e.g. `co_await change_current_scheduler_to(sched);`

But I guess this doesn't guard against the possibility of someone writing `co_await schedule(s);` and it just ending up being an expensive no-op which schedules on that other scheduler and then immediately reschedules back on the original context.

You could kind of guard against this somewhat by, say, deleting the `operator co_await()` on each of the schedule senders but that's not really a general solution. You could also define linters that check for the expression `co_await schedule(x);` and warn about it, but not all users will necessarily have access to those linters.

Ultimately, my concerns about the wrapper inhibiting use of customisations may not be warranted, so I'm not sure if we need to make any changes yet, but this approach does give me a little unease.

include/unifex/coroutine_concepts.hpp
```diff
@@ -70,10 +71,13 @@ inline const struct _get_awaiter_fn {
...
   }
-} get_awaiter {};
+};
+} // namespace _get_awaiter
+
+inline constexpr _get_awaiter::_fn get_awaiter {};
 
 template <typename Awaitable>
-using awaiter_type_t = decltype(get_awaiter(std::declval<Awaitable>()));
+using awaiter_type_t = remove_cvref_t<decltype(get_awaiter(std::declval<Awaitable>()))>;
```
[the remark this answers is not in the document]
ericniebler: I don't recall. I'll change it back and see what breaks.

include/unifex/detail/unifex_fwd.hpp
```diff
@@ -28,7 +28,29 @@ namespace unifex {
...
+  {
+    template <typename Key, typename Value>
+    struct _kv {
+      struct type {
+        using key_type = Key;
+        using value_type = Value;
+        UNIFEX_NO_UNIQUE_ADDRESS Key key;
+        Value value;
+      };
+    };
```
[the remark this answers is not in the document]
ericniebler: I never started. I have enough on my plate. Yes, it's painful.

include/unifex/finally.hpp
```diff
@@ -696,36 +716,50 @@ namespace unifex
...
+        }
+        template <typename CompletionSender>
+        constexpr auto operator()(CompletionSender&& completion) const
+            noexcept(is_nothrow_callable_v<
+              tag_t<bind_back>, _fn, CompletionSender>)
+            -> bind_back_result_t<_fn, CompletionSender> {
+          return bind_back(*this, (CompletionSender&&)completion);
+        }
+      };
+    } // namespace _cpo
```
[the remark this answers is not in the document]
ericniebler: > (presumably) isolate it from tag_invoke overloads introduced for the sender/receiver types

Yes, that's the reason. I am now partial to this particular formulation, and will slowly migrate the other CPOs to follow suit as time allows.

> Does it make a big difference?

The change is mostly driven by paranoia about overload sets and ambiguities.

include/unifex/scheduler_concepts.hpp
```diff
@@ -199,6 +205,34 @@ UNIFEX_CONCEPT //
...
+namespace _schedule
+{
+  struct _fn {
+  private:
+    template <typename Scheduler>
+    static auto impl_(Scheduler sched)
+        noexcept(noexcept(make_sender_for<schedule>(
+            _impl{}((Scheduler&&) sched), get_scheduler = Scheduler(sched)))) {
+      return make_sender_for<schedule>(
+          _impl{}((Scheduler&&) sched), get_scheduler = Scheduler(sched));
```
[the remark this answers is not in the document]
ericniebler: So, `get_scheduler` would only be used on scheduler providers like execution contexts, but we would use `get_completion_scheduler` when a task is going to run its continuation on a particular scheduler?

include/unifex/scheduler_concepts.hpp
```diff
@@ -199,6 +205,34 @@ UNIFEX_CONCEPT //
     UNIFEX_FRAGMENT(unifex::_scheduler_provider, SP);
 #endif
 
+namespace _schedule
+{
+  struct _fn {
+  private:
+    template <typename Scheduler>
+    static auto impl_(Scheduler sched)
+        noexcept(noexcept(make_sender_for<schedule>(
```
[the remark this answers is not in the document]
ericniebler: > Does the use of make_sender_for here mean that the return-type of a customised schedule() call is no longer the same as the return-type of the customisation? 

Yes, and it makes me a little uncomfortable.

> If so could this wrapping inhibit subsequent customisations where they may have customised some other algorithm on that specific type?

True, it could.

> An alternative to using this helper to implement the get_scheduler() query would be requiring that all of the scheduler implementations return senders that have get_scheduler() customised.

Lots of senders could have `get_scheduler()` customized for lots of different reasons. We need an unambiguous way to identify senders returned from `schedule` do we can make the appropriate magic happen for `co_await schedule(sched)`.

One thought I've had is to give a sender a way to opt in to "schedule-sender-ness" intrusively instead of always relying on a wrapper. And yes, it would help to eliminate the redundant storage in most schedule senders.

include/unifex/sender_for.hpp
ericniebler: I share your unease, but like you I don't know whether we're just being paranoid or if there is a real problem here. My feeling is, let's try it and see if it's problematic in practice. 🤷‍♂️

include/unifex/scheduler_concepts.hpp
```diff
@@ -199,6 +205,34 @@ UNIFEX_CONCEPT //
...
+            _impl{}((Scheduler&&) sched), get_scheduler = Scheduler(sched)))) {
```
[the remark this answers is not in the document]
ericniebler: Yes, it may be too clever. I am indeed using this as a kind of named parameter, except it's like a little DSL for permitting a caller to say, I would like the resulting type to customize this particular CPO and here is the corresponding value. Suggestions welcome.

include/unifex/scheduler_concepts.hpp
```diff
@@ -100,14 +102,14 @@ namespace _schedule {
...
-        typename decltype(_fn::_select<Scheduler>())::template apply<Scheduler>;
+        typename decltype(_impl::_select<Scheduler>())::template apply<Scheduler>;
   public:
     template(typename Scheduler)
       (requires _with_tag_invoke<Scheduler>)
     constexpr auto operator()(Scheduler&& s) const
         noexcept(is_nothrow_tag_invocable_v<_fn, Scheduler>)
         -> _result_t<Scheduler> {
-      return tag_invoke(_fn{}, static_cast<Scheduler&&>(s));
+      return tag_invoke(schedule, static_cast<Scheduler&&>(s));
```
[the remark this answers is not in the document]
ericniebler: > Where is the `schedule` name defined now?

`<unifex/detail/unifex_fwd.hpp>`

> And why pass `schedule` here instead of, say `_impl{}`?

Because when people provide customizations for the `schedule`, they will be use `tag_t<schedule>`, not `_schedule::_impl`.

include/unifex/scheduler_concepts.hpp
```diff
@@ -84,7 +85,8 @@ namespace _schedule {
   using _as_sender_t = typename _as_sender<remove_cvref_t<Executor>>::type;
 
   struct sender;
-  inline const struct _fn {
+
+  struct _impl {
```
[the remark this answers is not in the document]
ericniebler: I want to standardize on `_fn` to be the idiomatic way we name the types of CPOs. This gets us a bit closer.

include/unifex/connect_awaitable.hpp
```diff
@@ -182,31 +176,17 @@ namespace _await_cpo {
...
-                  constexpr size_t valueCount =
-                      sender_value_types_t<Awaitable, type_identity_t, count_types>::value;
-                  if constexpr (valueCount == 0) {
-                    unifex::set_value(std::move(receiver));
-                  } else if constexpr (valueCount == 1) {
-                    unifex::set_value(std::move(receiver), static_cast<result_type&&>(result));
-                  } else {
-                    std::apply(set_value_applicator<Receiver>{receiver}, (result_type&&)result);
-                  }
+                if constexpr (std::is_void_v<await_result_t<awaitable_type>>) {
```
ispeters, line 179: What do you think of checking for `unifex::same_as<result_type, unit>` instead?

I think the difference would be that an `operator co_await` that directly returns `unifex::unit` would be collapsed to no value instead of passed along.  Dunno if that's a win or if it would increase confusion.

include/unifex/connect_awaitable.hpp
```diff
@@ -182,31 +176,17 @@ namespace _await_cpo {
...
+                  unifex::set_value(std::move(receiver));
                 } else {
-                  // Shouldn't complete with a value if there are no value_types
-                  // specified.
-                  std::terminate();
+                  unifex::set_value(std::move(receiver), static_cast<result_type&&>(result));
```
ispeters, line 182: I'm kind of partial to the idea that we should splat `tuple`-like things into separate arguments here.  Dunno if it can be done generically, though.  I suppose you might be able to query for whether `std::tuple_size<result_type>` works and delegate to `std::apply` if so; the docs suggest it'd work for `std::pair`, `std::tuple`, `std::array`, and any type that specializes `tuple_size` and `get`.

If this idea isn't horrible, it can come later.

include/unifex/task.hpp
```diff
@@ -155,12 +174,85 @@ struct _promise {
...
-        (requires callable<decltype(unifex::await_transform), type&, Value>)
-    auto await_transform(Value&& value)
-        noexcept(is_nothrow_callable_v<decltype(unifex::await_transform), type&, Value>)
-        -> callable_result_t<decltype(unifex::await_transform), type&, Value> {
-      return unifex::await_transform(*this, (Value&&)value);
+    template <typename Value>
+    decltype(auto) await_transform(Value&& value) {
+      if constexpr (derived_from<remove_cvref_t<Value>, _task_base>) {
+        // We are co_await-ing a unifex::task or something that completes inline. We don't
+        // need an additional transition.
```
ispeters, line 181: This comment is out of date.
```suggestion
        // We are co_await-ing a unifex::task. We don't need an additional transition.
```

include/unifex/via.hpp
```diff
@@ -276,6 +276,7 @@ struct _sender<Predecessor, Successor>::type {
     sender_traits<Predecessor>::sends_done ||
     sender_traits<Successor>::sends_done;
 
+  // TODO: give this a compile-time blocking when possible.
```
ispeters, line 279: You've got a TODO here; do you intend to resolve it now or later?

include/unifex/connect_awaitable.hpp
ericniebler: That wouldn't be appropriate in this PR.

include/unifex/connect_awaitable.hpp
ericniebler: I dunno either. I'd rather not make an unrelated change like that in this PR.

PR #292, opened 2021-06-08
make the finally test actually test something

ericniebler:


test/finally_test.cpp
```diff
@@ -27,12 +32,39 @@
...
-  schedule(context.get_scheduler())
-    | finally(schedule(context.get_scheduler()) 
-        | transform([](){ std::printf("finally\n"); }))
-    | transform([](){ std::printf("transform\n"); })
+  auto res = just(42)
+    | finally(schedule(context.get_scheduler()))
+    | transform([](int i){ return std::make_pair(i, std::this_thread::get_id() ); })
     | sync_wait();
+
+  ASSERT_FALSE(!res);
```
ispeters, line 43: I find the double negative less clear than a positive assertion.
```suggestion
  ASSERT_TRUE(res);
```
You can be more explicit, too:
```suggestion
  ASSERT_TRUE(res.has_value());
```

ispeters (approved):
lgtm

PR #306, opened 2021-06-27
CI: update GCC

williamspatrick:


williamspatrick:
Based on another action that was ran, it appears that the 'ubuntu-latest' doesn't immediately have gcc-11 available.  Let me investigate that.  ubuntu-hirsuite has it in apt, so I assumed it would work out of the box on the GH image.  I may need to submit that commit as a separate PR also.

williamspatrick:
Github's "ubuntu-latest" is actually 20.04 right now, which does not have gcc-11.  Going to resubmit without removing gcc-10 and adding gcc-9.

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

PR #308, opened 2021-06-27
get_stop_token: add compile-guard to deduction guide

williamspatrick:


ericniebler:
I would prefer the CI changes in a separate PR.

williamspatrick:
> I would prefer the CI changes in a separate PR.

Sorry, that was my intention.  I must have messed up the branch when I pushed.  Let me update the PR.

williamspatrick:
It looks like based on this workflow run, MacOS is unhappy without the deduction-guide.

https://github.com/facebookexperimental/libunifex/actions/runs/976831250

include/unifex/get_stop_token.hpp
```diff
@@ -43,8 +43,6 @@ namespace _get_stop_token {
           return (StopToken&&) stoken_;
         }
       };
-      template <typename StopToken>
-      _awaiter(StopToken) -> _awaiter<StopToken>;
```
ericniebler, line 47: According to http://eel.is/c++draft/temp.deduct.guide#3, it is perfectly permissible to have a deduction guide for a member template at class scope, so long as it has the same access as the member class. This change _shouldn't_ be needed at all. And you're right, the guide also _shouldn't_ be needed, but Apple clang/gcc needs it. So maybe leave it in and guard it with some Apple macros.

ericniebler (changes requested):

include/unifex/get_stop_token.hpp
williamspatrick: Yes, deduction guide at class scope should work, but gcc has a bug that has been reported for a long while.  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79501 .

I _think_ I was able to make the code work by promoting `_awaiter` itself to the namespace level (outside of `_fn::_awaitable`).  This seems to go against the common pattern in the code-base though.

I'm not sure if the clang requirement is Apple specific or clang-12 specific.  I can do some more testing to find out.  Would you prefer compile guards or class promotion? 

Secondarily, is there any rationale to what is supported in CI or not?  I can add linux-clang-12 in a separate PR but I don't know how many outstanding versions you want to support.

ericniebler: > Would you prefer compile guards or class promotion?

I would prefer a compile guard for the broken compilers, with a comment linking to the relevant bug, so we can find and remove the workaround later.

> Secondarily, is there any rationale to what is supported in CI or not? 

I would welcome a PR to add clang-12, but I would prefer not to drop any older compilers until we have to.

williamspatrick: Added compile guard so GCC doesn't have the code but clang does.

ericniebler:
Thank you!

Issue #307, opened 2021-06-27
gcc-10: failures in copy-constructors due to [[no_unique_address]].


williamspatrick:
I am seeing lots of failures in gcc-10 reporting issues with copy-constructors "... is implicitly deleted because the default definition would be ill-formed".  This seems to be caused by the use of `UNIFEX_NO_UNIQUE_ADDRESS` but I'm stumped as to why.  (i.e. changing `UNIFEX_NO_UNIQUE_ADDRESS` to empty allows everything to compile.)

```
$ cmake -G Ninja -H. -Bbuild -DCMAKE_CXX_FLAGS:STRING=-fcoroutines -DCMAKE_CXX_STANDARD:STRING=20 -DUNIFEX_BUILD_EXAMPLES=yes -DBUILD_TESTING=yes
```

Using commit 797690d386dbf36f169a5c77a29f7e600fb45417.

```
$ gcc --version    
gcc (Gentoo 10.3.0-r1 p2) 10.3.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
FAILED: examples/CMakeFiles/never_stream_cancellation.dir/never_stream_cancellation.cpp.o 
/usr/bin/c++  -I../source/../include -Iinclude -fcoroutines -std=gnu++2a -MD -MT examples/CMakeFiles/never_stream_cancellation.dir/never_stream_cancellation.cpp.o -MF examples/CMakeFiles/never_stream_cancellation.dir/never_stream_cancellation.cpp.o.d -o examples/CMakeFiles/never_stream_cancellation.dir/never_stream_cancellation.cpp.o -c ../examples/never_stream_cancellation.cpp
In file included from ../examples/never_stream_cancellation.cpp:21:
[...]

PR #313, opened 2021-06-28
ci: gcc: add workflows for gcc-10 with coroutines

williamspatrick:


williamspatrick:
I'm half expecting this PR to fail due to #307 .

ericniebler:
This PR could contain two commits: the first would work around the gcc bug by conditionally defining `UNIFEX_NO_UNIQUE_ADDRESS` as empty for gcc-10 (with coroutines?), and the second which adds the CI support.

Would also be good to find a minimal repro and file a compiler bug, and have a link to the bug in a comment.

williamspatrick:
Marked this as draft until PR #314 is merged and GH Actions update gcc-10 to 10.3.  Will then rebase and mark for review.

ericniebler:
> Marked this as draft until PR #314 is merged 

Done.

> and GH Actions update gcc-10 to 10.3. Will then rebase and mark for review.

Why do we need to wait for GH Actions to update gcc-10? After #314, won't it Just Work now?

williamspatrick:
> Why do we need to wait for GH Actions to update gcc-10? After #314, won't it Just Work now?

After #314, everything passes with GCC 10.3.  With GCC 10.2, even before #314, everything compiles but does not pass unit-tests.  There were fixes that went into GCC 10.2 for coroutines (which allows the unit tests pass here).

ericniebler:
> There were fixes that went into GCC 10.2 for coroutines (which allows the unit tests pass here).

Ah, I see. Thanks.

williamspatrick:
> There were fixes that went into GCC 10.2 for coroutines (which allows the unit tests pass here).

I typo'd here.  GCC 10.3 has the fixes for coroutine execution.  GCC 10.2 will fail unit tests.

https://github.com/actions/virtual-environments shows the current deployment rate of Ubuntu images for the GH Action execution environment. `ubuntu20-20210628.1` is the image which upgraded GCC 10.2 -> GCC 10.3.  It is currently at a 22.28% deployment rate.  I will monitor over the next few days and once I see that image reach 100% I'll rebase/retrigger this PR to confirm the `GCC 10.3 w/Coroutines` stages pass and then this can be merged.

ericniebler:
Thanks!

PR #316, opened 2021-07-01
add unifex::create for trivially wrapping a C-style async API, fixes #294

### LINKED ISSUE #294: Add a utility for wrapping a C-style async API into a sender-based API
For example, the following shows how to do it specifically for `unifex::task`:
https://godbolt.org/z/WP81dsPex

Generalize this to typed senders.

```c++
#include <cassert>
#include <iostream>
#include <optional>
#include <functional>
#include <unifex/config.hpp>

#undef UNIFEX_NO_COROUTINES
#define UNIFEX_NO_COROUTINES 0
#define UNIFEX_COROUTINES_HEADER <experimental/coroutine>
#define UNIFEX_COROUTINES_NAMESPACE std::experimental;
#define UNIFEX_ALWAYS_INLINE \
    __attribute__((__always_inline__)) \
    __attribute__((__visibility__("hidden"))) \
    /**/
#define UNIFEX_ASSERT assert
#define UNIFEX_APPLE_CLANG_DISABLE_OPTIMIZATION
#define UNIFEX_NO_INLINE __attribute__((__noinline__))
#define UNIFEX_CLANG_DISABLE_OPTIMIZATION __attribute__((__noopt__))

#include <unifex/task.hpp>
#include <unifex/scheduler_concepts.hpp>
#include <unifex/single_thread_context.hpp>
#include <unifex/async_scope.hpp>
#include <unifex/sync_wait.hpp>
#include <unifex/typed_via.hpp>

namespace coro = unifex::coro;

// Library code starts here:
struct AsyncInvokerBase;

struct PromiseBase {
    coro::coroutine_handle<void> parent_{};
    std::exception_ptr eptr_{};

    void setError(std::exception_ptr eptr) {
        eptr_ = std::move(eptr);
        parent_.resume();
    }
};

template <typename Result>
struct Promise : PromiseBase {
private:
    friend AsyncInvokerBase;
    std::optional<Result> result_{};
public:
    void setValue(Result res) {
        result_.emplace(std::move(res));
        parent_.resume();
    }
    Result getResult() && {
        if (eptr_)
            std::rethrow_exception(eptr_);
        return std::move(*result_);
    }
};
template <typename Result>
using RampFunction = std::function<void(Promise<Result>&)>;

struct AsyncInvokerBase {
    static void setParent(PromiseBase& promise, coro::coroutine_handle<void> parent) noexcept {
        promise.parent_ = parent;
    }
};

template <typename Result>
struct AsyncInvoker : private AsyncInvokerBase {
private:
    struct [[nodiscard]] Awaiter {
    private:
        Promise<Result> promise_{};
        RampFunction<Result> fn_;
    public:
        explicit Awaiter(RampFunction<Result> fn)
          : fn_(std::move(fn))
        {}
        static bool await_ready() noexcept { return false; }
        void await_suspend(coro::coroutine_handle<void> cont) noexcept {
            setParent(promise_, cont);
            fn_(promise_);
        }
        Result await_resume() {
            return std::move(promise_).getResult();
        }
    };
public:
    Awaiter operator()(RampFunction<Result> fn) const noexcept {
        return Awaiter{std::move(fn)};
    }
};

template <typename Result>
inline constexpr AsyncInvoker<Result> asyncInvoke{};

/// Library code ends here.


/// User code starts here:

unifex::single_thread_context someThread;
unifex::async_scope someScope;

using CAsyncCodeCompleted = void (*)(void* context, int result);
void CAsyncCode(int a, int b, void* context, CAsyncCodeCompleted completed) {
    // Execute some work asynchronously on some other thread. When its
    // work is finished, pass the result to the callback.
    someScope.spawn_call_on(someThread.get_scheduler(), [=]() noexcept {
        auto result = a + b;
        completed(context, result);
    });
}

unifex::task<int> CAsyncCodeAdapter(int a, int b) {
    // Eventually this asyncInvoke will be able to inherit the current scheduler
    // from the parent coroutine and automatically thunk back onto that scheduler
    // after executing the callback.
    int result = co_await asyncInvoke<int>([=](Promise<int>& promise) {
        CAsyncCode(a, b, &promise, [](void* context, int result) {
            static_cast<Promise<int>*>(context)->setValue(result);
        });
    });

    co_return result;
}

unifex::single_thread_context rsysThread;

auto runOnRsys() {
    return unifex::schedule(rsysThread.get_scheduler());
}

unifex::task<int> run() {
    co_await runOnRsys();
    co_return co_await unifex::typed_via(
        CAsyncCodeAdapter(1, 2),
        rsysThread.get_scheduler());
}

int main() {
    std::optional<int> result = unifex::sync_wait(run());
    std::printf("The result of 1 + 2 is %d\n", *result);

    // Clean up the scope of someThread.
    unifex::sync_wait(someScope.cleanup());
}

#include <../source/async_manual_reset_event.cpp>
```

ericniebler:
###  `create<ValueTypes...>(callable)`

_Synopsis:_ A utility for building a (lazy) sender-based async API out of an eager C-style
async API that accepts a `void*` context and a callback.

_Example:_
```c++
// A void-returning C-style async API that accepts a context and a continuation:
using callback_t = void(void* /*context*/, int /*result*/);
void old_c_style_api(int a, int b, void* context, callback_t* callback_fn);

// A sender-based async API implemented in terms of the C-style API (using C++20):
unifex::typed_sender auto new_sender_api(int a, int b) {
  return unifex::create<int>([=](auto& rec) {
    static_assert(unifex::receiver_of<decltype(rec), int>);
    old_c_style_api(a, b, &rec, [](void* context, int result) {
      unifex::void_cast<decltype(rec)>(context).set_value(result);
    });
  });
}
```

`ValueTypes...` is a pack representing the value types of the resulting sender. It should
be the list of value type(s) accepted by the callback (with the exception of the `void*`
context). In the above example, since `callback_t` accepts an `int` as the result of the
async computation, we pass `int` as the template argument to `create`.

The first argument to `create` is a `void`-returning callable that accepts an lvalue
reference to an object whose type satisfies the `unifex::receiver_of<ValueTypes...>`
concept. This function should dispatch to the C-style callback (see example).

The second argument is an optional extra bit of data to be bundled with the receiver
passed to the callable. E.g., if the first argument to `create` is a lambda that accepts
`(auto& rec)` and the second argument is `42`, then from within the body of the lambda,
the value of the expression `rec.context()` is `42`.

`create` returns a typed sender that, when connected and started, dispatches to the
wrapped C-style API with the callback of your choosing. The receiver passed to the
callable wraps the receiver passed to `connect`. The callback should "complete" the
receiver passed to the callable, which will complete the receiver passed to `connect` in
turn.

### DIFF
--- a/include/unifex/create.hpp
+++ b/include/unifex/create.hpp
@@ -1,0 +1,197 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <unifex/receiver_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
+#include <unifex/type_traits.hpp>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+namespace _create {
+
+template <typename Receiver, typename Fn, typename Context>
+struct _op {
+  struct type {
+    explicit type(Receiver rec, Fn fn, Context ctx)
+        noexcept(std::is_nothrow_move_constructible_v<Receiver> &&
+          std::is_nothrow_move_constructible_v<Fn> &&
+          std::is_nothrow_move_constructible_v<Context>)
+      : rec_((Receiver&&) rec), fn_((Fn&&) fn), ctx_((Context&&) ctx) {}
+
+    template (typename... Ts)
+      (requires receiver_of<Receiver, Ts...>)
+    void set_value(Ts&&... ts) noexcept(is_nothrow_receiver_of_v<Receiver, Ts...>) {
+      unifex::set_value((Receiver&&) rec_, (Ts&&) ts...);
+    }
+
+    template (typename Error)
+      (requires receiver<Receiver, Error>)
+    void set_error(Error&& error) noexcept {
+      unifex::set_error((Receiver&&) rec_, (Error&&) error);
+    }
+
+    void set_done() noexcept {
+      unifex::set_done((Receiver&&) rec_);
+    }
+
+    template (class Ctx = Context)
+      (requires (!same_as<Ctx, detail::_empty<>>))
+    Context const& context() const & noexcept { return ctx_; }
+
+    template (class Ctx = Context)
+      (requires (!same_as<Ctx, detail::_empty<>>))
+    Context&& context() && noexcept { return (Context&&) ctx_; }
+
+  private:
+    friend void tag_invoke(tag_t<start>, type& self) noexcept try {
+      self.fn_(self);
+    } catch(...) {
+      unifex::set_error((Receiver&&) self.rec_, std::current_exception());
+    }
+
+    // Forward other receiver queries
+    template (typename CPO)
+      (requires is_receiver_query_cpo_v<CPO> AND is_callable_v<CPO, const Receiver&>)
+    friend auto tag_invoke(CPO cpo, const type& self)
+        noexcept(is_nothrow_callable_v<CPO, const Receiver&>)
+      -> callable_result_t<CPO, const Receiver&> {
+      return std::move(cpo)(self.rec_);
+    }
+
+    UNIFEX_NO_UNIQUE_ADDRESS Receiver rec_;
+    UNIFEX_NO_UNIQUE_ADDRESS Fn fn_;
+    UNIFEX_NO_UNIQUE_ADDRESS Context ctx_;
+  };
+};
+
+template <typename Receiver, typename Fn, typename Context>
+using _operation = typename _op<Receiver, Fn, Context>::type;
+
+template <typename Fn, typename Context>
+struct _snd_base {
+  struct type {
+    template <template<typename...> class Variant>
+    using error_types = Variant<std::exception_ptr>;
+
+    static constexpr bool sends_done = true;
+
+    template (typename Self, typename Receiver)
+      (requires derived_from<remove_cvref_t<Self>, type> AND
+        constructible_from<Fn, member_t<Self, Fn>> AND
+        constructible_from<Context, member_t<Self, Context>>)
+    friend _operation<remove_cvref_t<Receiver>, Fn, Context>
+    tag_invoke(tag_t<connect>, Self&& self, Receiver&& rec)
+        noexcept(std::is_nothrow_constructible_v<
+          _operation<Receiver, Fn, Context>,
+          Receiver,
+          member_t<Self, Fn>,
+          member_t<Self, Context>>) {
+      return _operation<remove_cvref_t<Receiver>, Fn, Context>{
+        (Receiver&&) rec,
+        ((Self&&) self).fn_,
+        ((Self&&) self).ctx_};
+    }
+
+    UNIFEX_NO_UNIQUE_ADDRESS Fn fn_;
+    UNIFEX_NO_UNIQUE_ADDRESS Context ctx_{};
+  };
+};
+
+template <typename Fn, typename Context, typename... ValueTypes>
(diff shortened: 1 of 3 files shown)


include/unifex/create.hpp
```diff
@@ -0,0 +1,196 @@
...
+        noexcept(std::is_nothrow_constructible_v<
+          _operation<Receiver, Fn, Context>,
+          member_t<Self, Fn>,
+          member_t<Self, Context>,
+          Receiver>) {
+      return _operation<remove_cvref_t<Receiver>, Fn, Context>{
+        (Receiver&&) rec,
+        (member_t<Self, Fn>&&) self.fn_,
+        (member_t<Self, Context>&&) self.ctx_};
+    }
```
ispeters, line 107: Can you explain what the `member_t` alias is doing here?  I've been staring at the definition and trying to cross-reference it to these uses and I can't figure it out.

include/unifex/create.hpp
```diff
@@ -0,0 +1,196 @@
...
+template <typename... ValueTypes>
+struct _fn {
+  template (typename Fn)
+    (requires move_constructible<Fn>)
+  _sender<Fn, ValueTypes...> operator()(Fn fn) const
+      noexcept(std::is_nothrow_constructible_v<_sender<Fn, ValueTypes...>, Fn>) {
+    return _sender<Fn, ValueTypes...>{(Fn&&) fn};
+  }
+  template (typename Fn, typename Context)
+    (requires move_constructible<Fn> AND move_constructible<Fn>)
```
ispeters, line 143: You've duplicated `move_constructible<Fn>`; is the second one supposed to be `move_constructible<Context>`?
```suggestion
    (requires move_constructible<Fn> AND move_constructible<Context>)
```

test/create_test.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+TEST_F(CreateTest, VoidEWithContextTest) {
+  bool called = false;
+  auto snd = [&called, this]() {
+    return create<>([&called, this](auto& rec) {
+      static_assert(receiver_of<decltype(rec)>);
+      aVoidAPI(&rec, [](void* context) {
+        auto& rec2 = unifex::void_cast<decltype(rec)>(context);
+        rec2.set_value();
+        rec2.context().get() = true;
```
ispeters, line 81: The bug is here somewhere.  In some compiler/build mode combinations, this fails to write to `called` and the test fails on the assertion on line 89, below; in others, this is a write to a null pointer and you crash.

I wonder if the problem is that you're writing to the operation state after completing the receiver contract.
```suggestion
        rec2.context().get() = true;
        rec2.set_value();
```

ispeters (changes requested):
Looks very nice; I'd like to know whether my suggestion on the test case fixes the crashing bug.

include/unifex/create.hpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    // Forward other receiver queries
+    template (typename CPO)
+      (requires is_receiver_query_cpo_v<CPO> AND is_callable_v<CPO, const Receiver&>)
+    friend auto tag_invoke(CPO cpo, const type& self)
+        noexcept(is_nothrow_callable_v<CPO, const Receiver&>)
+      -> callable_result_t<CPO, const Receiver&> {
+      return std::move(cpo)(self.rec_);
+    }
```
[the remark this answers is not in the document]
ericniebler: It should. If it doesn't, I'm interested in seeing the problem.

include/unifex/create.hpp
ericniebler: If `c` is an object of class type such that `decltype(c)` is `C`, _and_ the class has a data member `m` of type `M`, then `member_t<C, M>` is `decltype(c.m)`. If `Self` is not a reference type above, then it would be wrong to pass `self.fn_` to the `_operation` constructor because that would be an lvalue reference, inhibiting the move. So we cast to `member_t<Self, Fn>&&` which is an rvalue if `Self` is an rvalue reference or a non-reference type, and and lvalue if `Self` is an lvalue reference type.

Moves members when possible, copies when you need them.

include/unifex/create.hpp
ericniebler: Nice spot.

test/create_test.cpp
ericniebler: THANK YOU

include/unifex/create.hpp
ispeters: Nice.

Is that different from
```
static_cast<Self&&>(self).fn_
```
in this case?

ericniebler: Ah, no it's not. That would probably be simpler.

Issue #88, opened 2020-03-12
Generalise range_stream to adapt arbitrary ranges to async-streams


lewissbaker:
The current implementation of `range_stream` produces a stream of integers in some specified range.

This was largely built just so we had some kind of stream that we could test against, but it would be much more useful to change this to take an arbitrary range (view/container) and allow construction of a stream that produces the elements of that range.

Existing uses of the `range_stream` could be changed to `range_stream(std::views::iota(min, max))`.
Or we could rename the existing stream `iota_stream` and then add a more general `range_stream`.

PR #320, opened 2021-07-18
Generalize range stream

ManuelMeraz:
I'm very interested in using libunifex and one of the first things I need support for is supporting ranges beyond the simple range_stream in examples. I noticed [this](https://github.com/facebookexperimental/libunifex/issues/88#issue-580245635) issue was open and decided this was worth tackling. 

So far I modified `range_stream.hpp` to take in an rvalue and lvalue range, and then we can grab the iterators from that range. It got tricky because I couldn't figure out how to support both taking in a single range as an argument as well as begin and end iterators using the same `stream`  struct, but maybe this isn't necessary. 

I'd like to improve this pull request as well because I have definitely overlooked something and would appreciate a review plus feedback on how I can improve this PR. 

Changes made:
* Updated examples to use `range_stream{std::views::iota(0, 10)}` instead of `range_stream{0, 10}` 
* Added a test using a few simple examples beyond integers
* Updated the `range_stream.hpp` header file to support streams and constrained the type using `std::ranges::range`

### DIFF
--- a/include/unifex/range_stream.hpp
+++ b/include/unifex/range_stream.hpp
@@ -85,7 +85,7 @@ struct stream {
   sentinel_t end_;
 
   explicit stream(Range&& r) : begin_(std::begin(r)), end_(std::end(r)) {}
-  explicit stream(iterator_t begin, sentinel_t end) : begin_(begin), end_(end) {}
+  explicit stream(Range& r) : begin_(std::begin(r)), end_(std::end(r)) {}
 
   friend next_sender<Range> tag_invoke(tag_t<next>, stream<Range>& s) noexcept {
     return next_sender{s};
--- a/include/unifex/range_stream.hpp
+++ b/include/unifex/range_stream.hpp
@@ -22,30 +22,35 @@
 
 #include <type_traits>
 #include <utility>
+#include <ranges>
 
 #include <unifex/detail/prologue.hpp>
 
 namespace unifex {
 namespace _range {
+
+template <std::ranges::range Range>
 struct stream;
 
-template <typename Receiver>
+template <typename Receiver, std::ranges::range Range>
 struct _op {
   struct type;
 };
-template <typename Receiver>
-using operation = typename _op<remove_cvref_t<Receiver>>::type;
+template <typename Receiver, std::ranges::range Range>
+using operation = typename _op<remove_cvref_t<Receiver>, Range>::type;
 
-template <typename Receiver>
-struct _op<Receiver>::type {
-  stream& stream_;
+template <typename Receiver, std::ranges::range Range>
+struct _op<Receiver, Range>::type {
+
+  stream<Range>& stream_;
   Receiver receiver_;
 
   void start() noexcept;
 };
 
+template <std::ranges::range Range>
 struct next_sender {
-  stream& stream_;
+  stream<Range>& stream_;
 
   template <
       template <typename...> class Variant,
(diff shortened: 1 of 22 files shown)


include/unifex/range_stream.hpp
```diff
@@ -59,26 +64,30 @@ struct next_sender {
...
+  using iterator_t = std::ranges::iterator_t<Range>;
+  using sentinel_t = std::ranges::sentinel_t<Range>;
+
+  iterator_t begin_;
+  sentinel_t end_;
 
-  explicit stream(int max) : next_(0), max_(max) {}
-  explicit stream(int start, int max) : next_(start), max_(max) {}
+  explicit stream(Range&& r) : begin_(std::begin(r)), end_(std::end(r)) {}
+  explicit stream(Range& r) : begin_(std::begin(r)), end_(std::end(r)) {}
```
ManuelMeraz, line 88: I'll need some help here. I thought I understood templated universal references, but clearly not. From my understanding some function
```
template <typename T>
void foo(T&& t) {}
```
Would be in a state where it would support both an lvalue reference and an rvalue reference. When that function is called `t` would either be type `T&` or `T&&`, but we do not know. So we have to use `std::forward<T>` to do a conditional rvalue cast. 

With this logic I assumed that the `stream(Range&&)` constructor would be enough to handle both lvalue and rvalue types, but for some reason it didn't and I needed ot make the `stream(Range&)` constructor.

ericniebler: I can explain that. In this code:

```c++
template <typename T>
void foo(T&& t) {}
```

... the type `T` is deduced from the type of the argument. And you're correct, it could be an lvalue or an rvalue.

In this code, however:

```c++
template <typename T>
struct S {
  void foo(T&& t) {}
};
```

... the type `T` is _not_ deduced. It is specified when instantiating the class template as either an lvalue reference type or as an ordinary value type. If you specify it to be a value type -- say, `int` -- then the `T&&` in the argument list of `foo()` becomes `int&&`. That will bind _only_ to rvalue `int`s. It will _not_ bind to lvalues.

Hope this helps.

ericniebler:
Apologies about letting this PR languish for so long. It's a nice change. It comes with complications, though. This `range_stream` can only work with a C++20 compiler that supports concepts and a stdlib that implements ranges (also C++20). Libunifex only requires C++17 today. I'm not sure what the best thing to do is.

One thing I would like to see is that we leave the original `range_stream` and rename it to something like `irange_stream`. Then we could support your `range_stream` conditionally if the stdlib supports ranges.

That's really just a half-measure, though, because folks will want to pass, say, a `std::vector` to it from C++17 code. That _should_ work, but would require a fair bit of leg work.

lewissbaker:
> Apologies about letting this PR languish for so long. It's a nice change. It comes with complications, though. This `range_stream` can only work with a C++20 compiler that supports concepts and a stdlib that implements ranges (also C++20). Libunifex only requires C++17 today. I'm not sure what the best thing to do is.
> 
> One thing I would like to see is that we leave the original `range_stream` and rename it to something like `irange_stream`. Then we could support your `range_stream` conditionally if the stdlib supports ranges.
> 
> That's really just a half-measure, though, because folks will want to pass, say, a `std::vector` to it from C++17 code. That _should_ work, but would require a fair bit of leg work.

As far as I can see there isn't much in there that depends on the ranges library.
We could easily define local replacements for the `ranges::iterator_t` and `ranges::sentinel_t` metafunctions.
And the other main usages of ranges are just the constraints on the `range_stream` template parameters.
We could just not constrain these template parameters for now.

The test usage of `std::views::iota` could be replaced with a local utility range defined in the test/ directory.

include/unifex/range_stream.hpp
ManuelMeraz: Wow I really appreciate that explanation. So the gist is that if the template type parameters are deduced, then the _referencesness_ has different semantics than when they are not deduced. I'm going to have to make sure I retain this.

ManuelMeraz:
It sounds like a dependency on `std::ranges` is not ideal right now. So the two possible paths are:

1. Keep depending on `std::ranges` but accommodate depending if the feature is available
2. Implement a minimally effective `std::ranges` equivalent 

The first path will require a lot more cmake code to check for compile features and enabling them via compile options, which might be the way to go long term. 

The second path requires less work and we can have this feature out without a dependency on `std::ranges`

How about we break this task in to two tasks? 

1. Implement minimally viable `range_stream` with a `range_iota` that is basically the old `range_stream` for tests, examples, etc.
2. `rang_stream` is unconstrained as @lewissbaker suggested, or maybe we can at least add a contraint for `begin` and `sentinel`? Not sure if this is what `std::ranges::range` already does.
3. We can grab the `begin` and `sentinel` types via some in house compile time functions, with the other in house compile time functions

The second task is a bit larger and can be focused on integrating `std::ranges` if the compiler supports it, this could be where we define our own constraint and conditionally check with `std::ranges`if it's enabled, or use our own if it's disabled? There's probably some thinking to do here on how exactly to do this.

lewissbaker:
For the minimal bits needed to emulate ranges pieces needed here, much of it can be copied from the ranges emulation part near the top of https://github.com/lewissbaker/generator/blob/main/include/__generator.hpp with some modifications for concepts bits to use either concepts emulation or just plain SFINAE techniques so they work in c++17.

AnujYamdagni:
@ManuelMeraz, We're doing repo cleanup. This seems valuable but it seems like there's still more work to be done before we can merge this. Are you planning on making this compatible with C++17 or shall we just close this?

ManuelMeraz:
> @ManuelMeraz, We're doing repo cleanup. This seems valuable but it seems like there's still more work to be done before we can merge this. Are you planning on making this compatible with C++17 or shall we just close this?

Go ahead and close this. I don't have much time for personal projects, so I ended up just forking the repo and using it in my own project and working off C++ 20.

PR #321, opened 2021-07-20
Fix warnings reported by strict GCC settings

williamspatrick:
When the repository is used as a meson subproject, where the consuming top-level
project uses `warning_level=3`[1], libunifex fails to compile with many warnings
reported by GCC.  These errors are all trivially resolved so fix them and add
CMake set up to prevent future regressions.

1. https://mesonbuild.com/Builtin-options.html

### DIFF
--- a/cmake/unifex_env.cmake
+++ b/cmake/unifex_env.cmake
@@ -18,3 +18,11 @@ elseif("x${CMAKE_CXX_COMPILER_ID}" STREQUAL "xMSVC")
 else()
   message(WARNING "[unifex warning]: unknown compiler ${CMAKE_CXX_COMPILER_ID} !")
 endif()
+
+if (MSVC)
+    # warning level 3 and all warnings as errors
+    add_compile_options(/W3 /WX)
+else()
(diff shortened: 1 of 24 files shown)


williamspatrick:
I think this is finally ready for reviews.  Passes all CI scenarios.

cmake/unifex_env.cmake
```diff
@@ -18,3 +18,11 @@ elseif("x${CMAKE_CXX_COMPILER_ID}" STREQUAL "xMSVC")
 else()
   message(WARNING "[unifex warning]: unknown compiler ${CMAKE_CXX_COMPILER_ID} !")
 endif()
+
+if (MSVC)
+    # warning level 3 and all warnings as errors
```
williamspatrick, line 23: i gave up on enabling L4 for MSVC.  There were things it reports at this level that are just not feasible without sprinkling a lot of `#pragma`'s around and it probably isn't worth the effort.

ericniebler (approved):
Thanks! This is much appreciated.

PR #323, opened 2021-07-28
Change license to Apache 2.0 with LLVM Exceptions

LeeHowes:
Slight relaxation of the license for compatibility with LLVM and the standard library.


ispeters (approved):
The changes I see look good.

How'd you ensure the changes are exhaustive?

LeeHowes:
grep -r "Apache" * | grep -v LLVM

Although that shows Apache referenced in the Cmake files. I'll remove that because it is confusing.

PR #324, opened 2021-07-29
Remove the executor concept

ericniebler:
* Remove the `executor` concept
* Remove the mutually recursive definitions of `execute` and `connect`
* Change the `execute` algorithm to take a scheduler instead of a sender


include/unifex/detail/unifex_fwd.hpp
```diff
@@ -37,20 +28,20 @@ namespace unifex {
...
   namespace _submit_cpo {
     extern const struct _fn submit;
   } // namespace _submit_cpo
   using _submit_cpo::submit;
 
-  namespace _connect_cpo {
-    extern const struct _fn connect;
-  } // namespace _connect_cpo
-  using _connect_cpo::connect;
+  namespace _connect::_cpo {
```
LeeHowes, line 41: Why make this change in this PR but not a similar one for submit?

include/unifex/at_coroutine_exit.hpp
```diff
@@ -20,6 +20,8 @@
 #include <unifex/await_transform.hpp>
 #include <unifex/continuations.hpp>
 #include <unifex/stop_token_concepts.hpp>
+#include <unifex/unstoppable_token.hpp>
+#include <unifex/get_stop_token.hpp>
```
LeeHowes, line 24: Were these necessary for this change or simply missing?

ericniebler: Without these, the msvc build fails.

include/unifex/detail/unifex_fwd.hpp
ericniebler: Because I was touching connect but not submit, no other reason. I've been making this change in dribs and drabs as I maintain code. I can do submit too if you like.

LeeHowes:
With your last change it might be worth adding P2300 as a reference in the readme as well along with P0443.

LeeHowes (approved):

ericniebler:
> With your last change it might be worth adding P2300 as a reference in the readme as well along with P0443.

Thanks. My preference is to replace the reference, which I have now done.

PR #326, opened 2021-08-11
make unifex::when_all a tag_invoke customization point

ericniebler:


include/unifex/when_all.hpp
```diff
@@ -360,19 +360,26 @@ class _sender<Senders...>::type {
...
 };
-} // namespace _when_all
 
-namespace _when_all_cpo {
-  inline const struct _fn {
-    template <typename... Senders>
+namespace _cpo {
+  struct _fn {
+    template (typename... Senders)
+      (requires (unifex::sender<Senders> &&...) AND tag_invocable<_fn, Senders...>)
```
ispeters, line 367: Why does this require `sender<Senders>` but the default implementation requires `typed_sender<Senders>`?

ericniebler: I don't rule out the possibility that some custom implementation of `when_all` might only require `sender`s and not `typed_sender`s.

ispeters: The near-similarity of similar things makes it look like a bug that they're different.

The rationale you've given here makes me wonder why not leave the concept checking to the applicable `tag_invoke` overload and leave `requires sender<Senders> &&...` off altogether?

PR #327, opened 2021-08-11
Rename all the transform* and let* algorithms in accordance with P2300

ericniebler:


doc/api_reference.md
```diff
@@ -12,14 +12,16 @@
   * `just_error()`
   * `stop_if_requested()`
 * Sender Algorithms
-  * `transform()`
-  * `transform_done()`
+  * `then()`
```
LeeHowes, line 15: This PR makes me less comfortable with our naming. I'd really prefer consistency rather than have "then" stand out like this.

LeeHowes (approved):
This is big so it's hard to validate the details, but it looks right given that it's a rename.

doc/api_reference.md
ericniebler: So you'd prefer it to be called `upon_value`? I'm not wild about that either.

ericniebler:
Given that this PR is requiring me to add deprecated forwarding headers for back-compat, I think the time is right for a physical restructuring of the files in the repo, as we've discussed previously. I'll make it part of this PR, and try to get all the breaking changes in at one go.

doc/api_reference.md
LeeHowes: I would prefer that so the names are at least consistent. It's not the cleanest option.

PR #329, opened 2021-08-19
Hoist untyped RAII containers for coroutine_handle<> out of task<> and its awaiter

ispeters:
This PR hoists two RAII containers for `coroutine_handle<>` out of `unifex::task<>` and its awaiter.  The result is a fairly significant binary size reduction in one of our apps.

### DIFF
--- a/include/unifex/task.hpp
+++ b/include/unifex/task.hpp
@@ -184,8 +184,8 @@ struct _promise {
 
     auto final_suspend() noexcept {
       struct awaiter : _final_suspend_awaiter_base {
-        auto await_suspend(coro::coroutine_handle<type> h) noexcept {
-          return h.promise().continuation_.handle();
+        void await_suspend(coro::coroutine_handle<type> h) noexcept {
+          return h.promise().continuation_.handle().resume();
         }
       };
       return awaiter{};
--- a/include/unifex/task.hpp
+++ b/include/unifex/task.hpp
@@ -291,6 +291,7 @@ struct tagged_coro_holder {
   }
 
  protected:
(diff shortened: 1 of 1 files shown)


ispeters:
Hmm… looks like I've broken the tests with C++20; I'll have to look at that tomorrow.

include/unifex/task.hpp
```diff
@@ -386,7 +414,7 @@ struct _task<T>::type : _task_base {
   using awaiter = typename _awaiter<promise_type, OtherPromise>::type;
 
   explicit type(coro::coroutine_handle<promise_type> h) noexcept
-    : coro_(h) {}
+    : coro_holder(h) {}
 
   template<typename Promise>
   friend awaiter<Promise> tag_invoke(tag_t<unifex::await_transform>, Promise&, type&& t) noexcept {
```
lewissbaker, line 420: Doesn't the `await_transform` method need to be updated to pull the coroutine handle out of the new `coro_holder`?
e.g. it's still referencing `t.coro_` below but that data-member has been removed.

Also the awaiter it's constructing is expecting a typed coroutine handle but now you only have a `coroutine_handle<void>` coroutine handle strored in `coro_holder`.

lewissbaker: Ah, I see - the data-member from the base class is protected so you're still able to access it here.
Never mind.

ispeters: Ah; that was absolutely not intentional.  I'll take this change.

ispeters:
Hmm. So MSVC doesn't like the symmetric transfer in `task<>`'s final awaiter.

ericniebler (approved):
LGTM

PR #333, opened 2021-08-23
adding upon_* async algorithms

RishabhRD:
[P2300](https://wg21.link/P2300) proposes two algorithms similar to ``unifex::then``, ``upon_done`` and ``upon_error``. This PR intends to add those algorithms to libunifex with names: ``unifex::upon_done`` and ``unifex::upon_error``. I am trying to make coding conventions similar to already existing modules in libunifex and would be happy to change if some coding conventions doesn't match coding conventions of project.

Current Progress:
- upon_done implementation and test added and tests are passing. Tests are kept similar to unifex::then tests.
- upon_error implementation and tests added and tests are passing.

### DIFF
--- a/include/unifex/upon_done.hpp
+++ b/include/unifex/upon_done.hpp
@@ -112,11 +112,33 @@ struct _sender<Predecessor, Func>::type {
 
 private:
   /*
-   * This helper returns type_list<Result> if func returns Result
-   * else if func returns void then helper returns type_list<>
+   * This helper returns type_list<type_list<Result>> if func returns Result
+   * else if func returns void then helper returns type_list<type_list<>>
    */
   using invoked_result_t =
-      detail::result_overload_t<std::invoke_result_t<Func>>;
+      type_list<detail::result_overload_t<std::invoke_result_t<Func>>>;
+
+  template <
+      template <typename...>
+      class Variant,
+      template <typename...>
+      class Tuple>
+  using sets_done_predecessor_value_type_t = type_list_nested_apply_t<
+      concat_type_lists_unique_t<
+          sender_value_types_t<Predecessor, type_list, type_list>,
+          invoked_result_t>,
+      Variant,
+      Tuple>;
+
+  template <
+      template <typename...>
+      class Variant,
+      template <typename...>
+      class Tuple>
+  using no_sets_done_predecessor_value_type_t = type_list_nested_apply_t<
+      sender_value_types_t<Predecessor, type_list, type_list>,
+      Variant,
+      Tuple>;
 
 public:
   template <
@@ -124,15 +146,10 @@ public:
       class Variant,
       template <typename...>
       class Tuple>
-  using value_types = type_list_nested_apply_t<
-      type_list<concat_type_lists_unique_t<
-          sender_value_types_t<
-              Predecessor,
-              concat_type_lists_unique_t,
-              type_list>,
-          invoked_result_t>>,
-      Variant,
-      Tuple>;
+  using value_types = std::conditional_t<
+      Predecessor::sends_done,
+      sets_done_predecessor_value_type_t<Variant, Tuple>,
+      no_sets_done_predecessor_value_type_t<Variant, Tuple>>;
 
   template <template <typename...> class Variant>
   using error_types = typename concat_type_lists_unique_t<
--- a/include/unifex/upon_done.hpp
+++ b/include/unifex/upon_done.hpp
@@ -1,0 +1,169 @@
+#pragma once
+
+#include <unifex/config.hpp>
+#include <unifex/async_trace.hpp>
+#include <unifex/bind_back.hpp>
+#include <unifex/receiver_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
+#include <unifex/std_concepts.hpp>
+#include <unifex/tag_invoke.hpp>
+#include <unifex/type_list.hpp>
+#include <unifex/type_traits.hpp>
+
+#include <functional>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+namespace _upon_done {
+
+template <typename Receiver, std::invocable Func>
+struct _receiver {
+  struct type;
+};
+template <typename Receiver, std::invocable Func>
+using receiver_t = typename _receiver<Receiver, Func>::type;
+
+template <typename Receiver, std::invocable Func>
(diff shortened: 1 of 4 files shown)


include/unifex/upon_done.hpp
```diff
@@ -0,0 +1,169 @@
...
+#include <unifex/type_traits.hpp>
+
+#include <functional>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+namespace _upon_done {
+
+template <typename Receiver, std::invocable Func>
```
ericniebler, line 20: Libunifex targets C++17 compilers that don't support concepts. And generally we don't constrain implementation details in the interest of compile times. So check parameters at interface boundaries, and constrain overloads like `connect`, `set_*`, and `start`, and otherwise use unconstrained templates.

include/unifex/upon_done.hpp
```diff
@@ -0,0 +1,169 @@
...
+using receiver_t = typename _receiver<Receiver, Func>::type;
+
+template <typename Receiver, std::invocable Func>
+struct _receiver<Receiver, Func>::type {
+  UNIFEX_NO_UNIQUE_ADDRESS Func func_;
+  UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+
+  template <typename... Values>
+  void set_value(Values&&... values) && noexcept {
+    unifex::set_value(std::move(receiver_), std::forward<Values>(values)...);
```
ericniebler, line 34: Rather than using `std::move` and `std::forward` (which instantiate templates), we generally use `static_cast` directly to save compile time. And to be honest, we tend to use C-style casts for that (but don't tell Bjarne ;-).

include/unifex/upon_done.hpp
```diff
@@ -0,0 +1,169 @@
...
+        unifex::set_value(std::move(receiver_));
+      }
+      UNIFEX_CATCH(...) {
+        unifex::set_error(std::move(receiver_), std::current_exception());
+      }
+    }
+  }
+
+  template <typename CPO, typename R>
+  requires is_receiver_query_cpo_v<CPO> AND same_as<R, type>
```
ericniebler, line 58: We use a `template` macro (I know) to make it possible to use `requires` in C++17 code. It maps to `enable_if` under the covers.

```suggestion
  template (typename CPO, typename R)
    (requires is_receiver_query_cpo_v<CPO> AND same_as<R, type>)
```

And similar changes elsewhere.

ericniebler (changes requested):
Awesome, thanks! This looks very good overall. A few comments...

RishabhRD:
Thanks @ericniebler for comments. I tried to resolve most of your comments... Please review the new changes. If these changes match the project convention, I would try to implement upon_error with same conventions.

RishabhRD:
Hello @ericniebler I have few questions regarding the behavior of upon_error.

As receiver concept suggests, that Error can never be void. According to proposal, the upon_error callback expects the error sent by sender as parameter. So, I have doubts:
- Is the callback allowed to return some values?
- If callback is not allowed to return the values, then it seems intuitive to set the original error.
- However, if callback is allowed to return the values, then there are 2 cases:
   - If a value is actually returned, then it seems intuitive to set that value as error.
   - However, if a no value is returned, then what should we set as error? One option is we can constrain the callback to always return something other than void.

RishabhRD:
Hello @ericniebler I have added the upon_error interface implementation and tests. We would figure out more changes if required to the interface based on your answers to my doubts.

Currently the interface assumes following:

If ``upon_error(Sender, Func)`` is the function, where Func is callback, then I assumed following for Func:
- The result returned by Func would be set as error. (If it throws exception, then exception would be set instead)
- Func is constrained to always return some value other than void as its result is being used to set the error.
- Currently Func is constrained to not accept any arguments as parameter. This decision is induced by ``unifex::let_error`` implementation. However, it seems so limiting to me...

Currently we can code something like this:
```cpp
  int err_code = 0;
  try {
    just_error(42)
      | upon_error([]{ return 2; })
      | sync_wait();
  } catch (int err) {
    err_code = err;
  }
```

However, if upon_error callback is allowed to accept previous error(optionally) then we can also code something like this:
```cpp
  int err_code = 0;
  try {
    just_error(42)
      | upon_error([](auto x){ return x; })
      | sync_wait();
  } catch (int err) {
    err_code = err; // err_code is 42 now
  }
```

ericniebler:
Apologies for the delay. With `upon_error(s, f)`, if `set_error(e)` is called, then `invoke(f, e)` is executed and the result _are passed to `set_value(r, result)`_, not `set_error(r, result)`. I think that should clear up your confusion.

As a refinement, if `invoke(f, e)` returns void, then it executes `set_value(r)`.

Likewise, with `upon_done` when `set_done` is called the function is evaluated and the result, if any, is passed to `set_value`, not `set_done`.

LeeHowes:
The spec for the upon_ functions isn't very clear. I think the right answer depends on the goal. It could:
 1) Mutate the error and be a function from input_error_type -> output_error_type
 2) Consume the error: input_error_type -> prior_value_type (with the ability to throw an exception to regenerate an error)
 3) Consume the error: input_error_type -> new_value_type (with the ability to throw an exception to regenerate an error)
 4) Observe the error: input_error_type -> void

The way P2300 specifies this it appears to take option 3 - which could mean that we introduce new value type.

I think this is why P1897 didn't try to define Functor-like versions of these and stuck with the more obvious monadic versions of let_error and let_done.

ericniebler:
You're right, the `upon_*` functions have super weird semantics. Are they needed?

LeeHowes:
There is a symmetry argument. I don't know which way it goes, though. I think there's an argument for this particular set so that you have a convenient way of consuming errors  - it is analogous to try/catch in the sense that catch consumes an error and puts you back in a non-error state. It's clunky in that like catch you have to throw to get back in an error state, but if rethrowing is seen as rare the symmetry is there.

Functions would enforce single return type, of course. We don't have to do that - the error handler can actually change the return type and propagate multiple potential value types from then on.

ericniebler (approved):

ericniebler:
This is a great contribution, @RishabhRD. Thank you. If there's more from P2300 you'd like to implement in unifex, PRs are most welcome. :-)

RishabhRD:
Thanks @ericniebler for making the process so smooth. I will try to implement more things soon.

PR #334, opened 2021-08-25
get_scheduler is no longer required to be noexcept

ericniebler:


include/unifex/scheduler_concepts.hpp
```diff
@@ -341,12 +343,15 @@ namespace _schedule_at {
...
+                get_scheduler_result_t<const remove_cvref_t<Receiver>&>,
                 const TimePoint&>,
             Receiver> {
-      auto scheduler = get_scheduler(std::as_const(r));
-      return connect(schedule_at(scheduler, std::as_const(s.time_point_)), (Receiver&&) r);
+      return connect(
+          schedule_at(
+              get_scheduler(std::as_const(r)),
+              std::as_const(s.time_point_)),
+          (Receiver&&) r);
```
lewissbaker, line 354: This change is dangerous.
If `connect()` customisation happens to take the receiver by-value then the receiver might be in a moved-from state by the time that the `get_scheduler()` CPO is called.
This is why the code was structured the way that it was as two separate statements.

lewissbaker:
Is there a description of the rationale for this change somewhere?
What is the use-case that you are running into that is problematic?

The coroutine task use-case could make use of `any_ref` to type-erase the parent's scheduler.

> I would like to get @lewissbaker thoughts on this. Removing noexcept from get_scheduler to support type-erased schedulers is going to affect usage of get_scheduler in start()

Have we explored making use of reference-counted `any_shared`, or an always-inline-storage `any_inline_value` for type-erased schedulers that support non-throwing copying?

I guess that doesn't help, though, unless you have already pre-constructed an `any_shared` and are just returning a copy of that in `get_scheduler()`.

This is related to but separate from whether or not the scheduler concept should require `is_nothrow_copy_constructible` (which is something I would like to do).

ericniebler:
Closing this since the approach is wrong and I don't know when I'll get back to it.

PR #339, opened 2021-09-01
Introduce optional overrides for resource limits in tests

egorich239:
The default values of number of spawned threads and mutex iterations might be too high for some runtime environments.


ispeters:
Thanks!

PR #340, opened 2021-09-02
Sudoku example

kirkshoop:
ported the sudoku example from TBB examples/thread_group/sudoku


ericniebler:
Do we really want to merge my change to `get_scheduler` to make it possibly `noexcept(false)`? Lewis' feedback is that that is going to be problematic, and I haven't had a chance to go back and do different.

AnujYamdagni (changes requested):

PR #344, opened 2021-09-05
back out fix to just[_error] and fix let_value instead

ericniebler:
Fix the problem in the receivers, not the senders, per @kirkshoop

### DIFF
--- a/include/unifex/let_value.hpp
+++ b/include/unifex/let_value.hpp
@@ -56,12 +56,12 @@ struct _successor_receiver<Operation, Values...>::type {
 
   template <typename... SuccessorValues>
   void set_value(SuccessorValues&&... values) && noexcept {
-    cleanup();
     UNIFEX_TRY {
       // Taking by value here to force a copy on the offchance the value
       // objects lives in the operation state (e.g., just), in which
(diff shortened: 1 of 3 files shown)


include/unifex/let_value.hpp
```diff
@@ -58,8 +58,13 @@ struct _successor_receiver<Operation, Values...>::type {
...
     UNIFEX_TRY {
-      unifex::set_value(
-          std::move(op_.receiver_), (SuccessorValues &&) values...);
+      // Taking by value here to force a copy on the offchance the value
+      // objects lives in the operation state (e.g., just), in which
+      // case the call to cleanup() would invalidate them.
+      [this](auto... copies) {
+        unifex::set_value(
+            std::move(op_.receiver_), (decltype(copies) &&) copies...);
+      } ((SuccessorValues&&) values...);
```
kirkshoop, line 67: The comment for this lambda indicates that it must preceed the call to cleanup(), yes?

ericniebler: 🤦‍♂️

kirkshoop (approved):
👍🏻

PR #345, opened 2021-09-09
[unifex] Add let_value_with_stop_token algorithm

MarkCMann:
Adding a new sender algorithm to enable hooking cancellation callbacks onto the receivers stop token directly.

### DIFF
--- a/include/unifex/let_value_with_stop_token.hpp
+++ b/include/unifex/let_value_with_stop_token.hpp
@@ -1,0 +1,178 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <unifex/config.hpp>
+#include <unifex/just.hpp>
+#include <unifex/let_value.hpp>
+#include <unifex/receiver_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
+#include <unifex/inplace_stop_token.hpp>
+#include <unifex/execution_policy.hpp>
+#include <unifex/type_list.hpp>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+
+namespace _let_v_w_stop_tok {
+
+template<typename Operation, typename Receiver>
+struct _stop_token_receiver {
+    class type;
+};
+
+template<typename InnerOp, typename Receiver>
+struct _stop_token_operation {
+  struct type;
+};
+
+template <typename InnerOp, typename Receiver>
+using operation =  typename _stop_token_operation<InnerOp, remove_cvref_t<Receiver>>::type;
+
+template<typename Operation, typename Receiver>
+using stop_token_receiver = typename _stop_token_receiver<Operation, Receiver>::type;
+
+template<typename Operation, typename Receiver>
+class _stop_token_receiver<Operation, Receiver>::type {
+public:
+    explicit type(Operation& op, Receiver&& r) noexcept(std::is_nothrow_move_constructible_v<Receiver>) :
+        op_(&op), receiver_{(Receiver &&)r}
+    {}
+
+    template(typename... Values)
+        (requires receiver_of<Receiver, Values...>)
+    void set_value(Values&&... values) noexcept(is_nothrow_receiver_of_v<Receiver, Values...>) {
+        unifex::set_value(std::move(receiver_), (Values&&)values...);
+    }
+
+    template(typename Error)
+        (requires receiver<Receiver, Error>)
+    void set_error(Error&& error) noexcept {
+        unifex::set_error(std::move(receiver_), (Error&&)error);
+    }
+
+    void set_done() noexcept {
+        unifex::set_done(std::move(receiver_));
+    }
+
+    friend inplace_stop_token tag_invoke(tag_t<get_stop_token>, const type& r) noexcept {
+        return r.op_->stop_token_;
+    }
+
+    template(typename CPO, typename Self)
+        (requires
+            is_receiver_query_cpo_v<CPO> AND
+            same_as<Self, type>)
+    friend auto tag_invoke(CPO cpo, const Self& self)
+        noexcept(is_nothrow_callable_v<CPO, const Receiver&>)
+        -> callable_result_t<CPO, const Receiver&> {
+        return cpo(self.receiver_);
+    }
+
+private:
+    Operation* op_;
+    UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+};
+
+template<typename SuccessorFactory>
+struct _stop_token_sender {
+    class type;
+};
+
+template<typename SuccessorFactory>
+using stop_token_sender = typename _stop_token_sender<SuccessorFactory>::type;
+
+template<typename SuccessorFactory>
+class _stop_token_sender<SuccessorFactory>::type {
+    template<typename... Values>
+    using additional_arguments = type_list<type_list<Values...>>;
+
+public:
+    using InnerOp = std::invoke_result_t<SuccessorFactory, inplace_stop_token&>;
+
+    template<template<typename...> class Variant, template<typename...> class Tuple>
+    using value_types = sender_value_types_t<InnerOp, Variant, Tuple>;
+
+    template<template<typename...> class Variant>
+    using error_types = sender_error_types_t<InnerOp, Variant>;
+
+    static constexpr bool sends_done = sender_traits<InnerOp>::sends_done;
+
+    template<typename SuccessorFactory2>
+    explicit type(SuccessorFactory2&& func) : func_((SuccessorFactory2&&)func)
+    {}
+
+    template(typename Self, typename Receiver)
+        (requires same_as<remove_cvref_t<Self>, type> AND receiver<Receiver>)
+    friend auto tag_invoke(tag_t<unifex::connect>, Self&& self, Receiver&& r)
+        noexcept(
+            std::is_nothrow_invocable_v<member_t<Self, SuccessorFactory>, inplace_stop_token&> &&
+            std::is_nothrow_constructible_v<remove_cvref_t<Receiver>, Receiver>)
+        -> operation<member_t<Self, SuccessorFactory>, Receiver> {
+        return operation<member_t<Self, SuccessorFactory>, Receiver>(
+            static_cast<Self&&>(self).func_, static_cast<Receiver&&>(r));
+    }
+
+private:
+    UNIFEX_NO_UNIQUE_ADDRESS SuccessorFactory func_;
+};
+
+template<typename SuccessorFactory, typename Receiver>
+struct _stop_token_operation<SuccessorFactory, Receiver>::type {
+
+    template <typename SuccessorFactory2, typename Receiver2>
+    type(SuccessorFactory2&& func, Receiver2&& r) :
+        func_{(SuccessorFactory2&&)func},
+        stop_token_{stop_token_adapter_.subscribe(get_stop_token(r))},
+        innerOp_(
+              unifex::connect(
+                ((SuccessorFactory&&)func_)(stop_token_),
+                stop_token_receiver<operation<SuccessorFactory, Receiver>, remove_cvref_t<Receiver2>>{
+                    *this,
+                    static_cast<Receiver2&&>(r)})) {
+    }
+
+    void start() noexcept {
+        unifex::start(innerOp_);
+    }
+
+    UNIFEX_NO_UNIQUE_ADDRESS SuccessorFactory func_;
+    inplace_stop_token_adapter<stop_token_type_t<Receiver>> stop_token_adapter_;
+    inplace_stop_token stop_token_;
+    connect_result_t<
+        callable_result_t<SuccessorFactory, unifex::inplace_stop_token&>,
+        stop_token_receiver<operation<SuccessorFactory, Receiver>, remove_cvref_t<Receiver>>>
+        innerOp_;
+};
+
+namespace _cpo {
+struct _fn {
+    template<typename SuccessorFactory>
+    auto operator()(SuccessorFactory&& factory) const
+        noexcept(std::is_nothrow_constructible_v<std::decay_t<SuccessorFactory>, SuccessorFactory>)
+        -> stop_token_sender<std::decay_t<SuccessorFactory>> {
+        return stop_token_sender<std::decay_t<SuccessorFactory>>{(SuccessorFactory&&)factory};
+    }
+};
+} // namespace _cpo
+} // namespace _let_v_w_stop_tok
+
+inline constexpr _let_v_w_stop_tok::_cpo::_fn let_value_with_stop_token{};
+
+} // namespace unifex
+
+#include <unifex/detail/epilogue.hpp>
--- a/test/let_value_with_stop_token_test.cpp
+++ b/test/let_value_with_stop_token_test.cpp
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <unifex/just.hpp>
+#include <unifex/any_sender_of.hpp>
+#include <unifex/just_done.hpp>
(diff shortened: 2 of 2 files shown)


include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,178 @@
...
+template<typename SuccessorFactory, typename Receiver>
+struct _stop_token_operation<SuccessorFactory, Receiver>::type {
+
+    template <typename SuccessorFactory2, typename Receiver2>
+    type(SuccessorFactory2&& func, Receiver2&& r) :
+        func_{(SuccessorFactory2&&)func},
+        stop_token_{stop_token_adapter_.subscribe(get_stop_token(r))},
+        innerOp_(
+              unifex::connect(
+                ((SuccessorFactory&&)func_)(stop_token_),
```
lewissbaker, line 143: You are invoking the function immediately on connect() here, and this in turn requires calling `.subscribe()` on the adapter which in turn (potentially) registers a callback on parent receiver's stop-token.

Normally, I try to avoid registering a stop-callback until start() is called, I think mainly just in case the operation is never started and so you can avoid the synchronisation needed to register the callback, but also because deferring it until start() means we often can avoid needing to store an extra bool/optional that keeps track of whether or not the stop-callback is registered (e.g. if the destructor is called without calling start()).

However, doing this would mean you couldn't get the stop-token from the adapter until start(), requiring you to defer calling the function until start(), which would in turn mean deferring calling connect() on the returned sender until start().

Other let algorithms generally seem to call the lambda passed to them at connect() time rather than at start() time so making this change would be inconsistent with them.

You may have other options if you avoid using the adapter class and instead just implement different specialisations of the operation-state based on the type of the stop-token from the parent receiver.

include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,178 @@
...
+    UNIFEX_NO_UNIQUE_ADDRESS SuccessorFactory func_;
+};
+
+template<typename SuccessorFactory, typename Receiver>
+struct _stop_token_operation<SuccessorFactory, Receiver>::type {
+
+    template <typename SuccessorFactory2, typename Receiver2>
+    type(SuccessorFactory2&& func, Receiver2&& r) :
+        func_{(SuccessorFactory2&&)func},
+        stop_token_{stop_token_adapter_.subscribe(get_stop_token(r))},
```
lewissbaker, line 140: bug: You are calling `.subscribe()` here but never calling `.unsubscribe()`.

The `inplace_stop_token_adapter` type is like `manual_lifetime` in that it requires you to explicitly unsubscribe to clean it up - it doesn't keep track of whether or not there is a current subscription active or not.

Also, be careful to handle the case where the call to `connect()` below throws.

include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,178 @@
...
+    class type;
+};
+
+template<typename SuccessorFactory>
+using stop_token_sender = typename _stop_token_sender<SuccessorFactory>::type;
+
+template<typename SuccessorFactory>
+class _stop_token_sender<SuccessorFactory>::type {
+    template<typename... Values>
+    using additional_arguments = type_list<type_list<Values...>>;
```
lewissbaker, line 102: This typedef seems unused.

include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+private:
+    UNIFEX_NO_UNIQUE_ADDRESS SuccessorFactory func_;
+};
+
+template<typename SuccessorFactory, typename Receiver>
+struct _stop_token_operation<SuccessorFactory, Receiver>::type {
+
+    template <typename SuccessorFactory2, typename Receiver2>
+    type(SuccessorFactory2&& func, Receiver2&& r) :
```
lewissbaker, line 138: Consider making this conditionally noexcept (although I realise that would be a lengthy expression as there's a lot going on in this constructor).

lewissbaker (commented):
Here's a few comments from my first pass over this PR.
I'll have a bit more of a think about some of the issues and whether there are better strategies.

ispeters:
Can you add some documentation in `doc/api_reference.md` to mention this algorithm and how it works?

include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,178 @@
...
+using operation =  typename _stop_token_operation<InnerOp, remove_cvref_t<Receiver>>::type;
+
+template<typename Operation, typename Receiver>
+using stop_token_receiver = typename _stop_token_receiver<Operation, Receiver>::type;
+
+template<typename Operation, typename Receiver>
+class _stop_token_receiver<Operation, Receiver>::type {
+public:
+    explicit type(Operation& op, Receiver&& r) noexcept(std::is_nothrow_move_constructible_v<Receiver>) :
+        op_(&op), receiver_{(Receiver &&)r}
```
ispeters, line 53: The formatting is off here,

```suggestion
    explicit type(Operation& op, Receiver&& r) noexcept(std::is_nothrow_move_constructible_v<Receiver>)
        : op_(&op), receiver_{(Receiver &&)r}
```

include/unifex/let_value_with_stop_token.hpp
MarkCMann: I'm not sure what the right approach here should be. From my perspective, it makes sense to invoke the lambda in the call to connect() because that's required in order to connect the operation chain. In the case that lambda invocation is delayed to start() that introduces strange behavior with upstream senders where calls to connect() with them will be delayed until start() is called on this operation state. 

I attempted to remove the inplace_stop_token_adapter, but I ran into the problem where the Sender's value types are stored on the type of the Sender which requires knowing which Receiver will eventually be connected at the time of the Sender's construction. I think we said that one way around this was storing the value types on the operation state, but that seems like a large change. I'm open to exploring other ways to remove the stop_token_adapter from this algorithm though.

include/unifex/let_value_with_stop_token.hpp
MarkCMann: I think that I've made all of the functions in this file appropriately noexcept, although it seems that it made a lot of the code a bit harder to read.

include/unifex/let_value_with_stop_token.hpp
lewissbaker: One approach is to provide different specialisations of the operation-state based on what the receiver's stop-token type is.
e.g.
```c++
// generic one that adapts
template<typename Func, typename Receiver, typename StopTokenType, typename = void>
struct operation_state {
  struct receiver {
    operation_state& op_;
    
    template<typename... Values>
    void set_value(Values&&... values) noexcept {
      op_.stop_callback_.destroy();
       try { unifex::set_value(std::move(op_.receiver_), (Values&&)values...); }
       catch (...) { unifex::set_error(std::move(op_.receiver_), std::current_exception()); }
    }
   
    // etc. for set_error/set_done and also for other queries
    
    friend inplace_stop_token tag_invoke(tag_t<get_stop_token>, const receiver& r) {
      return op.stop_source_.get_token();
    }
  };
  
  void start() noexcept {
    stop_callback_.construct(get_stop_token(receiver_), detail::forward_stop_request_to_inplace_stop_source{stop_source_});
    unifex::start(sourceOp_);
  }
  
  using callback_type = typename StopTokenType::template callback_type<detail::forward_stop_request_to_inplace_stop_source>;
  
  using Source = invoke_result_t<Func, inplace_stop_token>;
  
  Receiver receiver_;
  Func func_;
  inplace_stop_source stop_source_;
  manual_lifetime<callback_type> stop_callback_;
  connect_result_t<Source, receiver> sourceOp_;
};

template<typename Func, typename Receiver, typename StopTokenType>
struct operation_state<Func, Receiver, StopTokenType, std::enable_if_t<is_stop_never_possible_v<StopTokenType>>> {

  template<typename Func2>
  operation_state(Func2&& func, Receiver&& r)
  : func_((Func2&&)func)
  , sourceOp_(unifex::connect(func_(inplace_stop_token()), (Receiver&&)r))
  {}

  void start() noexcept {
    unifex::start(sourceOp_);
  }

  using Source = invoke_result_t<Func, inplace_stop_token>;

  Func func_;
  connect_result_t<Source, Receiver> sourceOp_;
};

template<typename Func, typename Receiver>
struct operation_state<Func, Receiver, inplace_stop_token, void> {

  template<typename Func2, typename Receiver2>
  operation_state(Func2&& f, Receiver2&& r)
  : func_((Func2&&)f)
  , sourceOp_(make_source_op((Receiver2&&)r, get_stop_token(r)))
  {}

  void start() noexcept {
    unifex::start(sourceOp_);
  }

  using Source = invoke_result_t<Func, inplace_stop_token>;

  template<typename Receiver2>
  Source make_source_op(Receiver2&& r, inplace_stop_token st) {
    return unifex::connect(((Func&&)func_)(std::move(st)), (Receiver2&&)r);
  }

  Func func_;
  connect_result_t<Source, Receiver>
};
```

include/unifex/let_value_with_stop_token.hpp
lewissbaker: Yep, if only we had `noexcept(auto)`...

doc/api_reference.md
```diff
@@ -349,6 +350,28 @@ cancellation of the operation returned by the function. Note that cancellation
...
+created by subscribing the `inplace_stop_token_adapter` to the stop token
+returned by the invocation of `get_stop_token` on the provided receiver.
+This invocation must return a Sender.
+
+The reference passed to `func` will remain valid until the returned sender
+completes, at which point the `inplace_stop_token_adapter` goes out of scope.
+
+For example:
+```c++
+let_value_with_stop_token(
```
lewissbaker, line 369: I wonder if this should be named more simply `let_stop_token` or `let_current_stop_token` instead of `let_value_with_stop_token`.

doc/api_reference.md
```diff
@@ -349,6 +350,28 @@ cancellation of the operation returned by the function. Note that cancellation
...
+`inplace_stop_token_adapter` that remains alive for the duration of an
+operation.
+
+`func` is invoked with an lvalue reference to an `inplace_stop_token`
+created by subscribing the `inplace_stop_token_adapter` to the stop token
+returned by the invocation of `get_stop_token` on the provided receiver.
+This invocation must return a Sender.
+
+The reference passed to `func` will remain valid until the returned sender
+completes, at which point the `inplace_stop_token_adapter` goes out of scope.
```
lewissbaker, line 365: I don't think the documentation here needs to refer to the use of `inplace_stop_token_adapter` as that is an implementation detail.

It just needs to say something along the lines of:
> The `let_value_with_stop_token()` algorithm takes a function object that is invoked at `unifex::connect()` time with an `inplace_stop_token` object that can be used to receive a stop-request sent by the parent operation through the receiver it passes to `connect()`.
>
> The function must return a sender, which is immediately connected. The result of the sender returned from the function becomes the result of the `let_value_with_stop_token()` sender.
>
> The function is guaranteed to be kept alive until the operation completes.
>
> The stop-token passed to the function is only guaranteed to be valid until the sender returned by the function completes.

include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,234 @@
...
+    return r.stop_token_;
+  }
+
+  template(typename CPO, typename Self)(
+      requires is_receiver_query_cpo_v<CPO> AND same_as<
+          Self,
+          type>) friend auto tag_invoke(CPO cpo, const Self& self) noexcept(is_nothrow_callable_v<CPO, const Receiver&>)
+      -> callable_result_t<CPO, const Receiver&> {
+    return cpo(self.receiver_);
+  }
```
ispeters, line 81: The formatting here is what Eric meant when he said Clang Format "makes a hash" of the concept-alike macros.
```suggestion
  template(typename... Values)
      (requires receiver_of<Receiver, Values...>)
  void set_value(Values&&... values) noexcept(is_nothrow_receiver_of_v<Receiver, Values...>) {
    unifex::set_value(std::move(receiver_), (Values &&) values...);
  }

  template(typename Error)
      (requires receiver<Receiver, Error>)
  void set_error(Error&& error) noexcept {
    unifex::set_error(std::move(receiver_), (Error &&) error);
  }

  void set_done() noexcept { unifex::set_done(std::move(receiver_)); }

  friend inplace_stop_token
  tag_invoke(tag_t<get_stop_token>, const type& r) noexcept {
    return r.stop_token_;
  }

  template(typename CPO, typename Self)
      (requires is_receiver_query_cpo_v<CPO> AND same_as<Self, type>)
  friend auto tag_invoke(CPO cpo, const Self& self) noexcept(is_nothrow_callable_v<CPO, const Receiver&>)
      -> callable_result_t<CPO, const Receiver&> {
    return cpo(self.receiver_);
  }
```

I have had success "tricking" Clang Format by `#if 0`ing out the `template` and `requires` clauses, running Clang Format, and then manually formatting the `template` and `requires` clauses afterwards.

include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,234 @@
...
+
```
ispeters, line 82: Above, you handle most receiver queries, but you're not handling `visit_continuations` (e.g. [`let_error`s implementation](https://github.com/facebookexperimental/libunifex/blob/0947fbbb2af4595ecad2a9a780c3e41c9537ab65/include/unifex/let_error.hpp#L133)).

include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,234 @@
...
+          r) noexcept(is_nothrow_invocable_v<member_t<Self, SuccessorFactory>, inplace_stop_token&>&&
+                          is_nothrow_constructible_v<
+                              remove_cvref_t<Receiver>,
+                              Receiver>&&
+                              is_nothrow_constructible_v<
+                                  operation<
+                                      member_t<Self, SuccessorFactory>,
+                                      unifex::remove_cvref_t<Receiver>>,
+                                  SuccessorFactory&&,
+                                  Receiver&&>)
```
ispeters, line 132: Clang Format has really done a number on this one. :(

include/unifex/let_value_with_stop_token.hpp
```diff
@@ -0,0 +1,234 @@
...
+public:
+  template <typename SuccessorFactory2, typename Receiver2>
+  type(SuccessorFactory2&& func, Receiver2&& r) noexcept(
+      is_nothrow_constructible_v<SuccessorFactory, SuccessorFactory2&&>&&
+          is_nothrow_constructible_v<std::decay_t<Receiver2>, Receiver2>&& noexcept(
+              connect_inner_op(
+                  std::declval<SuccessorFactory2&&>(),
+                  std::declval<inplace_stop_token>(),
+                  std::declval<Receiver2&&>())))
+    : innerOp_(connect_inner_op(
```
ispeters, line 205: GCC seems to want some `this`es.  Also, since the `noexcept` clause follows the argument list, you can refer to the arguments here instead of using `declval`, which I find clearer.

```suggestion
              this->connect_inner_op(
                  (SuccessorFactory &&) func,
                  stop_token_adapter_.subscribe(get_stop_token(r)),
                  (Receiver2 &&) r)))
    : innerOp_(connect_inner_op(
```

test/let_value_with_stop_token_test.cpp
```diff
@@ -0,0 +1,99 @@
...
+                });
+            });
+          })
+      );
+
+  EXPECT_TRUE(!result);
+  EXPECT_EQ(external_context, 42);
+}
+
+TEST(LetWithStopToken, SimpleNoExceptSuccessor) {
```
ispeters, line 70: How's this test case different from the one above?  Did you mean to add a `noexcept` somewhere?

doc/api_reference.md
MarkCMann: I'm not too familiar with the naming of the other let_* operators, but I used the same naming scheme as the `let_value_with_stop_source` -- I think it'd be okay to rename this operator but then it would inconsistent with the stop source variant.

test/let_value_with_stop_token_test.cpp
MarkCMann: In this test case the lambda passed to the `let_value_with_stop_token` is noexcept. Maybe this is a meaningless test? I guess I should probably add a static_assert that the operator is `noexcept`?

ispeters: Oh, I see the difference, now.  It _has_ occurred to me that Unifex would benefit from `static_assert`s in the tests that validate our conditional `noexcept` clauses.  I think we don't currently have that anywhere else so it's not a reasonable bar to hold you to, but it'd be a valuable addition.

test/let_value_with_stop_token_test.cpp
```diff
@@ -0,0 +1,273 @@
...
+          return let_value_with(
+              [&]() noexcept {
+                auto stopCallback = [&]() noexcept {
+                  external_context = 42;
+                };
+                using stop_token_t =
+                    unifex::remove_cvref_t<decltype(stopToken)>;
+                using stop_callback_t =
+                    typename stop_token_t::template callback_type<
+                        decltype(stopCallback)>;
```
ispeters, line 93:
```suggestion
        return let_value_with_stop_token([&](inplace_stop_token stopToken) {
          return let_value_with(
              [&]() noexcept {
                auto stopCallback = [&]() noexcept {
                  external_context = 42;
                };
                using stop_callback_t = inplace_stop_token::callback_type<decltype(stopCallback)>;
```

ispeters (approved):
A few nits, but LGTM.

test/let_value_with_stop_token_test.cpp
```diff
@@ -0,0 +1,241 @@
...
+      let_value_with_stop_token(std::move(stop_token_functor)),
+      InplaceStoppableIntReceiver{stopSource});
+  unifex::start(op);
+
+  EXPECT_EQ(external_context, 42);
+}
+
+TEST(LetWithStopToken, SimpleUnstoppable) {
+  static_assert(is_stop_never_possible_v<
+                stop_token_type_t<UnstoppableSimpleIntReceiver>>);
```
ispeters, line 149: Huh.  This `static_assert` is failing when building with MSVC 2019 in C++20 mode.  Weird.

PR #347, opened 2021-09-14
make let_error pass error value to factory function

kirkshoop:
let_error required the Factory function to take no args and ignored the error value passed to set_error.
These changes remove the constraints on the function and calculate all the possible senders that the Factory function will return for different types of error values. set_error now passes the error value to the Factory function.


include/unifex/let_error.hpp
```diff
@@ -208,19 +220,20 @@ class _frcvr<Source, Error, Receiver>::type {
   operation* op_;
 };
 
-template <typename Source, typename Error, typename Receiver>
-class _op<Source, Error, Receiver>::type {
-  using source_receiver = receiver_type<Source, Error, Receiver>;
-  using final_receiver = final_receiver_type<Source, Error, Receiver>;
+template <typename Source, typename Func, typename Receiver>
+class _op<Source, Func, Receiver>::type {
```
lewissbaker, line 224: You've changed the template parameter from `Error` to `Func` here, which is a great improvement.
But there are still a number of other uses of the `Error` template parameter name in receiver types above that could also do with changing as well for consistency.

include/unifex/let_error.hpp
```diff
@@ -208,19 +220,20 @@ class _frcvr<Source, Error, Receiver>::type {
...
+  using source_receiver = receiver_type<Source, Func, Receiver>;
+  using final_receiver = final_receiver_type<Source, Func, Receiver>;
 
 public:
-  template <typename Error2, typename Receiver2>
-  explicit type(Source&& source, Error2&& error, Receiver2&& dest)
+  template <typename Func2, typename Receiver2>
+  explicit type(Source&& source, Func2&& error, Receiver2&& dest)
```
lewissbaker, line 230: Can we change the parameter name from `error` to `func` here to match the template parameter name?
And maybe also the member name?

include/unifex/let_error.hpp
```diff
@@ -252,68 +265,108 @@ class _op<Source, Error, Receiver>::type {
...
+struct sends_done_impl : std::bool_constant<sender_traits<Sender>::sends_done> {};
+
+template<typename... Senders>
+using any_sends_done = std::disjunction<sends_done_impl<Senders>...>;
+
+template <typename Source, typename Func>
+class _sndr<Source, Func>::type {
+
+  template <typename Error>
+  using final_sender = callable_result_t<Func&, remove_cvref_t<Error>>;
```
lewissbaker, line 304: You are using `Func&` here but you are actually calling a moved func in the implementation.

include/unifex/let_error.hpp
```diff
@@ -252,68 +265,108 @@ class _op<Source, Error, Receiver>::type {
...
+
+  template <template <typename...> class Tuple>
+  struct make_value_type_list {
+    template<typename... Errors>
+    using make = typename concat_type_lists_unique<
+      sender_value_types_t<final_sender<Errors>, type_list, Tuple>...>::type;
```
lewissbaker, line 310: Be careful using `concate_type_lists_unique` when passing the user-provided `Tuple` or `Variant` as elements.
e.g. what if tuple was `template<typename...> Tuple = void;`

Usually I build everything uniquely using `type_list` and then use the `type_list_nested_apply_t` helper to apply the user-provided `Tuple` and `Variant` versions.

include/unifex/let_error.hpp
```diff
@@ -252,68 +265,108 @@ class _op<Source, Error, Receiver>::type {
...
   using error_types =
       typename concat_type_lists_unique_t<
-          sender_error_types_t<Source, type_list>,
-          sender_error_types_t<final_sender_t, type_list>,
+          sender_error_types_t<Source, make_error_type_list>,
           type_list<std::exception_ptr>>::template apply<Variant>;
 
-  static constexpr bool sends_done = sender_traits<final_sender_t>::sends_done;
+  static constexpr bool sends_done = 
+    sender_traits<Source>::sends_done |
```
lewissbaker, line 340: `||` instead of `|`?

include/unifex/let_error.hpp
```diff
@@ -91,28 +92,39 @@ class _rcvr<Source, Error, Receiver>::type final {
...
         });
+        op->deactivate_ = [](auto& op) noexcept {
+          unifex::deactivate_union_member<final_op_t>(op);
+        };
         op->startedOp_ = state::final;
-        unifex::start(op->finalOp_.get());
+        unifex::start(finalOp);
       } UNIFEX_CATCH (...) {
         unifex::set_error(std::move(op->receiver_), std::current_exception());
       }
```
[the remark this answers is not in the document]
egorich239: I removed the no-throw branch of this condition for now, for the consistency with `let_value` behavior.

include/unifex/let_error.hpp
```diff
@@ -201,167 +206,202 @@ class _frcvr<Source, Error, Receiver>::type {
...
+      //   sender<callable_result_t<remove_cvref_t<Error>>>)
+  constexpr auto operator()(Func&& func) const
       noexcept(is_nothrow_callable_v<
-        tag_t<bind_back>, _fn, Error>)
-      -> bind_back_result_t<_fn, Error> {
-    return bind_back(*this, (Error&&)error);
+        tag_t<bind_back>, _fn, Func>)
+      -> bind_back_result_t<_fn, Func> {
+    return bind_back(*this, (Func&&)func);
   }
```
kirkshoop, line 405: IMO, It is a good idea to remove the commented out constraints for these operator() overloads. They are misleading since they refer to Error instead of Func and they expect the func to be callable with no args, which is no longer supported.

kirkshoop (commented):
Looks great! one tiny cleanup and it is ready to merge. (as author of this PR I am unable to approve)

include/unifex/let_error.hpp
egorich239: We could probably specify that `Func` is a callable for all possible tuples of `Sender`'s `error_types` but that won't look pretty, so I just deleted these comments indeed.

include/unifex/let_error.hpp
```diff
@@ -33,89 +34,89 @@
...
-  {}
- 
   template(typename... Values)
     (requires receiver_of<Receiver, Values...>)
   void set_value(Values&&... values) noexcept(
       is_nothrow_receiver_of_v<Receiver, Values...>) {
     UNIFEX_ASSERT(op_ != nullptr);
+    unifex::deactivate_union_member(op_->sourceOp_);
     unifex::set_value(std::move(op_->receiver_), (Values&&)values...);
   }
```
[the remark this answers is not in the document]
egorich239: These operations cannot be easily flipped - `op_` may be destroyed as a side-effect of calling `set_value` (e.g. because the final receiver has already destroyed the full state of the operation). I changed the types in `set_value()` to `Values... values`, similar to how `set_error()` accepts error by value in `let_value` combinator

include/unifex/type_list.hpp
```diff
@@ -33,6 +33,36 @@ namespace unifex
...
 
+  // apply_to_type_list<F, type_list<Ts...>>
+  //
+  // Shortcut to `typename type_list<Ts...>::template apply<F>`.
+  template <template <typename...> class F, typename List>
+  struct apply_to_type_list;
+
+  template <template <typename...> class F, typename... Ts>
+  struct apply_to_type_list<F, type_list<Ts...>> {
+    using type = typename type_list<Ts...>::template apply<F>;
```
lewissbaker, line 44: This could be simplified to just: `using type = F<Ts...>;`

include/unifex/type_list.hpp
```diff
@@ -33,6 +33,36 @@ namespace unifex
...
+  };
+
+  template <template <typename...> class F, typename List>
+  using apply_to_type_list_t = typename apply_to_type_list<F, List>::type;
```
lewissbaker, line 48: Or alternatively, this could be simplified to just:
```
template <template <typename...> class F, typename List>
using apply_to_type_list_t = typename List::template apply<F>;
```

include/unifex/type_list.hpp
egorich239: I chose this option.

include/unifex/type_list.hpp
egorich239: I decided against this option though, as I decided to enforce that `List` is a `type_list<Ts...>`.

include/unifex/let_error.hpp
```diff
@@ -201,167 +213,198 @@ class _frcvr<Source, Error, Receiver>::type {
...
-      noexcept(std::is_nothrow_move_constructible_v<Receiver> &&
-               std::is_nothrow_move_constructible_v<Error> &&
-               is_nothrow_connectable_v<Source, source_receiver>)
-  : error_((Error2&&)error)
-  , receiver_((Receiver2&&)dest)
-  {
+  template <typename Func2, typename Receiver2>
+  explicit type(Source&& source, Func2&& func, Receiver2&& dest) noexcept(
+      std::is_nothrow_move_constructible_v<Receiver>&&
+          std::is_nothrow_move_constructible_v<Func>&&
```
ispeters, line 233: These feel like the wrong `noexcept` specifications; you're not necessarily move-constructing either a `Receiver` or a `Func` below.

include/unifex/let_error.hpp
```diff
@@ -201,167 +213,198 @@ class _frcvr<Source, Error, Receiver>::type {
...
   friend auto tag_invoke(tag_t<unifex::connect>, Sender&& s, Receiver&& r)
        noexcept(
         is_nothrow_connectable_v<member_t<Sender, Source>, SourceReceiver> &&
-        std::is_nothrow_constructible_v<Error, member_t<Sender, Error>> &&
+        std::is_nothrow_constructible_v<Func, member_t<Sender, Func>> &&
         std::is_nothrow_constructible_v<remove_cvref_t<Receiver>, Receiver>)
-      -> operation_type<member_t<Sender, Source>, Error, Receiver> {
-    return operation_type<member_t<Sender, Source>, Error, Receiver>{
+      -> operation_type<member_t<Sender, Source>, Func, Receiver> {
+    return operation_type<member_t<Sender, Source>, Func, Receiver>{
```
ispeters, line 363: I think you should remove `member_t` here so that the operation is instantiated with a non-reference type on `Source`.
```suggestion
    return operation_type<Source, Func, Receiver>{
```

ispeters (approved):
LGTM!

PR #348, opened 2021-09-15
async_scope - separate cancellation and completion

kirkshoop:
I exposed the stop_token and request_stop() and moved the end_of_life() to happen in a stop_callback and in complete(). cleanup() is now in terms of request_stop() and complete()

### DIFF
--- a/include/unifex/async_scope.hpp
+++ b/include/unifex/async_scope.hpp
@@ -105,6 +105,20 @@ private:
   using _on_result_t =
     decltype(on(UNIFEX_DECLVAL(Scheduler&&), UNIFEX_DECLVAL(Sender&&)));
 
+  inplace_stop_source stopSource_;
+  // (opState_ & 1) is 1 until we've been stopped
+  // (opState_ >> 1) is the number of outstanding operations
+  std::atomic<std::size_t> opState_{1};
+  async_manual_reset_event evt_;
+
+  [[nodiscard]] auto await_and_sync() noexcept {
+    return then(evt_.async_wait(), [this]() noexcept {
+      // make sure to synchronize with all the fetch_subs being done while
+      // operations complete
+      (void)opState_.load(std::memory_order_acquire);
+    });
+  }
+
 public:
   async_scope() noexcept = default;
 
@@ -172,19 +186,15 @@ public:
         just_from([this] () noexcept {
           end_of_scope();
         }),
-        then(evt_.async_wait(), [this]() noexcept {
-            // make sure to synchronize with all the fetch_subs being done while
-            // operations complete
-            (void)opState_.load(std::memory_order_acquire);
-        }));
+        await_and_sync());
   }
 
   [[nodiscard]] auto cleanup() noexcept {
     return sequence(
         just_from([this]() noexcept {
-          stopSource_.request_stop();
+          request_stop();
         }),
-        complete());
+        await_and_sync());
   }
 
   inplace_stop_token get_stop_token() noexcept {
@@ -197,11 +207,6 @@ public:
   }
 
  private:
-  inplace_stop_source stopSource_;
-  // (opState_ & 1) is 1 until we've been stopped
-  // (opState_ >> 1) is the number of outstanding operations
-  std::atomic<std::size_t> opState_{1};
-  async_manual_reset_event evt_;
 
   static constexpr std::size_t stoppedBit{1};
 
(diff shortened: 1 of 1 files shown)


ispeters:
Looks like you've got two changes commingled here.  Could you separate out the changes to `let_error`?

include/unifex/async_scope.hpp
```diff
@@ -167,20 +167,44 @@ struct async_scope {
...
  private:
+  struct stop_callback {
+    async_scope& self_;
+    void operator()() noexcept {
+      self_.end_of_scope();
+    }
+  };
+
   inplace_stop_source stopSource_;
+  inplace_stop_token::template callback_type<stop_callback> stopCallback_;
```
ispeters, line 207: It feels unfortunate that this diff adds a stop callback, especially given that the stop source is encapsulated behind the interface and so the type controls when stop is requested.

Perhaps `end_of_scope` should take a `bool` indicating whether to request stop, and `request_stop` and `cleanup` can call it with `true`, while `complete` can call it with `false`.

kirkshoop: I originally planned to add a get_stop_source() method and that was why I took this approach. Ultimately, I left that method out because I was unsure of exporting a reference to the inplace_stop_source. In the current state of this PR your suggestion will work - but I hate bool params - they are always undocumented when reading the source.

kirkshoop: BTW, What is unfortunate about the callback? the overhead of calling spawn once should make this cost disappear, right?

kirkshoop:
good catch! the let_error changes are in PR #347  I forgot to remove them from this branch.

include/unifex/async_scope.hpp
ispeters: > What is unfortunate about the callback?

I think the thing that tingles my Spidey sense is that the callback adds "spooky action at a distance" between `request_stop` and `end_of_scope`.  It's both lexically disconnected and a more complicated coupling.  Incidentally, in the current implementation, `cleanup` deterministically calls `end_of_scope` twice (once in the stop callback and once in `complete`); it's idempotent, but it's also an atomic read-modify-write plus a branch.

I suspect you're right that the overhead of the complexity is small, especially compared to other operations that this type performs, but it feels like unnecessary overhead that's _also_ more complicated.

> I hate bool params - they are always undocumented when reading the source.

Fair; I also find `bool` params unwieldy.  My fall-back suggestion instead of `bool`s is usually to think of `bool` as the most generic possible `enum` and to use a more specific one instead, e.g.
```
enum class EndScopeStrategy {
  RequestStop,
  DoNotRequestStop
};
```

Another option would be to pass the address of the stop source and only request stop if it's not `nullptr`.

Last, I could imagine factoring it thus:
```
void stop_and_end_scope() {
  stopSource_.request_stop();
  end_scope();
}

void end_scope() {
  // ... muck about with opState_
}
```

I don't prefer this last option because it leaves a window where someone trying to spawn a task can observe "it's still ok to spawn work" _and_ "that work should immediately stop", but this is a) a narrow window, and b) probably not a real problem.

include/unifex/async_scope.hpp
```diff
@@ -167,18 +167,35 @@ struct async_scope {
...
         }));
   }
 
+  [[nodiscard]] auto cleanup() noexcept {
+    return sequence(
+        just_from([this]() noexcept {
+          stopSource_.request_stop();
+        }),
+        complete());
+  }
```
ispeters, line 188: This is pretty nitpicky of me, but it's jumping out at me that `cleanup()` and `request_stop()` run `stopSource_.request_stop()` and `end_of_scope()` in different orders.  If this were my diff, I'd add a private `await_and_sync()`, (or something) that looks like this:
```
[[nodiscard]] auto await_and_sync() noexcept {
  return then(evt_.async_wait(), [this]() noexcept {
    // make sure to synchronize with all the fetch_subs being done while
    // operations complete
    (void)opState_.load(std::memory_order_acquire);
  });
}
```

Then I'd write `complete` and `cleanup` like this:
```
[[nodiscard]] auto complete() noexcept {
  return sequence(
      just_from([this]() noexcept {
        end_of_scope();
      }),
      await_and_sync());
 }

[[nodiscard]] auto cleanup() noexcept {
  return sequence(
      just_from([this]() noexcept {
        request_stop();
      }),
      await_and_sync());
}
```

ispeters (approved):
One nit for you to consider, but LGTM.  Sorry this took so long to get to.

PR #349, opened 2021-09-16
add file_copy example

kirkshoop:
requires #347 and #348 

This is the first working version. Cleanup and improvements needed.
It implements a throttling algo adhoc. :(
It runs a bit slower than the std::filesystem version - I need someone with more linux knowledge to point me in the right direction. I speculate that std::filesystem does the entire copy in kernel by using a file copy syscall per file.
this is using liburing to actually open and read and write 64000 byte blocks from user mode.


AnujYamdagni (changes requested):

ispeters:
This PR is 4+ years old at this point and I don't think @kirkshoop is super interest in Unifex now that Nvidia's stdexec exists so I'm cleaning up the PR queue by closing this. Feel free to reopen if I've judged wrongly.

PR #350, opened 2021-09-17
P2300 into_variant implementation to libunifex

RishabhRD:
This PR intends to add implementation of P2300 ``std::execution::into_variant`` implementation to libunifex under name ``unifex::into_variant``

### DIFF
--- a/include/unifex/into_variant.hpp
+++ b/include/unifex/into_variant.hpp
@@ -1,3 +1,18 @@
+/*
+ * Copyright (c) Rishabh Dwivedi <rishabhdwivedi17@gmail.com>
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 #pragma once
 
 #include <unifex/config.hpp>
--- a/include/unifex/into_variant.hpp
+++ b/include/unifex/into_variant.hpp
@@ -1,0 +1,136 @@
+#pragma once
+
+#include <unifex/config.hpp>
+#include <unifex/async_trace.hpp>
+#include <unifex/bind_back.hpp>
+#include <unifex/receiver_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
+#include <unifex/type_list.hpp>
+
+#include <tuple>
+#include <variant>
+
+#include <unifex/detail/prologue.hpp>
+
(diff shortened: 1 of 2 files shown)


include/unifex/into_variant.hpp
```diff
@@ -0,0 +1,136 @@
...
+template <typename Receiver>
+struct _receiver<Receiver>::type {
+  UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+
+  template(typename... Values)(
+  requires receiver_of< Receiver, Values...>)
+  void set_value(Values&&... values) && {
+    unifex::set_value((Receiver &&)(receiver_),
+        std::variant<std::tuple<remove_cvref_t<Values>...>>(
+        std::make_tuple((Values &&)(values)...)));
```
ericniebler, line 33: This isn't quite right. This sends a variant with only a single alternative. The predecessor sender may invoke `set_value` with one of a set of argument types, and this is only one of those. The variant type you compute below in your sender, namely `sender_value_types_t<Predecessor, std::variant, std::tuple>`, is the type of the variant this should be forwarding to the nested receiver. Your sender with have to parameterize this receiver on that type.

ericniebler (changes requested):
On the whole, all the pieces are here and this looks very good. Thanks! Your receiver is forwarding the wrong type to the inner receiver. That's the only real problem.

My only other request is that you look more closely at how the rest of libunifex is formatted and match the style, please.

RishabhRD:
Hello @ericniebler , thanks for the comment. I have parameterized the receiver for the variant type as you suggested. I may have gotten it wrong possibly... but I believe you were talking about parameterising receiver class by template. If not please correct me.

Regarding formatting, I have updated the formatting and tried to match the formatting of file ``then.hpp`` as it had similar contents.

ericniebler:
All correct now, and the formatting looks good too. Thanks.

I notice that we're missing a test of the case where a sender can complete with one of a set of types. You can make such a sender with `just(42) | let_done([]{return just("hello");})`.

test/into_variant_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+TEST(IntoVariant, OneOfPossibleValues){
+  bool called = false;
+  auto x = just(42)
+    | let_done([&]{ called = true; return just(42.5); }) 
+    | into_variant()
+    | sync_wait();
+  EXPECT_FALSE(called);
+  EXPECT_TRUE(x.has_value());
+  const auto& [val] = std::get<0>(x.value());
+  EXPECT_EQ(val, 42);
```
RishabhRD, line 110: I wanted to add ``EXPECT_ANY_THROWS(std::get<1>(x.value()));`` to signify that there was only one value in inner variant. However, this breaks the msvc build as [[nodiscard]] warning was treated as error for ``std::get<1>(x.value())``... Hence I removed that assert....

RishabhRD:
Hello @ericniebler I have added the tests you suggested. I added both static type check and gtest TEST_CASE for that.

ericniebler:
Great. I just noticed one last thing. These two new files need license headers at the top. Just copy from another file and change the Facebook copyright to yours (if you like).

RishabhRD:
Hi @ericniebler I added the license header as you suggested.

PR #351, opened 2021-09-23
Fix compile time bug

MBkkt:
Arch Linux
```bash
$ uname -r
5.10.67-1-lts
$ clang --version
clang version 12.0.1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

https://github.com/torvalds/linux/commit/9ba6a1c06279ce499fcf755d8134d679a1f3b4ed
A bit strange, perhaps due to the fact that I also have a non-LTS kernel, its headers were used instead of the correct ones

### DIFF
--- a/include/unifex/linux/io_uring_context.hpp
+++ b/include/unifex/linux/io_uring_context.hpp
@@ -33,9 +33,11 @@
 #include <unifex/linux/monotonic_clock.hpp>
 #include <unifex/linux/safe_file_descriptor.hpp>
 
+#include <algorithm>
 #include <atomic>
 #include <cstddef>
 #include <cstdint>
+#include <iterator>
 #include <optional>
 #include <system_error>
 #include <utility>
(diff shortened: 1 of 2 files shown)


include/unifex/linux/io_uring_context.hpp
```diff
@@ -439,7 +441,7 @@ class io_uring_context::read_sender {
         sqe.rw_flags = 0;
         sqe.user_data = reinterpret_cast<std::uintptr_t>(
             static_cast<completion_base*>(this));
-        sqe.__pad2[0] = sqe.__pad2[1] = sqe.__pad2[2] = 0;
+        std::fill(std::begin(sqe.__pad2), std::end(sqe.__pad2), 0);
```
lewissbaker, line 444: Under the new layout of the structure we would no longer be zeroing out the `buf_index`, `personality`, etc, fields.

Maybe this code should take a different approach and just zero out the entire `io_uring_sqe` struct in `try_submit_io()` and then have the lambdas populate only the fields that they actually need to set for that operation?

PR #352, opened 2021-09-24
add co_file_copy example

kirkshoop:
This is just like the file_copy example, but uses the compilers implementation of C++20 coroutines to compose the senders and manage the lifetimes of variables.

The result is much fewer lines of code and a more intuitive code organization.
The tradeoff is less control over how many allocations there are.

### DIFF
--- a/examples/co_file_copy.cpp
+++ b/examples/co_file_copy.cpp
@@ -1,0 +1,361 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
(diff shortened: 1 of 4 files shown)


kirkshoop:
This should not be merged yet. This is expected to fail CI.
There are changes needed to the build system to support coroutines, with clang on unix at least..
This example manually forces coroutines on and I do not expect that to work.

AnujYamdagni (changes requested):

PR #357, opened 2021-10-08
Implement just_void_or_done combinator

egorich239:


include/unifex/just_void_or_done.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <unifex/config.hpp>
+#include <unifex/blocking.hpp>
+#include <unifex/receiver_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
+#include <unifex/std_concepts.hpp>
+
+#include <exception>
+#include <tuple>
+#include <type_traits>
+#include <utility>
```
ispeters, line 27: As you mentioned in person, you don't need all of these.

include/unifex/just_void_or_done.hpp
```diff
@@ -0,0 +1,96 @@
...
+struct _op {
+  struct type;
+};
+template <typename Receiver>
+using operation = typename _op<remove_cvref_t<Receiver>>::type;
+
+template <typename Receiver>
+struct _op<Receiver>::type {
+  bool is_void_;
+  UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
```
ispeters, line 44: I think you might get a smaller object if you reorder these members.  As far as I understand, `[[no_unique_address]]` allows the compiler to put subsequent members in trailing padding so, given that `sizeof(bool)` and `alignof(bool)` are both one, `is_void_` should fit into `receiver_`'s trailing padding if there is any.
```suggestion
  UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
  bool is_void_;
```

include/unifex/just_void_or_done.hpp
```diff
@@ -0,0 +1,96 @@
...
+  friend constexpr auto tag_invoke(tag_t<blocking>, const _sender&) noexcept {
+    return blocking_kind::always_inline;
+  }
+};
+}  // namespace _just_void_or_done
+
+namespace _just_void_or_done_cpo {
+inline const struct just_void_or_done_fn {
+  constexpr auto operator()(bool is_void) const noexcept {
+    return _just_void_or_done::_sender{is_void};
```
ispeters, line 88: I think the naming conventions here are that locals and members should be `camelCase`.
```suggestion
  constexpr auto operator()(bool isVoid) const noexcept {
    return _just_void_or_done::_sender{isVoid};
```

ispeters (changes requested):
Looks good, just a few nits.

include/unifex/just_void_or_done.hpp
```diff
@@ -0,0 +1,96 @@
...
+  template <template <typename...> class Variant>
+  using error_types = Variant<>;
+
+  static constexpr bool sends_done = true;
+
+  // clang-format off
+  UNIFEX_TEMPLATE(typename This, typename Receiver)
+      (requires same_as<remove_cvref_t<This>, _sender> UNIFEX_AND
+          receiver<Receiver>)
+  friend auto tag_invoke(tag_t<connect>, This&& that, Receiver&& r) noexcept {
```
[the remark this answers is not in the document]
egorich239: yeah, I overloooked this move-ctor, you're right... I'll fix it.

include/unifex/just_void_or_done.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+  static constexpr bool sends_done = true;
+
+  // clang-format off
+  UNIFEX_TEMPLATE(typename This, typename Receiver)
+      (requires same_as<remove_cvref_t<This>, _sender> UNIFEX_AND
+          receiver<Receiver>)
+  friend auto tag_invoke(tag_t<connect>, This&& that, Receiver&& r)
+      noexcept(std::is_nothrow_move_constructible_v<Receiver>) {
+    return operation<Receiver>{static_cast<Receiver&&>(r), that.isVoid_};
```
ispeters, line 73: I don't think you're benefiting from preserving the value category of the passed-in sender here; regardless, you're going to copy the `bool`.  Given that, I'd simplify the declaration to take the sender by value.

```suggestion
  UNIFEX_TEMPLATE(typename Receiver)
      (requires receiver<Receiver>)
  friend auto tag_invoke(tag_t<connect>, _sender s, Receiver&& r)
      noexcept(std::is_nothrow_move_constructible_v<Receiver>) {
    return operation<Receiver>{static_cast<Receiver&&>(r), s.isVoid_};
```

include/unifex/just_void_or_done.hpp
```diff
@@ -0,0 +1,94 @@
...
+  // clang-format on
+
+  friend constexpr auto tag_invoke(tag_t<blocking>, const _sender&) noexcept {
+    return blocking_kind::always_inline;
+  }
+};
+}  // namespace _just_void_or_done
+
+namespace _just_void_or_done_cpo {
+inline const struct just_void_or_done_fn {
```
ispeters, line 84: I think you can use `constexpr` here.

```suggestion
constexpr struct just_void_or_done_fn {
```

include/unifex/just_void_or_done.hpp
egorich239: Indeed, I overlooked that

ispeters:
Looks like the failures in macOS Clang Optimised (C++20) exist at the head of `unstable` as is and you didn't introduce them so I'm going to merge what you've got here.

PR #361, opened 2021-10-15
Add detach on cancel

jesswong:


include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+      }
+    }
+  };
+
+  struct type final {
+    type(UpstreamSender&& s, DownstreamReceiver&& r)
+      : receiver_(std::move(r))
+      , state_(new detached_state(this, (UpstreamSender &&) s)){};
+
+    friend auto tag_invoke(tag_t<unifex::start>, typename operation_state<UpstreamSender, DownstreamReceiver>::type& op) noexcept {
```
lewissbaker, line 41: `start()` always returns `void` so could use `void` return-type here instead of `auto`.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+        op->callback_.destruct();
+        unifex::set_done(std::move(op->receiver_));
+      }
+    }
+  };
+
+  struct type final {
+    type(UpstreamSender&& s, DownstreamReceiver&& r)
+      : receiver_(std::move(r))
+      , state_(new detached_state(this, (UpstreamSender &&) s)){};
```
lewissbaker, line 39: Prefer to use `std::make_unique` instead of raw `operator new`.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+        callback_;
+    std::unique_ptr<detached_state> state_;
+  };
+};
+
+template <typename UpstreamSender, typename DownstreamReceiver>
+struct operation_state<UpstreamSender, DownstreamReceiver>::_receiver final {
+  auto tryGetOp() noexcept {
+    // if cancellation won, parentOp_ would be null at this point
+    // see line24
```
lewissbaker, line 60: nit: Can we refer to a particular function name instead of line number?
line number references are likely to quickly go out of date as code is modified.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+      template <typename...>
+      class Variant,
+      template <typename...>
+      class Tuple>
+  using value_types =
+      typename sender_traits<Sender>::template value_types<Variant, Tuple>;
+
+  template <template <typename...> class Variant>
+  using error_types =
+      typename sender_traits<Sender>::template error_types<Variant>;
```
lewissbaker, line 135: I think you also need to make sure that `std::exception_ptr` is added to the list of errors since you are catching exceptions and calling `set_error()` with `std::current_exception()`.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+  struct detached_state;
+
+  struct cancel_callback final {
+    detached_state* state_;
+
+    void operator()() noexcept {
+      auto* op = state_->parentOp_.exchange(nullptr);
+      if (op != nullptr) {
+        // We won the race for cancellation
+        op->state_->stopSource_.request_stop();
```
lewissbaker, line 27: I think this call may need to be performed before the call to `state_->parentOp_->exchange()` above.
As soon as you have exchanged the parentOp with null it's possible that the operation may complete concurrently on another thread, see that parentOp is now null and then destroy the detached_state object (which means this call might be made on a dangling pointer).

If you move it before the call to `exchange()` above then if the operation completes concurrently on another thread then if it sees a null parentOp it will be safe to destroy, or if it sees a non-null parentOp it will then call .destruct() on the callback which will block waiting for the cancel callback to complete (and thus the call to `stopSource_.request_stop()` will have returned and won't be operating on a dangling object).

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+struct operation_state {
+  struct _receiver;
+
+  struct detached_state;
+
+  struct cancel_callback final {
+    detached_state* state_;
+
+    void operator()() noexcept {
+      auto* op = state_->parentOp_.exchange(nullptr);
```
lewissbaker, line 24: Use `std::memory_order_acq_rel` here.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+    std::unique_ptr<detached_state> state_;
+  };
+};
+
+template <typename UpstreamSender, typename DownstreamReceiver>
+struct operation_state<UpstreamSender, DownstreamReceiver>::_receiver final {
+  auto tryGetOp() noexcept {
+    // if cancellation won, parentOp_ would be null at this point
+    // see line24
+    auto* op = state_->parentOp_.exchange(nullptr);
```
lewissbaker, line 61: Use `std::memory_order_acq_rel` here.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+  template <template <typename...> class Variant>
+  using error_types =
+      typename sender_traits<Sender>::template error_types<Variant>;
+
+  static constexpr bool sends_done = true;
+
+  friend constexpr auto
+  tag_invoke(tag_t<blocking>, const type& sender) noexcept {
+    return blocking(sender.upstreamSender_);
+  }
```
lewissbaker, line 142: I think this can potentially complete inline in the case that stop has been requested.
So if `blocking(upstreamSender)` returns `never` then you should return `maybe`, otherwise you can just return `blocking(upstreamSender)`.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+  template(typename This, typename Receiver)(requires same_as<remove_cvref_t<This>, type> AND
+  receiver<Receiver>)
+  friend auto tag_invoke(
```
lewissbaker, line 146: A possible future improvement here would be to specialise this for the case where the receiver's stop-token is unstoppable, in which case you could just call through to `connect(upstreamSender, receiver)` since a stop-request will never be made and so you don't need the overhead of the detachable_state.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+  }
+};
+}  // namespace _detach_on_cancel
+
+namespace detach_on_cancel_impl {
+inline constexpr struct detach_on_cancel_fn {
+  template(typename Sender)(requires sender<Sender>) constexpr auto
+  operator()(Sender&& sender) const noexcept(
+      is_nothrow_constructible_v<_detach_on_cancel::sender<remove_cvref_t<Sender>>, Sender>)
+      -> _detach_on_cancel::sender<Sender> {
```
lewissbaker, line 161: Return type here doesn't match the actual type you're returning.
Should be `_detach_on_cancel::sender<remove_cvref_t<Sender>>`

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  template(typename This, typename Receiver)(requires same_as<remove_cvref_t<This>, type> AND
+  receiver<Receiver>)
+  friend auto tag_invoke(
+      tag_t<unifex::connect>,
+      This&& s,
+      Receiver&&
+          r) noexcept(false) {
+    return typename operation_state<Sender, Receiver>::type{((This &&) s).upstreamSender_, (Receiver &&) r};
```
lewissbaker, line 151: I think this should be `operation_state<Sender, remove_cvref_t<Receiver>>`.
Or alternatively, the data-member in the operation-state should be:
```
remove_cvref_t<DownstreamReceiver> receiver_;
```

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+  struct detached_state;
+
+  struct cancel_callback final {
+    detached_state* state_;
+
+    void operator()() noexcept {
+      if (state_ != nullptr) {
+        state_->stopSource_.request_stop();
+      }
```
ispeters, line 43: I don't see a way for `state_` to ever be `nullptr`.
```suggestion
      state_->stopSource_.request_stop();
```

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+  static constexpr bool sends_done = true;
+
+  friend constexpr auto
+  tag_invoke(tag_t<blocking>, const type& sender) noexcept {
+    auto block_value = blocking(sender.upstream_sender_);
+    if (block_value == _block::_enum::never) {
+      return _block::_enum::maybe;
+    }
+    return block_value;
```
ispeters, line 176: This isn't _quite_ right, because `blocking(sender.upstream_sender_)` might return an instance of `std::integral_constant` rather than an instance of `unifex::blocking`.  I think you need something like this:
```suggestion
    if constexpr (same_as<blocking_kind, decltype(blocking(sender.upstream_sender_))>) {
      // the sender returns a runtime-determined blocking_kind
      blocking_kind blockValue = blocking(sender.upstream_sender_);
      if (blockValue == blocking_kind::never) {
        blockValue = blocking_kind::maybe;
      }
      return blockValue;
    }
    else if constexpr (blocking_kind::never == cblocking<Sender>()) {
      // the sender always returns never
      return blocking_kind::maybe;
    }
    else {
      return cblocking<Sender>();
    }      
```
You may need @ericniebler's eyes on this one.

ispeters (commented):
Looking good; just a couple of nits from me.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,212 @@
...
+        op->callback_.destruct();
+        unifex::set_done(std::move(op->receiver_));
+      }
+    }
+  };
+
+  struct type final {
+    type(UpstreamSender&& s, DownstreamReceiver&& r)
+      : receiver_(std::move(r))
+      , state_(std::make_unique<detached_state>(this, (UpstreamSender &&) s)){};
```
lewissbaker, line 56: Semicolon not needed at end of line here.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+    friend void tag_invoke(
+        tag_t<unifex::start>,
+        typename operation_state<UpstreamSender, DownstreamReceiver>::type&
```
lewissbaker, line 60: Should be able to simplify parameter type to just `type&`.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,212 @@
...
+  detached_state* state_;
+};
+
+template <typename UpstreamSender, typename DownstreamReceiver>
+struct operation_state<UpstreamSender, DownstreamReceiver>::detached_state
+    final {
+  detached_state(
+      typename operation_state<UpstreamSender, DownstreamReceiver>::type* op,
+      UpstreamSender&&
+          s) noexcept(is_nothrow_connectable_v<UpstreamSender, _receiver>)
```
lewissbaker, line 128: Some weird formatting that breaks a single-character parameter name onto a separate line here.
clang-format's just being weird?

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,212 @@
...
+    : parentOp_(op)
+    , childOp_(connect(std::move(s), _receiver{this})){};
```
lewissbaker, line 130: Semicolon not necessary after constructor function body.

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,212 @@
...
+      // the sender always returns never
+      return blocking_kind::maybe;
+    } else {
+      return cblocking<Sender>();
+    }
+  }
+
+  template(typename This, typename Receiver)(
+      requires same_as<remove_cvref_t<This>, type> AND receiver<
+          Receiver>) friend auto tag_invoke(tag_t<unifex::connect>, This&& s, Receiver&& r) noexcept(false) {
```
lewissbaker, line 189: You can get slightly better clang-formatting of the template/requires macros by putting an empty comment at the end of a line to force clang-format to insert a line-break there.
e.g.
```
template(typename This, typename Receiver) //
  (requires same_as<remove_cvref_t<This>, type> AND receiver<Receiver) //
  friend auto tag_invoke(....)
```

include/unifex/detach_on_cancel.hpp
```diff
@@ -0,0 +1,212 @@
...
+    } else if constexpr (blocking_kind::never == cblocking<Sender>()) {
+      // the sender always returns never
+      return blocking_kind::maybe;
+    } else {
+      return cblocking<Sender>();
+    }
+  }
+
+  template(typename This, typename Receiver)(
+      requires same_as<remove_cvref_t<This>, type> AND receiver<
```
lewissbaker, line 188: We should also consider constraining this method to require that `member_t<This, Sender>` is connectable to `operation_state<...>::_receiver` using the `sender_to` concept check.

PR #363, opened 2021-10-21
Add done_as_optional per P2300

ericniebler:


include/unifex/done_as_optional.hpp
```diff
@@ -0,0 +1,58 @@
...
+ */
+#pragma once
+
+#include <optional>
+
+#include <unifex/config.hpp>
+#include <unifex/then.hpp>
+#include <unifex/let_done.hpp>
+#include <unifex/just.hpp>
+#include <unifex/sender_concepts.hpp>
```
lewissbaker, line 24: Missing include of `<unifex/type_traits.hpp>` for `non_void_t`

include/unifex/done_as_optional.hpp
```diff
@@ -0,0 +1,58 @@
+/*
```
lewissbaker, line 1: Add docs to `doc/api_reference.md`?

include/unifex/sender_concepts.hpp
```diff
@@ -356,6 +356,30 @@ template <typename Sender>
 constexpr bool is_sender_nofail_v =
     sender_error_types_t<Sender, is_empty_list>::value;
 
-} // namespace unifex
+/// \cond
+namespace _single_sender {
+template <typename... Types>
+using _is_single_valued_tuple = std::bool_constant<1 >= sizeof...(Types)>;
```
lewissbaker, line 362: Does `std::bool_constant<1 >= sizeof...(Types)>` need parentheses around bool expression to avoid it being parsed incorrectly?

include/unifex/done_as_optional.hpp
ericniebler: OK, but it'll be quick and dirty.

PR #369, opened 2021-11-04
Make sync_wait a proper CPO.

egorich239:
The `sync_wait` CPO by default defers to `sync_wait_r</*deduced-type*/>` which matches the preexisting behavior. Overriding `sync_wait_r<T>` hence by default is enough to consistently customize the behavior of both algorithms.

As an aside I matched piped API of sync_wait and sync_wait_r.

### DIFF
--- a/include/unifex/sync_wait.hpp
+++ b/include/unifex/sync_wait.hpp
@@ -140,32 +140,23 @@ std::optional<Result> _impl(Sender&& sender) {
 }
 } // namespace _sync_wait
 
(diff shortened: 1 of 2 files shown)


AnujYamdagni:
We're doing repo cleanup - Can you add more information in your summary about the motivation behind this PR?

LeeHowes:
The POR is that we don't want pipe syntax for sync_wait, so that aspect of this we shouldn't merge.

PR #372, opened 2021-11-05
Make unifex::async_scope::spawn return a unifex::future<>

ispeters:
This diff changes `unifex::async_scope::spawn(Sender auto)` to return a
new type, `unifex::future<...>`.  `future<>` is a _Sender_ that completes
with the result of the _Sender_ given to `spawn`.


ispeters:
Squashing and merging so we can start using this internally; I'll re-up this PR as a PR against `main` and we can do a proper review there.

PR #373, opened 2021-11-07
Fix compile issue related to io_uring

adevress:
Fix a compilation problem when using liburing support related to a change of layout in a io_uring struct (remove padding). Adapted the usage accordingly.

### DIFF
--- a/.github/workflows/libunifex-ci.yml
+++ b/.github/workflows/libunifex-ci.yml
@@ -83,6 +83,22 @@ jobs:
             cmake_args: "-D CMAKE_CXX_STANDARD:STRING=20 -D CMAKE_CXX_FLAGS:STRING=-fcoroutines"
           }
         - {
+            name: "Linux GCC 10 Debug (C++20) w/coroutines w/liburing", artifact: "Linux.tar.xz",
+            os: ubuntu-latest,
+            build_type: Debug,
+            io_sys: io_uring,            
+            cc: "gcc-10", cxx: "g++-10",
+            cmake_args: "-D CMAKE_CXX_STANDARD:STRING=20 -D CMAKE_CXX_FLAGS:STRING=-fcoroutines"
+          }
+        - {
+            name: "Linux GCC 10 Optimised (C++20) w/coroutines w/liburing", artifact: "Linux.tar.xz",
+            os: ubuntu-latest,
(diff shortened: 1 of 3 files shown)


ispeters:
This seems like a good idea, and the change looks like it does what you intend, but I don't really understand the consequences outside the library.

Does this cause backwards-compatibility concerns depending on the version of Linux in use?

williamspatrick:
> Does this cause backwards-compatibility concerns depending on the version of Linux in use?

I don't believe it would cause any backwards-compatibility issues.  It is on io_uring to maintain forward and backwards compatibility at a kernel ABI level.  The io_uring library itself can't know at compile time which kernel version it will be running against.  They have done this by having these unused "pad" words in the structure and as they add additional options / features they start to consume the pad.

The change proposed here does not add or remove any initialization but instead switches to "initialize the whole structure to zero" model of initialization.  This is actually *more* forward / backwards compatible because we're now ensuring everything is set to zero even if it was unused in a previous version of the io_uring API/ABI.

Again, it is on io_uring to ensure the ABI is stable at a kernel level, so I don't think there is anything for us to worry about here.  If they change the ABI enough to use all the padding up and need more, they'll have to introduce a new structure (and likely API) to accommodate.

.github/workflows/libunifex-ci.yml
```diff
@@ -82,6 +82,22 @@ jobs:
             cc: "gcc-10", cxx: "g++-10",
             cmake_args: "-D CMAKE_CXX_STANDARD:STRING=20 -D CMAKE_CXX_FLAGS:STRING=-fcoroutines"
           }
+        - {
+            name: "Linux GCC 10 Debug (C++20) w/coroutines w/liburing", artifact: "Linux.tar.xz",
```
williamspatrick, line 86: I do wonder how many various combinations here we realistically need to support for CI.  I'd be fine if we condense this down into a single `w/coroutines` mode.

Might as well get this commit in as-is though to resolve the compile issues.

williamspatrick (approved):
LGTM.  I'm running into this compile issue as well, so I'd appreciate it being merged!

ispeters (approved):
Looks good and my compatibility concerns have been addressed so let's stamp this.

PR #375, opened 2021-11-12
fix non-virtual destructor warnings

williamspatrick:
When compiling the coroutine_stream_consumer example with
-Werror, I observed errors such as:

```
    ... has accessible non-virtual destructor [-Werror=non-virtual-dtor]
    220 |         struct concrete_receiver final : next_receiver_base {
```

Fix these by adding default virtual destructors to all base structures
which define virtual functions.

Signed-off-by: Patrick Williams <patrick@stwcx.xyz>

### DIFF
--- a/include/unifex/stop_immediately.hpp
+++ b/include/unifex/stop_immediately.hpp
@@ -179,6 +179,11 @@ struct _stream<SourceStream, Values...>::type {
         }
       }
 
+      if (oldState == state::cleanup_completed) {
+          // Cleanup was already completed elsewhere.
+          return;
+      }
+
       // Otherwise, cleanup() was requested before this operation completed.
       // We are responsible for starting cleanup now that next() has finished.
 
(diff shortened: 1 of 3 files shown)


ericniebler:
A better fix would be to eliminate all uses of `virtual` in libunifex.

williamspatrick:
> A better fix would be to eliminate all uses of `virtual` in libunifex.

Can you give me some pointers on how that would be done?

You'll see that I just pushed up another commit to this PR which includes what looks like a terrible hack.  When I added the proper virtual destructors, it aggravated failures in GCC in anything using `stop_immediately` because the destructor on the `cleanup_operation_base` ends up being called twice.

I think if we were to eliminate `virtual` here it would at least make GCC happier w.r.t. missing virtual function calls, but we'd still have the behavior of calling the destructor twice.  Isn't that UB?

williamspatrick:
> > A better fix would be to eliminate all uses of `virtual` in libunifex.
> 
> Can you give me some pointers on how that would be done?

I’m guessing the answer here is something involving tag_invoke.  I’ve read through P1895 and the implementations of `any_unique`/`any_ref` here.  What I don’t understand yet is how this can be leveraged to remove the virtual base class `cleanup_operation_base`.  

The primary usage of this is as a pointer that is updated to point to a cleanup operation depending on state.  Value semantics don’t work for this as the cleanup operation is already held elsewhere as well and any_ref can’t hold a `nullptr`.  Do we need an `any_ptr`?

ericniebler:
In most of the library, runtime polymorphic behavior is implemented with function pointers. It's more work, but the added benefit is that it doesn't generate RTTI that bloats the binary.

williamspatrick:
I'm going to close this PR and open a separate one which starts on the removal of virtual.

PR #377, opened 2021-11-16
remove virtual base classes

williamspatrick:
- vtable: non-virtual runtime polymorphism support
- take_until: remove virtual base class
- vtable: support r-value qualified members
- stop_immediately: remove virtual base classes

### DIFF
--- a/include/unifex/vtable.hpp
+++ b/include/unifex/vtable.hpp
@@ -32,9 +32,22 @@
   }                                                     \
   entry##_t entry##_ = nullptr
 
+#define UNIFEX_VTABLE_ENTRY_RVALUE(entry, ret, /*types*/...) \
+  using entry##_t = ret (*)(void*, __VA_ARGS__);             \
+  template <typename... Args>                                \
+  ret entry(Args&&... args)&& {                              \
+    return entry##_(ctx_, std::forward<Args>(args)...);      \
+  }                                                          \
+  entry##_t entry##_ = nullptr
+
 #define UNIFEX_VTABLE_ENTRY_VOID(entry, ret) \
   using entry##_t = ret (*)(void*);          \
   ret entry() { return entry##_(ctx_); };    \
+  entry##_t entry##_ = nullptr
+
+#define UNIFEX_VTABLE_ENTRY_VOID_RVALUE(entry, ret) \
+  using entry##_t = ret (*)(void*);                 \
+  ret entry()&& { return entry##_(ctx_); };         \
   entry##_t entry##_ = nullptr
 
 #define UNIFEX_VTABLE_CONSTRUCT_FN(fn) \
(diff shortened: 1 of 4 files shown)


williamspatrick:
@ericniebler As you mentioned in #375
> A better fix would be to eliminate all uses of virtual in libunifex.

What do you think about this approach?  There are a few more classes to clean up.

test/vtable_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+  void foo() && {};
+  void bar(int) && {};
+
+  vtable table_ = UNIFEX_VTABLE_CONSTRUCT(&class_four::foo, &class_four::bar);
+};
+
+TEST(vtable, ref_qualifier) {
+  auto instance = std::make_unique<class_four>();
+  std::move(instance->table_).foo();
```
egorich239, line 127: Does it have to be `std::move()`? Can it be called on a (const?) reference? The vtable state does not change as a result of the call.

Besides, current design seems to imply that a vtable is always assigned to `this` object, and the table is a part of the class definition. Is such limitation desired? Can I define a vtable for `class_four` outside of `class_four`?

egorich239 (commented):
I am not sure if the questions below have been discussed elsewhere. Please excuse me if they have.

test/vtable_test.cpp
williamspatrick: > Does it have to be std::move()? Can it be called on a (const?) reference?

This code was attempting to show that ref-qualified functions can also be added to a vtable.  I'm pretty sure you can call on any r-value reference'd instance.  I could add an example on a local temporary here as well.

> The vtable state does not change as a result of the call.

The vtable is really just an indirection to the object itself.  In this case the object is needing to be called with r-value semantics (due to the ref-qualifier on the function), so the vtable does as well.  Otherwise the vtable code would be silently turning `this` into an r-value ref, which is dangerous.

> Besides, current design seems to imply that a vtable is always assigned to this object, and the table is a part of the class definition. Is such limitation desired? Can I define a vtable for class_four outside of class_four?

The vtable definition is allowed to be outside the class, and would be in most cases.  See lines 23 and 78.  `class_four` is the only example where I did it _inside_ the class, partially to show that it could be done there as well and partially for the implementation of the ref-qualifier test to be smaller.

The vtable instance _must_ be in the class (using the UNIFEX_VTABLE_CONSTRUCT call) just like a normal vtable would be, but the definition can and most likely will be outside the class.  This is similar to a "base class" with normal virtualization.

include/unifex/vtable.hpp
```diff
@@ -0,0 +1,99 @@
...
+  using entry##_t = ret (*)(void*);                 \
+  ret entry()&& { return entry##_(ctx_); };         \
+  entry##_t entry##_ = nullptr
+
+#define UNIFEX_VTABLE_CONSTRUCT_FN(fn) \
+  (unifex::_vtable::                   \
+       _construct_indirect<std::remove_reference_t<decltype(*this)>, fn>(fn)),
+
+#define UNIFEX_VTABLE_CONSTRUCT(/*functions*/...) \
+  { this, UNIFEX_PP_FOR_EACH(UNIFEX_VTABLE_CONSTRUCT_FN, __VA_ARGS__) }
```
williamspatrick, line 58: Using `this` causes stale reference issues when the underlying object is copied.  The context ends up pointing at the original instance and not the current location.

I might be able to eliminate this by using something like `offsetof` to find the offset of the vtable within the class.  Need to revisit.

lewissbaker:
Note that the PR #304 also does some refactoring to make the vtable facilities used by `any_unique` more reusable in other type-erasing wrappers.

For the use-cases here I wonder if we could make use of `any_ref` to type-erase the concrete receivers instead of virtual base classes?

AnujYamdagni (changes requested):
We're doing repo cleanup - are you still interested in merging this or shall we close it?

PR #255, opened 2021-04-13
Add an install target for the library

maikel:
This PR enables the library to be installed. 

It is still work on progress and should *NOT* be merged.  It needs testing or improvements for

 - [x] Makefile generator
 - [ ] Ninja generator
 - [ ] Finding and linking liburing when using this install target
 - [ ] a GitHub Action

If you have more wishes, feel free to comment.

PR #383, opened 2021-12-02
Add install target for library

williamspatrick:
Enable an install target for unifex and a pkgconfig file so that unifex
can be used by non-CMake based repositories.

- Add an install target for the library
- Add library versioning
- Add pkgconfig file
- ci: add install target testing

### DIFF
--- a/.github/workflows/libunifex-ci.yml
+++ b/.github/workflows/libunifex-ci.yml
@@ -417,3 +417,35 @@ jobs:
         if (NOT result EQUAL 0)
           message(FATAL_ERROR "Running tests failed!")
         endif()
+
+    - name: Install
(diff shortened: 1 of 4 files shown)


williamspatrick:
This is a continuation / completion of #255.  I have completed all the outstanding tasks listed in that older PR and tested integration with a meson-based repository (openbmc/sdbusplus).  This will also allow me to submit a Yocto recipe for unifex so it can be included in that Linux distribution.

williamspatrick:
Looks like the CI workflow is successful now.  Took me a few iterations to get the correct cmake invocation in the CI environment working.  PR is ready for review.

ispeters (approved):
LGTM

PR #384, opened 2021-12-07
Fix logic bug that determines when to disable inlining for await_suspend.

mnbannister:
The clang compilers on windows define _MSC_VER so will inline await_suspend which, as per the comment for the method, cannot be tolerated. The new logic enables inlining only for MSVC and clang11+.

### DIFF
--- a/include/unifex/at_coroutine_exit.hpp
+++ b/include/unifex/at_coroutine_exit.hpp
@@ -61,7 +61,7 @@ struct _cleanup_promise_base {
     template <typename CleanupPromise>
     // Apple-clang and clang-10 and prior need for await_suspend to be noinline.
     // MSVC and clang-11 can tolerate await_suspend to be inlined, so force it.
-#if defined(_MSC_VER) || !(defined(__apple_build_version__) || __clang_major__ < 11)
+#if defined(_MSC_VER) && !(defined(__apple_build_version__) || __clang_major__ < 11)
     UNIFEX_ALWAYS_INLINE
 #else
     UNIFEX_NO_INLINE
--- a/include/unifex/at_coroutine_exit.hpp
+++ b/include/unifex/at_coroutine_exit.hpp
@@ -61,10 +61,10 @@ struct _cleanup_promise_base {
     template <typename CleanupPromise>
     // Apple-clang and clang-10 and prior need for await_suspend to be noinline.
     // MSVC and clang-11 can tolerate await_suspend to be inlined, so force it.
-#if defined(_MSC_VER) && !(defined(__apple_build_version__) || __clang_major__ < 11)
-    UNIFEX_ALWAYS_INLINE
-#else
+#if defined(__apple_build_version__) || (defined(__clang__) && __clang_major__ <= 10)
     UNIFEX_NO_INLINE
+#else
+    UNIFEX_ALWAYS_INLINE
 #endif
     void await_suspend(coro::coroutine_handle<CleanupPromise> h) const noexcept {
       auto continuation = h.promise().next();


include/unifex/at_coroutine_exit.hpp
```diff
@@ -61,7 +61,7 @@ struct _cleanup_promise_base {
     template <typename CleanupPromise>
     // Apple-clang and clang-10 and prior need for await_suspend to be noinline.
     // MSVC and clang-11 can tolerate await_suspend to be inlined, so force it.
-#if defined(_MSC_VER) || !(defined(__apple_build_version__) || __clang_major__ < 11)
+#if defined(_MSC_VER) && !(defined(__apple_build_version__) || __clang_major__ < 11)
     UNIFEX_ALWAYS_INLINE
 #else
     UNIFEX_NO_INLINE
```
ispeters, line 67: I think the intent of this PR is to handle the confusion created by Clang for Windows defining `_MSC_VER`, and I don't think this change does that.  The logic is a bit confusing to me, though, so here's my reasoning:
 - applying DeMorgan's law, we can distribute the `!` over the `||` to get `defined(_MSC_VER) && !defined(__apple_build_version__) && !(__clang_major__ < 11)`
 - I'm pretty sure that `_MSC_VER` and `__apple_build_version__` are never both defined so, assuming `_MSC_VER` _is_ defined, `__apple_build_version__` is _not_ defined, and we can remove it from the conjunction
 - simplifying `!a < b` to `a >= b` leaves `defined(_MSC_VER) && __clang_major__ >= 11`
 - we're in a context where `__clang__` is defined and `__clang_major__ < 12` so we can simplify `>= 11` to `== 11`

Assuming my reasoning is correct, then this change restricts the force-inline behaviour to Clang 11 for Windows, which isn't what we want.

Based on the comment on lines 62-63, I think we want these conditions:
 - Apple Clang? force outline
 - Clang 10 or earlier? force outline
 - MSVC? force inline
 - Clang 11 or later, but not Apple Clang? force inline

I think we get "MSVC?" with `defined(_MSC_VER) && !defined(__clang__)`; I think we get "non-Apple Clang 11 or later?" with `!defined(__apple_build_version__) && __clang_major__ >= 11`.  Combining them, I think we need this:
```suggestion
#if (defined(_MSC_VER) && !defined(__clang__)) || (!defined(__apple_build_version__) && __clang_major__ >= 11)
    UNIFEX_ALWAYS_INLINE
#else
    UNIFEX_NO_INLINE
```

Alternatively, we could translate the comment more directly:
```suggestion
#if defined(__apple_build_version__) || (defined(__clang__) && __clang_major__ <= 10)
    UNIFEX_NO_INLINE
#else
    UNIFEX_ALWAYS_INLINE
```

What do you think?

ispeters (changes requested):
I sometimes get tripped up by complicated Boolean logic so please check my reasoning in the inline comment.

include/unifex/at_coroutine_exit.hpp
mnbannister: Yes, I believe you're absolutely correct. The proposed change does not consider the case of clang 11 not on windows. Of the two suggestions I prefer the latter as it's concise. Let me get a pull request together for that.

mnbannister: I'm not sure how much you care about this at this point but I've just noticed that the logic just above this captures clang 12 on windows, which means it won't use symmetric transfer with clang 12 on windows. This could be considered benign (i.e. at least it won't crash) but might be something you want to capture in a task.

ispeters: I think you're talking about this code:
```
    // Clang before clang-12 has a bug with coroutines that self-destruct in an
    // await_suspend that uses symmetric transfer. It appears that MSVC has the same
    // bug. So instead of symmetric transfer, we accept the stack growth and resume
    // the continuation from within await_suspend.
#if (defined(__clang__) && (defined(__apple_build_version__) || __clang_major__ < 12)) || \
    defined(_MSC_VER)
```
And, now that I can see it next your comment, I agree.  Clang 12 on Windows will have the first clause of the main disjunction `false` because `__clang__` will be defined, but `__apple_build_version__` will be undefined and  `__clang_major__` will not be less than 12, but the second clause will be `true` because `_MSC_VER` will be defined.

Yeah, let's file an issue to get that corrected.

ispeters (approved):
Thanks for the fix!

PR #386, opened 2021-12-13
delete move operators from inplace_stop_token

janondrusek:
- copy constructor makes a trivial copy of a pointer
- avoid leaving source object in an empty state


ispeters (approved):
Thanks!

PR #388, opened 2021-12-23
preserve operation state in let_value_with_stop_token

janondrusek:
* SuccessorFactory should live past `unifex::connect`
* SuccessorFactory may be destroyed after `unifex::start`

### DIFF
--- a/test/let_value_with_stop_token_test.cpp
+++ b/test/let_value_with_stop_token_test.cpp
@@ -72,7 +72,33 @@ struct NonInplaceStoppableIntReceiver : public UnstoppableSimpleIntReceiver {
   inplace_stop_source& source_;
 };
 
-TEST(LetWithStopToken, Simple) {
+struct LetWithStopToken : testing::Test {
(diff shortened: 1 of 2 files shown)


test/let_value_with_stop_token_test.cpp
```diff
@@ -72,7 +72,33 @@ struct NonInplaceStoppableIntReceiver : public UnstoppableSimpleIntReceiver {
...
+    }
+
+    ~DestructionCounter() {
+      if (!wasMoved) {
+        destroyCount()++;
+      }
+    }
+  };
+
+  void SetUp() override { DestructionCounter::destroyCount() = 0; }
```
ispeters, line 98: For future, I'd do it this way.  I'm just gonna merge your diff, though.

```suggestion
  LetWithStopToken() noexcept { DestructionCounter::destroyCount() = 0; }
```

ispeters (approved):
Probably should've approved before merging but whatever.

PR #391, opened 2022-02-03
Minimal changes to allow compilation to succeed on gcc and clang on MSYS2

bustercopley:
Fix typos, add missing #includes, avoid redefining macros, suppress warnings. Work around some gcc quirks.

### DIFF
--- a/.github/workflows/libunifex-ci.yml
+++ b/.github/workflows/libunifex-ci.yml
@@ -238,7 +238,7 @@ jobs:
           }
         - {
             name: "Windows MSVC 2019 Debug (C++17)", artifact: "Windows-MSVC.tar.xz",
-            os: windows-latest,
+            os: windows-2019,
             build_type: Debug,
             cc: "cl", cxx: "cl",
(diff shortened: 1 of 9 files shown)


ispeters:
Thanks for submitting this PR and apologies for the delay in reviewing it.  It's disconcerting to me that the Windows builds are all broken—I don't see how that could be introduced by this change, which seems like we ought to merge unmodified, but I'd like to figure out the cause before merging.

bustercopley:
Thanks. No rush! I am relying on the CI to test this on MSVC. The (first?) problem seems to be that the CI cscript `.github\workflows\libunifex-ci.yml` specifies `windows-latest` (now [updated](https://github.com/actions/virtual-environments/issues/4856) to MSVC 2022 in Github Actions) but then specifies the `PATH` to MSVC 2019. Suggest changing that to `windows-2019`.

ispeters (approved):
Thanks for the changes to support MSYS2 and for fixing our CI as a bonus!

PR #392, opened 2022-02-03
add async_scope::attach

janondrusek:
* returned `Sender` needs to be connected and started
* avoids paying penalty of `future<>`

### DIFF
--- a/include/unifex/async_scope.hpp
+++ b/include/unifex/async_scope.hpp
@@ -17,6 +17,7 @@
 
 #include <unifex/config.hpp>
 #include <unifex/async_manual_reset_event.hpp>
+#include <unifex/fused_stop_source.hpp>
 #include <unifex/get_stop_token.hpp>
 #include <unifex/inplace_stop_token.hpp>
 #include <unifex/just_from.hpp>
@@ -44,6 +45,8 @@ namespace _async_scope {
 struct async_scope;
 
 void record_done(async_scope*) noexcept;
+
+[[nodiscard]] bool try_record_start(async_scope* scope) noexcept;
 
 inplace_stop_token get_stop_token_from_scope(async_scope*) noexcept;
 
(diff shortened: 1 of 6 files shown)


include/unifex/async_scope.hpp
```diff
@@ -561,6 +564,201 @@ struct future final {
...
+
+  void set_done() noexcept {
+    op_->cleanup();
+    unifex::set_done(std::move(receiver_));
+  }
+
+  template(typename CPO)                       //
+      (requires is_receiver_query_cpo_v<CPO>)  //
+      friend auto tag_invoke(CPO cpo, const type& r) noexcept(
+          noexcept(static_cast<CPO&&>(cpo)(r.receiver_)))
```
ispeters, line 598: I wonder if there's a `is_nothrow_callable_v` somewhere.

include/unifex/async_scope.hpp
```diff
@@ -561,6 +564,201 @@ struct future final {
...
+          -> callable_result_t<CPO, const Receiver&> {
+    return static_cast<CPO&&>(cpo)(r.receiver_);
+  }
+
+  friend inplace_stop_token
+  tag_invoke(tag_t<get_stop_token>, const type& r) noexcept {
+    return r.op_->get_token();
+  }
+  Operation* op_;
```
ispeters, line 607: Nit: whitespace.
```suggestion

  Operation* op_;
```

include/unifex/async_scope.hpp
```diff
@@ -561,6 +564,201 @@ struct future final {
...
+    } else {
+      unifex::deactivate_union_member(receiver_);
+    }
+  }
+
+  auto get_token() { return stopSource_.get_token(); }
+
+  void cleanup() noexcept {
+    UNIFEX_ASSERT(scope_ref() != nullptr);
+    stopSource_.deregister_callbacks();
```
ispeters, line 661: Should we do this unconditionally?  I _think_ it's OK because it's invoking `optional<>::reset()` under the covers, which is safe if the optional's already empty.

include/unifex/async_scope.hpp
```diff
@@ -561,6 +564,201 @@ struct future final {
...
+          op.receiverToken_, op.scope_ref()->get_stop_token());
+      unifex::start(op.op_.get());
+    } else {
+      unifex::set_done(std::move(op).receiver_.get());
+    }
+  }
+
+private:
+  static constexpr std::uintptr_t mask = ~(std::uintptr_t{1u});
+  // async_scope* stored as integer, so LSB can be used to signal start
```
ispeters, line 678: Perhaps we should have `static_assert(alignof(async_scope) > 1)` to guarantee the LSB is available?

include/unifex/async_scope.hpp
```diff
@@ -561,6 +564,201 @@ struct future final {
...
+  std::uintptr_t scope_;
+  using stop_token_type = stop_token_type_t<Receiver>;
+  fused_stop_source<stop_token_type, inplace_stop_token> stopSource_;
```
ispeters, line 681: I wonder if it'd be worth choosing some alias names to make it clear which token comes from where.  The type you've named `stop_token_type` could be `receiver_stop_token_t`, and you're using `inplace_stop_token` directly, but it could be `scope_stop_token_t`.
```suggestion
  using receiver_stop_token_t = stop_token_type_t<Receiver>;
  using scope_stop_token_t = inplace_stop_token;
  fused_stop_source<receiver_stop_token_t, scope_stop_token_t> stopSource_;
```

include/unifex/async_scope.hpp
```diff
@@ -561,6 +564,201 @@ struct future final {
...
+      op.stopSource_.register_callbacks(
+          op.receiverToken_, op.scope_ref()->get_stop_token());
+      unifex::start(op.op_.get());
+    } else {
+      unifex::set_done(std::move(op).receiver_.get());
+    }
+  }
+
+private:
+  static constexpr std::uintptr_t mask = ~(std::uintptr_t{1u});
```
ispeters, line 677: What do you think of the following for a little more specificity in the names?
```suggestion
  static constexpr std::uintptr_t started_bit{1};
  static constexpr std::uintptr_t started_mask = ~started_bit;
```

include/unifex/async_scope.hpp
```diff
@@ -561,6 +564,201 @@ struct future final {
...
+  union {
+    manual_lifetime<nested_operation_t> op_;
+    manual_lifetime<Receiver> receiver_;
+  };
+
+  async_scope* scope_ref() const noexcept {
+    return reinterpret_cast<async_scope*>(scope_ & mask);
+  }
+  bool has_started() const noexcept { return scope_ & ~mask; }
+  void set_started() noexcept { scope_ |= 1u; }
```
ispeters, line 692: If you take the suggestion above re: names, this would be:
```suggestion
  async_scope* scope_ref() const noexcept {
    return reinterpret_cast<async_scope*>(scope_ & started_mask);
  }
  bool has_started() const noexcept { return scope_ & started_bit; }
  void set_started() noexcept { scope_ |= started_bit; }
```

include/unifex/async_scope.hpp
```diff
@@ -561,6 +564,201 @@ struct future final {
...
+};
+
+template <typename Sender>
+struct _attached_sender {
```
ispeters, line 696: Dunno that it makes much difference, but I like the consistency of marking these things `final`.
```suggestion
struct _attached_sender final {
```

test/async_scope_test.cpp
```diff
@@ -485,3 +469,201 @@ TEST_F(async_scope_test, lots_of_threads_works) {
 
   EXPECT_EQ(count.load(std::memory_order_relaxed), 0);
 }
+
+TEST_F(async_scope_test, attach) {
+  mock_receiver<void()> receiver;
```
ispeters, line 474: What's this doing here?
```suggestion
```

ispeters (approved):
LGTM

include/unifex/async_scope.hpp
janondrusek: That was my reasoning as well. Rather than introducing another branch, that obfuscates control flow.

include/unifex/async_scope.hpp
janondrusek: Agreed. At some point we should make a pass on the whole codebase and fix patterns, that we do not wish to spread.

test/async_scope_test.cpp
janondrusek: Waiting for removal :)

include/unifex/async_scope.hpp
janondrusek: There is, fixed!

include/unifex/async_scope.hpp
janondrusek: Added into tests, since `async_scope` is an incomplete type at this point.

PR #393, opened 2022-02-09
add nothrow_connectable test for `let_value_with_stop_source()`

janondrusek:
* missing regression test case, like `let_value_with_stop_token()`
* reduce code duplication in `let_value_with_stop_token_test`

### DIFF
--- a/test/let_value_with_stop_token_test.cpp
+++ b/test/let_value_with_stop_token_test.cpp
@@ -63,10 +63,16 @@ struct LetWithStopToken : testing::Test {
 
   void SetUp() override { DestructionCounter::destroyCount() = 0; }
 };
+namespace {
+template <typename StopToken, typename Callback>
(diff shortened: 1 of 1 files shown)


test/let_value_with_stop_token_test.cpp
```diff
@@ -97,6 +98,40 @@ TEST_F(LetWithStopToken, Simple) {
   EXPECT_EQ(external_context, 42);
 }
 
+TEST_F(LetWithStopToken, InplaceStoppableStopSourceNoexcept) {
```
MarkCMann, line 101: Can you add a test for a sender that isn't noexcept qualified? It'd be good to ensure that let_value_with_stop_source isn't noexcept qualified when the lambda isn't noexcept qualified.

MarkCMann (commented):
Code changes look good, but we need a test for when `let_value_with_stop_source` shouldn't be no throw connectable.

PR #397, opened 2022-02-22
Adding Variant Sender

MarkCMann:
Returning multiple senders from within a `let_value*` function requires either type erasing the results or inventing new types to represent the multiple senders. This is cumbersome since type erasure isn't always needed. 

This PR introduces the VariantSender type that wraps an underlying sender and provides an easy mechanism to create new Sender types that collapse multiple other Senders.

### DIFF
--- a/include/unifex/variant_sender.hpp
+++ b/include/unifex/variant_sender.hpp
@@ -78,9 +78,6 @@ struct _sender {
 template <typename... Senders>
 using sender = typename _sender<Senders...>::type;
 
-template<typename Sender>
-struct sends_done_impl : std::bool_constant<sender_traits<Sender>::sends_done> {};
-
 template <typename... Senders>
 class _sender<Senders...>::type {
   std::variant<Senders...> sender_variant;
@@ -103,7 +100,7 @@ class _sender<Senders...>::type {
     : sender_variant(std::forward<ConcreteSender>(concrete_sender)) {}
 
   template(typename This, typename Receiver)
-    (requires same_as<remove_cvref_t<This>, type> AND receiver<Receiver> AND std::conjunction_v<unifex::is_connectable<Senders, Receiver>...>)
+    (requires same_as<remove_cvref_t<This>, type> AND std::conjunction_v<std::bool_constant<sender_to<member_t<This, Senders>, Receiver>>...>)
   friend auto tag_invoke(tag_t<connect>, This&& that, Receiver&& r)
     noexcept(std::conjunction_v<unifex::is_nothrow_connectable<Senders, Receiver>...>)
   {
--- a/include/unifex/variant_sender.hpp
+++ b/include/unifex/variant_sender.hpp
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <unifex/config.hpp>
+#include <unifex/receiver_concepts.hpp>
+#include <unifex/sender_concepts.hpp>
+#include <unifex/blocking.hpp>
+#include <unifex/std_concepts.hpp>
+
+#include <exception>
+#include <tuple>
+#include <type_traits>
+#include <utility>
+#include <variant>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+namespace _variant_sender {
+
+template <typename... Ops>
+struct _op {
+  struct type;
+};
+template <typename... Ops>
+using operation = typename _op<Ops...>::type;
+
+template <typename... Ops>
+struct _op<Ops...>::type {
+  void start() & noexcept {
+    std::visit([](auto& op) noexcept(noexcept(unifex::start(op))) { unifex::start(op); }, variant_op);
+  }
+
+  std::variant<Ops...> variant_op;
+};
+
+
+template <_block::_enum First, _block::_enum... Rest>
+struct max_blocking_kind {
+  constexpr auto operator()() noexcept { return First; }
+};
+
+template <_block::_enum First, _block::_enum Second, _block::_enum... Rest>
+struct max_blocking_kind<First, Second, Rest...> {
+  constexpr auto operator()() noexcept {
+    if constexpr (First == Second) {
+      return max_blocking_kind<First, Rest...>{}();
+    } else if constexpr (
+        (First == blocking_kind::always &&
+        Second == blocking_kind::always_inline) ||
+        (Second == blocking_kind::always &&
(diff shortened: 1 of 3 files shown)


include/unifex/variant_sender.hpp
```diff
@@ -0,0 +1,124 @@
...
+    noexcept(std::is_nothrow_constructible_v<std::variant<Senders...>, decltype(concrete_sender)>)
+    : sender_variant(std::forward<ConcreteSender>(concrete_sender)) {}
+
+  template(typename This, typename Receiver)
+    (requires same_as<remove_cvref_t<This>, type> AND std::conjunction_v<std::bool_constant<sender_to<member_t<This, Senders>, Receiver>>...>)
+  friend auto tag_invoke(tag_t<connect>, This&& that, Receiver&& r)
+    noexcept(std::conjunction_v<unifex::is_nothrow_connectable<Senders, Receiver>...>)
+  {
+    return operation<connect_result_t<Senders, Receiver>...>{
+        std::visit([r = static_cast<Receiver&&>(r)](auto&& sender) mutable noexcept(unifex::is_nothrow_connectable_v<decltype(sender), Receiver>) {
```
ispeters, line 108: This lambda is executed synchronously so I think you can capture `r` by reference and forward directly into the nested `connect` instead of forwarding and then moving.
```suggestion
        std::visit([&r](auto&& sender) noexcept(unifex::is_nothrow_connectable_v<decltype(sender), Receiver>) {
```

include/unifex/variant_sender.hpp
```diff
@@ -0,0 +1,124 @@
...
+            return std::variant<connect_result_t<Senders, Receiver>...>{unifex::connect(std::move(sender), static_cast<Receiver&&>(r))};
+        }, std::move(static_cast<This&&>(that).sender_variant))
```
ispeters, line 110: I think you're moving the sender when you should be forwarding it here.
```suggestion
            return std::variant<connect_result_t<Senders, Receiver>...>{unifex::connect(static_cast<decltype(sender)>(sender), static_cast<Receiver&&>(r))};
        }, static_cast<This&&>(that).sender_variant)
```

include/unifex/variant_sender.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  template<typename ConcreteSender>
+  type(ConcreteSender&& concrete_sender)
+    noexcept(std::is_nothrow_constructible_v<std::variant<Senders...>, decltype(concrete_sender)>)
+    : sender_variant(std::forward<ConcreteSender>(concrete_sender)) {}
+
+  template(typename This, typename Receiver)
+    (requires same_as<remove_cvref_t<This>, type> AND std::conjunction_v<std::bool_constant<sender_to<member_t<This, Senders>, Receiver>>...>)
+  friend auto tag_invoke(tag_t<connect>, This&& that, Receiver&& r)
+    noexcept(std::conjunction_v<unifex::is_nothrow_connectable<Senders, Receiver>...>)
```
ispeters, line 105: I think this `noexcept` clause is imprecise.  I think the member variant ought to be forwarded into `visit`, which means the `connect` expression may be given either an lvalue or an rvalue for any of the constituent senders.

I wonder what the value categories end up being in the cross product of (`This` is an value, `This` is an rvalue) x (one of `Senders` is a plain type, one of `Senders` is an lvalue reference, one of `Senders` is an rvalue reference).

ispeters (changes requested):
Looks good, but I have some questions about `noexcept` and value categories.

Can you add some test cases to cover the weird value category corner cases, and at least some minimal documentation to the API reference, please?

include/unifex/variant_sender.hpp
MarkCMann: Seems that [`std::variant` isn't permitted to hold references](https://en.cppreference.com/w/cpp/utility/variant):
> A variant is not permitted to hold references, arrays, or the type void. Empty variants are also ill-formed (std::variant<[std::monostate](http://en.cppreference.com/w/cpp/utility/variant/monostate)> can be used instead).

So I'll add some tests for when `This` is an lvalue/rvalue.

test/variant_sender_test.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    void set_error(std::exception_ptr) noexcept {}
+};
+
+template<bool lvalueConnectNoexcept = true, bool rvalueConnectNoexcept = true>
+struct TestSender {
+  template <
+    template <typename...> class Variant,
+    template <typename...> class Tuple>
+  using value_types = Variant<std::string>;
```
ispeters, line 55: I think you need to use `Tuple` somewhere to make this compatible with all the type machinery, no?
```suggestion
  using value_types = Variant<Tuple<std::string>>;
```

test/variant_sender_test.cpp
```diff
@@ -0,0 +1,204 @@
...
+}
+
+template<bool lvalueConnectNoexcept, bool rvalueConnectNoexcept>
+using test_sender_t = variant_sender<TestSender<lvalueConnectNoexcept, rvalueConnectNoexcept>>;
+
+template<typename T, bool lvalue>
+using conditionally_lvalue_t = std::conditional_t<lvalue, std::add_lvalue_reference_t<T>, T>;
+
+template<bool lvalueConnectNoexcept, bool rvalueConnectNoexcept, bool isLvalueReference = true>
+using is_noexcept = unifex::is_nothrow_connectable<conditionally_lvalue_t<test_sender_t<lvalueConnectNoexcept, rvalueConnectNoexcept>, isLvalueReference>, IntAndStringReceiver>;
```
ispeters, line 176: These should be in an anonymous namespace, I think.
```suggestion
namespace {

template<bool lvalueConnectNoexcept, bool rvalueConnectNoexcept>
using test_sender_t = variant_sender<TestSender<lvalueConnectNoexcept, rvalueConnectNoexcept>>;

template<typename T, bool lvalue>
using conditionally_lvalue_t = std::conditional_t<lvalue, std::add_lvalue_reference_t<T>, T>;

template<bool lvalueConnectNoexcept, bool rvalueConnectNoexcept, bool isLvalueReference = true>
using is_noexcept = unifex::is_nothrow_connectable<conditionally_lvalue_t<test_sender_t<lvalueConnectNoexcept, rvalueConnectNoexcept>, isLvalueReference>, IntAndStringReceiver>;

} // namespace
```

test/variant_sender_test.cpp
```diff
@@ -0,0 +1,204 @@
...
+  static constexpr bool sends_done = true;
+
+  struct op {};
+
+  template<typename Receiver>
+  op connect(Receiver&& r) & noexcept(lvalueConnectNoexcept);
+
+  template<typename Receiver>
+  op connect(Receiver&& r) && noexcept(rvalueConnectNoexcept);
+};
```
ispeters, line 69: I think these should be in an anonymous namespace.
```suggestion
namespace {

struct IntAndStringReceiver {
    void set_value(int) noexcept {

    }

    void set_value(std::string) noexcept {

    }

    void set_done() noexcept {}

    void set_error(std::exception_ptr) noexcept {}
};

template<bool lvalueConnectNoexcept = true, bool rvalueConnectNoexcept = true>
struct TestSender {
  template <
    template <typename...> class Variant,
    template <typename...> class Tuple>
  using value_types = Variant<std::string>;

  template <template <typename...> class Variant>
  using error_types = Variant<>;

  static constexpr bool sends_done = true;

  struct op {};

  template<typename Receiver>
  op connect(Receiver&& r) & noexcept(lvalueConnectNoexcept);

  template<typename Receiver>
  op connect(Receiver&& r) && noexcept(rvalueConnectNoexcept);
};

} // namespace
```

ispeters (approved):
Some nits; I think you can squash and merge after you address them.

PR #398, opened 2022-02-26
Remove <unifex/tuple.hpp>

ispeters:
We don't need the `unifex::apply` polyfill anymore so let's get rid of
it.


MarkCMann (approved):
Looks great, thanks for cleaning up the now unnecessary polyfill!

PR #400, opened 2022-03-01
Fix detach_on_cancel error_types propagation

MarkCMann:
`detach_on_cancel:: _sender<Sender>::type::error_types` was producing an invalid variant type causing composition of other algorithms on top to fail. This is a fix + unit test to ensure the issue doesn't recur moving forward.

### DIFF
--- a/test/detach_on_cancel_test.cpp
+++ b/test/detach_on_cancel_test.cpp
@@ -6,6 +6,7 @@
 #include <atomic>
 #include <memory>
 #include <stdexcept>
+#include <variant>
 
 #include <unifex/inline_scheduler.hpp>
 #include <unifex/inplace_stop_token.hpp>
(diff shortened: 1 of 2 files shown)


jesswong (approved):

test/detach_on_cancel_test.cpp
```diff
@@ -162,3 +163,12 @@ TEST_F(detach_on_cancel_test, cancellation_and_completion_race) {
 
   EXPECT_EQ(max_iterations, count.load());
 }
+
+TEST_F(detach_on_cancel_test, error_types_propagate) {
+  using namespace unifex;
+  using error_types =
+    sender_error_types_t<decltype(detach_on_cancel(just())), type_list>;
+  using v = typename error_types::template apply<std::variant>;
```
ispeters, line 171: In the `broken-stdlib` branch, we're generally using `unifex:: variant` instead of `std::variant`; can't decide if it matters here, and I'm not sure whether we've got `unifex::variant_size` for the assertion, below.

```suggestion
  using v = typename error_types::template apply<unifex::variant>;
```

ispeters (approved):
Lgtm

PR #404, opened 2022-03-08
Make `type_erased_stream` `Scheduler` aware

janondrusek:
* expose `unifex::type_erased_stream` (result of `unifex::type_erase()`)
* add `get_scheduler` CPO into `type_erased_stream`

### DIFF
--- a/include/unifex/type_erased_stream.hpp
+++ b/include/unifex/type_erased_stream.hpp
@@ -15,9 +15,11 @@
  */
 #pragma once
 
+#include <unifex/any_scheduler.hpp>
 #include <unifex/async_trace.hpp>
 #include <unifex/config.hpp>
 #include <unifex/receiver_concepts.hpp>
+#include <unifex/scheduler_concepts.hpp>
 #include <unifex/stream_concepts.hpp>
 #include <unifex/sender_concepts.hpp>
 #include <unifex/inplace_stop_token.hpp>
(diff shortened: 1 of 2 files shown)


include/unifex/type_erased_stream.hpp
```diff
@@ -107,18 +126,23 @@ struct _stream<Values...>::type {
       continuation_info get_continuation_info() const noexcept override {
         return continuation_info::from_continuation(receiver_);
       }
+
+      any_scheduler_ref get_scheduler() const noexcept override {
+        return scheduler_;
```
MarkCMann, line 131: Do you need the scheduler to be a member variable? Seems that you could delay evaluation of `get_scheduler` until this point and save on some byes.

include/unifex/type_erased_stream.hpp
```diff
@@ -132,22 +156,26 @@ struct _stream<Values...>::type {
...
```
MarkCMann, line 161: Same question as above

include/unifex/type_erased_stream.hpp
MarkCMann: get_scheduler returns schedulers by value in all of the other cases, so this class needs to store the downstream receivers scheduler by value as a member to ensure that the scheduler reference passed to any_scheduler_ref lives beyond the scope of this function call.

janondrusek: Although not documented as a requirement: all existing `get_scheduler()` implementations return by value. If we returned inline, `any_scheduler_ref` would take a temporary.

ispeters (approved):
I'm assuming you've rebased onto the recent `task<>` fix so lgtm.

PR #405, opened 2022-03-12
Fix scheduler affinity

ispeters:
We have been storing a `task<>`'s scheduler as an `any_scheduler_ref`,
which has proven to be a source of use-after-free bugs.  This change
switches all the `any_scheduler_ref`s to `any_scheduler`s, fixing the
lifetime issues.


janondrusek (approved):
Tested internally.

PR #408, opened 2022-03-14
Remove as_const polyfill from utility

MarkCMann:
We no longer need the `as_const` polyfill in the broken-stdlib branch. Let's remove it to bring broken-stdlib closer to main


ispeters (approved):
LGTM

PR #410, opened 2022-03-22
new algorithm unifex::unstoppable

lehecka:
This algorithm is helping composing the non-cancellable sender.


ispeters (approved):
This looks good; accepting and merging to unblock some Meta-internal work, but we're going to need to follow up with some documentation—we've been bad about landing changes without docs.

PR #411, opened 2022-03-23
adding complete_adapt_stream algorithm

lehecka:
similar to next_adapt_steram, adding complete_adapt_stream

### DIFF
--- a/include/unifex/complete_adapt_stream.hpp
+++ b/include/unifex/complete_adapt_stream.hpp
@@ -1,0 +1,52 @@
+// (c) Meta Platforms, Inc. and affiliates. Confidential and proprietary.
+
+#pragma once
+
+#include <unifex/stream_concepts.hpp>
+
+#include <functional>
+#include <type_traits>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+namespace _complete_adapt {
+template <typename Stream, typename AdaptFunc>
+struct _stream {
+  struct type;
+};
+template <typename Stream, typename AdaptFunc>
+using stream =
+    typename _stream<remove_cvref_t<Stream>, remove_cvref_t<AdaptFunc>>::type;
+
+template <typename Stream, typename AdaptFunc>
+struct _stream<Stream, AdaptFunc>::type {
+  Stream innerStream_;
+  AdaptFunc adapter_;
+
+  friend auto tag_invoke(tag_t<next>, type& s) -> next_sender_t<Stream> {
+    return next(s.innerStream_);
+  }
+
+  friend auto tag_invoke(tag_t<cleanup>, type& s)
+      -> std::invoke_result_t<AdaptFunc&, cleanup_sender_t<Stream>> {
+    return std::invoke(s.adapter_, cleanup(s.innerStream_));
+  }
+};
+}  // namespace _complete_adapt
+
+namespace _complete_adapt_cpo {
+inline const struct _fn {
+  template <typename Stream, typename AdaptFunc>
+  auto operator()(Stream&& stream, AdaptFunc&& adapt) const {
+    return _complete_adapt::stream<Stream, AdaptFunc>{
+        (Stream &&) stream, (AdaptFunc &&) adapt};
+  }
+} complete_adapt_stream{};
+}  // namespace _complete_adapt_cpo
+
+using _complete_adapt_cpo::complete_adapt_stream;
+
+}  // namespace unifex
+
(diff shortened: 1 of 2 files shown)


include/unifex/complete_adapt_stream.hpp
```diff
@@ -0,0 +1,52 @@
...
+  struct type;
+};
+template <typename Stream, typename AdaptFunc>
+using stream =
+    typename _stream<remove_cvref_t<Stream>, remove_cvref_t<AdaptFunc>>::type;
+
+template <typename Stream, typename AdaptFunc>
+struct _stream<Stream, AdaptFunc>::type {
+  Stream innerStream_;
+  AdaptFunc adapter_;
```
MarkCMann, line 25: These would likely benefit from being annotated as [[no_unique_address]] since both types could in theory be memberless.

include/unifex/complete_adapt_stream.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace unifex {
+namespace _complete_adapt {
+template <typename Stream, typename AdaptFunc>
+struct _stream {
+  struct type;
+};
+template <typename Stream, typename AdaptFunc>
+using stream =
+    typename _stream<remove_cvref_t<Stream>, remove_cvref_t<AdaptFunc>>::type;
```
MarkCMann, line 20: The AdaptFunc should likely be decay'd here instead of just removing cvref -- similar to the [`adapt_stream` algorithm](https://github.com/facebookexperimental/libunifex/blob/2547d87fb98a1fe70eccc170a8c91b144648517d/include/unifex/adapt_stream.hpp#L33-L35)

MarkCMann:
Could this instead be written as using the existing `adapt_stream` algorithm but passing the identity function for the [`adapt_next` param](https://github.com/facebookexperimental/libunifex/blob/2547d87fb98a1fe70eccc170a8c91b144648517d/include/unifex/adapt_stream.hpp#L89)? That would make this algorithm more maintainable in the long run.

lehecka:
I used the next_adapt_stream implementation as a template for this algorithm. I can see how both next_adapt_stream and complete_adapt_stream can be written on top of adapt_stream. I am assuming this would be more optimal implementation since it saves some bytes for storing the transformation lambda for the other signal which is not used.

include/unifex/complete_adapt_stream.hpp
lehecka: if there is a consensus on the suggested change, I would add it to next_adapt_stream as well.

MarkCMann:
If the function is just an identity transformation then it shouldn't take up any additional space in the `adapt_stream` type (so long as the functions are annotated with `[[no_unique_address]]`). Check out [this godbolt](https://godbolt.org/z/nnj8aqss9) for an example of how wrapping an identity transform function doesn't require additional storage in the wrapper class.

include/unifex/complete_adapt_stream.hpp
ispeters: Looks to me like we _need_ `std::decay_t` here to support `AdaptFunc` being deduced as a function rather than a function pointer: https://godbolt.org/z/eMqW8x93b.

The error in that Godbolt goes away if you manually take the address of `foo`: https://godbolt.org/z/YzE3zvPYn

I expect to be able to construct a function pointer by just naming the function whose address I want (i.e. I don't expect to have to write the `&` in `&foo`) so I think this is a real ergonomics problem.

include/unifex/complete_adapt_stream.hpp
```diff
@@ -0,0 +1,52 @@
...
+}  // namespace _complete_adapt
+
+namespace _complete_adapt_cpo {
+inline const struct _fn {
+  template <typename Stream, typename AdaptFunc>
+  auto operator()(Stream&& stream, AdaptFunc&& adapt) const {
+    return _complete_adapt::stream<Stream, AdaptFunc>{
+        (Stream &&) stream, (AdaptFunc &&) adapt};
+  }
+} complete_adapt_stream{};
```
ispeters, line 45: I think `next_adapt_stream` is so-named because it only adapts the next-sender of the adapted stream.  To mirror that, I think this should be called `cleanup_adapt_stream` rather than `complete_adapt_stream`.
```suggestion
} cleanup_adapt_stream{};
```

ispeters (changes requested):
Looks great, but let's make the requested changes first.

janondrusek:
replaced by #517

PR #412, opened 2022-03-23
Use liburing headers instead of system headers for io_uring

Skory:
Since we use liburing to enable io_uring context it makes sense to consistently use its headers instead of system headers.

### DIFF
--- a/source/linux/io_uring_syscall.hpp
+++ b/source/linux/io_uring_syscall.hpp
@@ -16,7 +16,7 @@
 #pragma once
 
 #include <signal.h>
-#include <linux/io_uring.h>
+#include <liburing/io_uring.h>
 
 namespace unifex::linuxos
 {


janondrusek (approved):
[FindLibUring.cmake](https://github.com/facebookexperimental/libunifex/blob/8cb099da2d4e1401232714cf7d7d17b6d817f65e/cmake/FindLibUring.cmake) is meant to find and append `liburing/` to include paths.

lewissbaker:
I wonder, do we actually need to depend on liburing?
The original intent was to just depend on `<linux/io_uring.h>` and low-level syscall API.
My memory is a little hazy but I think the reason we added a dependency on liburing was just some limitation with using the `<linux/io_uring.h>` header (there was some type it used that wasn't available at one stage - which liburing provided a polyfill for). It may also have been that the linux system headers on the system I was building with did not yet have io_uring.h at the time.
However, I suspect that linux/io_uring.h header is much more widely available now (2+ years later).

Should we just try to remove the dependency on liburing altogether instead?

PR #414, opened 2022-03-28
fixing linter error

lehecka:
static member accessed through instance

fixing the above lister which complains that we are calling static method like a member method


ispeters (approved):

ispeters:
I'll merge this once I see the CI is all green.

PR #416, opened 2022-04-26
Fix asan error for let_value with coroutines

jesswong:
There's an ASAN issue in let_value that only repros when a coroutine is being used to evaluate the value. This diff adds the test that will repro without the fix, and will show that the fix causes this test to pass.

This was tested by enabling coroutines and asan in unifex tests.

### DIFF
--- a/include/unifex/let_value.hpp
+++ b/include/unifex/let_value.hpp
@@ -56,17 +56,18 @@ struct _successor_receiver<Operation, Values...>::type {
 
   template <typename... SuccessorValues>
   void set_value(SuccessorValues&&... values) && noexcept {
+    auto& op = op_;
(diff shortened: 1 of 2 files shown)


include/unifex/let_value.hpp
```diff
@@ -56,17 +56,18 @@ struct _successor_receiver<Operation, Values...>::type {
 
   template <typename... SuccessorValues>
   void set_value(SuccessorValues&&... values) && noexcept {
+    auto& op = op_;
```
lewissbaker, line 59: You also need to apply the same change to the `set_done()` and `set_error()` methods.
Otherwise, looks good.

ispeters (approved):
LGTM!

PR #418, opened 2022-05-12
Fix let_value to query sender properties through sender_traits

AnujYamdagni:
let_value queries its predecessor Sender's value_types incorrectly.
It assumes the Sender has a value_types member alias template, but it may not.  The "right" way is to use sender_traits<Predecessor>.


ispeters (approved):
LGTM!

PR #420, opened 2022-05-19
Fix compile issue related to io_uring (#373)

janondrusek:
* Fix the compilation bug due to a layout change

Layout change in struct io_uring_sqe of io_uring and liburing causes a missing padding field.
Change strategy and zero all struct member like suggested in #351 review

* Update commented sqe layout with the last version from the kernel

* Add two new build targets with the liburing support


ispeters (approved):
LGTM

PR #421, opened 2022-05-19
Fix race condition in `when_all`

janondrusek:
* `request_stop()` from `set_error()` in one *arm* is *racing* with `set_done()` in another *arm*
* introduce ownership transfer of `deliver_result` to and from
  `cancel_operation`

### DIFF
--- a/include/unifex/when_all.hpp
+++ b/include/unifex/when_all.hpp
@@ -93,20 +93,20 @@ struct _operation_tuple<Index, Receiver>::type {
   void start() noexcept {}
 };
 
-struct cancel_operation {
-  inplace_stop_source& stopSource_;
-
-  void operator()() noexcept {
-    stopSource_.request_stop();
-  }
-};
-
 template <typename Receiver, typename... Senders>
 struct _op {
   struct type;
 };
 template <typename Receiver, typename... Senders>
 using operation = typename _op<remove_cvref_t<Receiver>, Senders...>::type;
+
+template <typename Receiver, typename... Senders>
+struct cancel_operation {
+  operation<Receiver, Senders...>& op_;
+  void operator()() noexcept {
+    op_.request_stop();
+  }
+};
 
 template <typename... Errors>
 using unique_decayed_error_types = concat_type_lists_unique_t<
@@ -213,19 +213,39 @@ struct _op<Receiver, Senders...>::type {
 
   void start() noexcept {
     stopCallback_.construct(
-        get_stop_token(receiver_), cancel_operation{stopSource_});
+        get_stop_token(receiver_),
+        cancel_operation<Receiver, Senders...>{*this});
     ops_.start();
   }
 
-  private:
+  void request_stop() noexcept {
+    // mark callback as running (own deliver_result)
+    auto oldCount = refCount_.fetch_add(1, std::memory_order_relaxed);
+    UNIFEX_ASSERT(!(oldCount & callback_running_bit));
+    stopSource_.request_stop();
+    // unmark (deliver_result back to operation)
+    oldCount = refCount_.fetch_sub(1, std::memory_order_relaxed);
+    UNIFEX_ASSERT(oldCount & callback_running_bit);
+    // deliver_result as there are no more operations
+    if (oldCount == 1) {
+      unifex::set_done(std::move(receiver_));
+    }
+  }
+
+private:
   void element_complete() noexcept {
-    if (refCount_.fetch_sub(1, std::memory_order_acq_rel) == 1) {
+    // +2 for last operation, +1 for callback running
+    if (refCount_.fetch_sub(2, std::memory_order_release) < 4) {
       deliver_result();
     }
   }
 
   void deliver_result() noexcept {
     stopCallback_.destruct();
+    if (refCount_.load(std::memory_order_acquire) == 1) {
+      // request_stop owns deliver_result (callback running)
+      return;
+    }
 
     if (get_stop_token(receiver_).stop_requested()) {
       unifex::set_done(std::move(receiver_));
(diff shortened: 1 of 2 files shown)


include/unifex/when_all.hpp
```diff
@@ -255,13 +275,16 @@ struct _op<Receiver, Senders...>::type {
     }
   }
 
+  static constexpr std::size_t callback_running_bit{1};
   std::tuple<optional<value_variant_for_sender<remove_cvref_t<Senders>>>...> values_;
   optional<error_types<variant, remove_cvref_t<Senders>...>> error_;
-  std::atomic<std::size_t> refCount_{sizeof...(Senders)};
+  // count multiplied by to to allow add 1 when stop is requested
```
ispeters, line 281: Nit: "two", not "to"
```suggestion
  // count multiplied by two to allow add 1 when stop is requested
```
Alternatively,
```suggestion
  // count left-shifted by one to allow add 1 when stop is requested
```

ispeters (approved):
LGTM

include/unifex/when_all.hpp
```diff
@@ -213,19 +213,39 @@ struct _op<Receiver, Senders...>::type {
...
+    if (oldCount == 1) {
+      unifex::set_done(std::move(receiver_));
+    }
+  }
+
+private:
   void element_complete() noexcept {
-    if (refCount_.fetch_sub(1, std::memory_order_acq_rel) == 1) {
+    // +2 for last operation, +1 for callback running
+    if (refCount_.fetch_sub(2, std::memory_order_acq_rel) < 4) {
```
MarkCMann, line 238: You should be able to relax this to just a release operation if you make the below load in `deliver_result` an acquire operation. 

As is, I'm not sure how other threads would be able to take advantage of this acquire operation (no way to determine which threads have already completed).

janondrusek: ASAN problem was only triggered in a single threaded scenario when `request_stop` resulted in calling `set_done()` inline. If another thread 'wins' the `deliver_result`, it would be blocked on `stopCallback_.destruct()`.

include/unifex/when_all.hpp
```diff
@@ -213,19 +213,39 @@ struct _op<Receiver, Senders...>::type {
...
   }
 
-  private:
+  void request_stop() noexcept {
+    // mark callback as running (own deliver_result)
+    auto oldCount = refCount_.fetch_add(1, std::memory_order_release);
+    UNIFEX_ASSERT(!(oldCount & callback_running_bit));
+    stopSource_.request_stop();
+    // unmark (deliver_result back to operation)
+    oldCount = refCount_.fetch_sub(1, std::memory_order_acq_rel);
```
ispeters, line 227: If I've understood @MarkCMann's comments below, I think we can relax these atomic operations, too.
```suggestion
    auto oldCount = refCount_.fetch_add(1, std::memory_order_relaxed);
    UNIFEX_ASSERT(!(oldCount & callback_running_bit));
    stopSource_.request_stop();
    // unmark (deliver_result back to operation)
    oldCount = refCount_.fetch_sub(1, std::memory_order_relaxed);
```

The buggy scenario this diff fixes is that a single thread is running `request_stop()`, which synchronously invokes a stop callback that synchronously completes the last of the `when_all`'s child operations.  In that scenario, the old code would destroy the `when_all`'s operation state out from under the `stopSource_` while it's still processing stop callbacks.  In this scenario, the new code defers operation state destruction until we invoke `set_done` on line 231.

In the buggy scenario in the new code, we have the following sequence, all on the same thread:
 1. _Receiver_ requests stop, invoking `request_stop()`,
 2. `request_stop()` makes `refCount_` odd and invokes `stopSource_.request_stop()`,
 3. `stopSource_.request_stop()` invokes stop callbacks,
 4. some stop callback invokes the last `element_complete()`,
 5. `element_complete()` decrements `refCount_` from 3 to 1 and invokes `deliver_result()`,
 6. `deliver_result()` destroys the stop callback (doing not much because it's still running on the current thread),
 7. `deliver_result()` observes `refCount_` is 1 and returns, unwinding back up to `request_stop()`.
 8. `request_stop()` decrements `refCount_` from 1 to 0 and invokes `set_done`.

In this flow, everything happens on the same thread so no synchronization is necessary between `request_stop()` and `element_complete()` or `deliver_result()`.  We still do require that the thread observe the results of all previously-completed operations, but that's accomplished by the load-acquire in `deliver_result()`, which happens-before the relaxed subtraction in `request_stop()`.

There are a few other scenarios:
 1. _Receiver_ never requests stop;
 2. _Receiver_ requests stop on thread T1 and this completes before the `when_all`'s final child operation completes on thread T2;
 3. _Receiver_ requests stop on T1 while `when_all`'s final child is busy completing on T2.

Scenario 1 is trivial—there's no race.

In scenario 2, the stop callback increments and decrements `refCount_` before the `when_all`'s final child drops `refCount_` to zero.  There are no memory effects on the stop-requesting thread that need to be observed by the final child so it's enough to ensure that refcount manipulations are well-ordered, which is provided by relaxed ordering.

In scenario 3, the stop callback is racing with `when_all` completion and the race is resolved by destroying the stop callback in `deliver_result()`.  Either the stop callback is destroyed:
 - _before_ it is invoked and will thus never be invoked;
 - _after_ it was invoked and we're back in scenario 2; or
 - _while_ it's being invoked, in which case destruction blocks, forcing us into scenario 2 with a full memory barrier.

ispeters (changes requested):
I think we should take @MarkCMann's advice; we should relax the acquire-release semantics in `element_complete()` to release, strengthen the load-relaxed to load-acquire in `deliver_result()`, and relax the increment and decrement to relaxed in `request_stop()`.

ispeters (approved):

include/unifex/when_all.hpp
```diff
@@ -213,19 +213,39 @@ struct _op<Receiver, Senders...>::type {
...
   }
 
-  private:
+  void request_stop() noexcept {
+    // mark callback as running (own deliver_result)
+    auto oldCount = refCount_.fetch_add(1, std::memory_order_relaxed);
+    UNIFEX_ASSERT(!(oldCount & callback_running_bit));
+    stopSource_.request_stop();
+    // unmark (deliver_result back to operation)
+    oldCount = refCount_.fetch_sub(1, std::memory_order_relaxed);
```
lewissbaker, line 227: I think this at least needs to be a `acq_rel` so that the `set_done()` call below has visibility of any other operations that completed while we were calling `request_stop()`.

It needs to be `release` to ensure that the store to the stopSource_ in `request_stop()` is visible to another thread that sees our write and goes on to destroy the operation state.

PR #428, opened 2022-06-20
make `async_scope::attached_sender` copyable

janondrusek:
* copy constructor calls `async_scope::try_record_start` internally
* copy of attached Sender will increment outstanding number of
  operations on async_scope


ispeters (approved):
LGTM

PR #429, opened 2022-07-01
Reduce lock contentions in static_thread_pool

arauter:
The global lock used for the `static_thread_pool` is still being held before notifying a waiting thread, therefore immediately blocking the waiting thread again. Instead the lock should be released beforehand to reduce lock contentions.

### DIFF
--- a/source/static_thread_pool.cpp
+++ b/source/static_thread_pool.cpp
@@ -122,12 +122,15 @@ namespace _static_thread_pool {
   }
 
   bool context::thread_state::try_push(task_base* task) {
-    std::unique_lock lk{mut_, std::try_to_lock};
-    if (!lk) {
-      return false;
+    bool wasEmpty;
+    {
+      std::unique_lock lk{mut_, std::try_to_lock};
+      if (!lk) {
+        return false;
+      }
+      wasEmpty = queue_.empty();
+      queue_.push_back(task);
     }
-    const bool wasEmpty = queue_.empty();
-    queue_.push_back(task);
     if (wasEmpty) {
       cv_.notify_one();
     }
@@ -135,17 +138,22 @@ namespace _static_thread_pool {
   }
 
   void context::thread_state::push(task_base* task) {
-    std::lock_guard lk{mut_};
-    const bool wasEmpty = queue_.empty();
-    queue_.push_back(task);
+    bool wasEmpty;
+    {
+      std::lock_guard lk{mut_};
+      wasEmpty = queue_.empty();
+      queue_.push_back(task);
+    }
     if (wasEmpty) {
       cv_.notify_one();
     }
   }
 
   void context::thread_state::request_stop() {
-    std::lock_guard lk{mut_};
-    stopRequested_ = true;
+    {
+      std::lock_guard lk{mut_};
+      stopRequested_ = true;
+    }
     cv_.notify_one();
   }
 


source/static_thread_pool.cpp
```diff
@@ -122,30 +122,38 @@ namespace _static_thread_pool {
...
-    queue_.push_back(task);
+    bool wasEmpty;
+    {
+      std::lock_guard lk{mut_};
+      wasEmpty = queue_.empty();
+      queue_.push_back(task);
+    }
     if (wasEmpty) {
       cv_.notify_one();
     }
```
lewissbaker, line 149: I'm pretty sure we _need_ to hold the lock while calling `cv_notify_one()` in this case.

Consider the case where another thread is spawning some work on the thread-pool and waiting for it to complete.
e.g. something like:
```
static_thread_pool tp1;
static_thread_pool tp2;
sync_wait(sequence(
  schedule(tp1.get_scheduler()),
  schedule(tp2.get_scheduler())));
```

Here the main thread spawns some work on tp1 which then enqueues work onto tp2.
Thus a tp1-owned thread will end up calling this `push()` method on tp2 thread-pool.

If the tp1-owned thread were to release the lock before calling `cv_.notify_one()` then it's possible that the tp1-thread might be context switched out just after releasing the lock and before it calls `cv_.notify_one()`.

Then a tp2 thread might then acquire the lock, see the task and execute it, without waiting for the condition_variable to be notified. The task executes which then signals that the `sync_wait()` can return.
The main thread's `sync_wait()` call then unblocks, returns, and then the main thread proceeds to destroy `tp2`.
This signals all tp2 threads to exit and waits for them to exit, then destroys both the mutex and condition_variable objects owned by tp2.

Then the main thread will try to destroy tp1 and this will then request those threads to exit and then join them.

If the tp1 thread that was originally calling `push()` on tp2 now wakes up and then calls `cv_.notify_one()` this is now going to be called on a dangling reference as tp1 (and all of its sub-objects) have now been destroyed.

To avoid this situation it is important for the `cv_.notify_one()` call to be made before the lock is released.

The same data-race applies to the `try_push()` method above.

Note that I don't think this same race can occur with the `request_stop()` methods below as this is only called from the destructor of the thread-pool and so we can guarantee that the calls to destructors of the condition_variables objects are sequenced after the calls to `notify_one()` on those condition variables.
However, the `request_stop()` calls only run during shutdown, which only runs once - so optimising this case is probably not going to yield much of a performance improvement.

arauter: @lewissbaker thanks for the quick review, I appreciate it.

I understand why you would want to hold the lock before calling `cv_notify_one()` in your example. I didn't consider a case where a tp1-owned thread would end up calling `push()` on the tp2 thread-pool. In most cases, wouldn't a tp1-owned thread always end up using the tp1 thread-pool? So is this something we should even be considering?

Assuming, tpX-owned threads will always end up calling `push()` on their original thread pool, there is certainly some performance penalty you pay for not releasing the lock before notifying a waiting thread. Or am I missing something here?

lewissbaker: The schedule() operation is often used to transition between different execution contexts (in fact it's impossible to schedule work onto thread pool yhreads without this) and so needs to take into account that it might be started from a non thread-pool thread.

Regarding performance, the existing behaviour only has the worker thread wait on the cv and the enqueuing thread calling `notify_one()` if the queue becomes empty. So under high contention scenarios where the queue never empties, the proposed change won't see any improvement.

The main case where there might be some improvement is where work is trickling in such that threads wake up, process one task and then go back to sleep before the next task is enqueued. However, even in this case, how often does the thread get woken up and go back to sleep because it can't acquire the mutex? 

Do you have some benchmarks that show that the existing behaviour is problematic?

One thing we could do to reduce the frequency of calling `notify_one()` is maintain an extra flag that is used to indicate the thread is waiting on the cv separately from the queue being empty. This would mean that a thread that is processing the last item in the queue (so that the queue is currently empty) won't cause a subsequent enqueue to its thread to notify until the worker thread has finished that task and is about to wait on the cv.

arauter: @lewissbaker I see your point, thanks for explaining it, I'm rather new to libunifex.

I would love to share some benchmarks but we use libunifex internally in my company so I can't really share those :) However, I looked at them again and realized that that the issue actually stems from the global lock that is being used inside the `static_thread_pool`. Under high load, there is a lot of lock contention especially inside`thread_state::pop()` which kind of makes sense. This PR won't obviously fix this.

PR #430, opened 2022-07-05
compiles and passes tests with clang-cl on windows

ericniebler:


cmake/unifex_env.cmake
```diff
@@ -19,7 +19,7 @@ else()
   message(WARNING "[unifex warning]: unknown compiler ${CMAKE_CXX_COMPILER_ID} !")
 endif()
 
-if (MSVC)
```
ericniebler, line 22: Not sure why the cmake variable here was called `MSVC`. I assume it was a mistake? Or an incomplete edit during maintenance? @ispeters ?

test/any_sender_of_test.cpp
```diff
@@ -55,14 +55,14 @@ namespace {
...
-template <bool Noexcept, typename... T>
+template <bool NoExcept, typename... T>
 struct AnySenderOfTestImpl : Test {
   using any_sender = any_sender_of<T...>;
 
   static constexpr size_t value_count = sizeof...(T);
 
   static_assert(typed_sender<any_sender>);
-  static_assert(sender_to<any_sender, mock_receiver<void(T...)>>);
+  static_assert(sender_to<any_sender, mock_receiver<void(T...) noexcept(NoExcept)>>);
```
ericniebler, line 65: @ispeters, before this change the `NoExcept` template parameter was unused, which seemed strange to me. My assumption is that the intention had been to test both receivers with potentially-throwing and non-throwing `set_value` functions.

test/mock_receiver.hpp
```diff
@@ -23,54 +23,59 @@
 
 namespace unifex_test {
```
ericniebler, line 25: _Note:_ All of the changes in this file are to accommodate clang-cl not mangling `noexcept` into function types. This leads to linker errors. Rather than parameterizing template on `noexcept`-qualified function types, we pass `NoExcept` as a separate Boolean template parameter, forcing the compiler to generate different mangled names for `noexcept` and non-`noexcept` instantiations.

test/any_sender_of_test.cpp
ispeters: 🤷🏼‍♂️
This change seems good.

ispeters (approved):
LGTM; it'd be nice to see a clang-cl CI job added to keep it working, but this seems mergeable.

ericniebler:
Thanks

PR #431, opened 2022-07-05
[WIP] `task<>` scheduler affinity

### LINKED ISSUE #216: usage of current_scheduler inside of coroutine
Is it expected that next code doesn't work?
```
  timed_single_thread_context ctx;

  auto makeTask = [&]() -> task<int> {
    co_return co_await via(schedule(current_scheduler), just(42));
  };

  sync_wait(with_query_value(makeTask(), get_scheduler, ctx.get_scheduler()));
```

### LINKED ISSUE #341: Passing current_scheduler to on() results in infinite recursion in get_scheduler()

ericniebler:
These are the scheduler affinity changes from the unstable branch, with minor changes. I hope to refine these changes to make them ready for trunk (see below). I will use the review feedback from #290 as a starting point.

Currently, `co_await schedule(sch)` is _magic_; it has the effect of transitioning execution context mid-coroutine, and scheduling a thunk back to the current execution context with `at_coroutine_exit()`. I would like to replace this magic behavior with a sender adaptor called `complete_inline()` that `co_await` recognizes. `co_await complete_inline(schedule(sch))` would do what `co_await schedule(sch)` does today.

#### TODO:
- [ ] Remove `sender_for` and the special behavior of `co_await schedule(sch)`.
- [ ] Add a `complete_inline()` sender adaptor that suppresses the thunk back to the current execution context when `co_await`-ing a sender.
- [ ] Fix #216 
- [ ] Fix #341 
- [ ] Investigate coroutine-related binary bloat issue, which could be related to this change.
- [ ] Solve double-thunking caused by using `blocking` to signal whether to repost or not.


ispeters:
We can close this now, right?

PR #437, opened 2022-07-11
fix `async_scope_test::attach_record_done`

janondrusek:
* inplace execution of `set_done()` causes the test to hang as `just_from()` arm of `when_all()` won't `start()`
* receiver may `set()` the event after `async_wait()` suspends

### DIFF
--- a/test/async_scope_test.cpp
+++ b/test/async_scope_test.cpp
@@ -668,7 +668,9 @@ TEST_F(async_scope_test, attach_record_done) {
 
     void set_done() noexcept {
       auto& localEvt = evt;
-      sync_wait(localEvt.async_wait());
+      sync_wait(when_all(localEvt.async_wait(), just_from([&]() noexcept {
+                           localEvt.set();
+                         })));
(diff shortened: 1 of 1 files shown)


test/async_scope_test.cpp
```diff
@@ -668,7 +668,9 @@ TEST_F(async_scope_test, attach_record_done) {
 
     void set_done() noexcept {
       auto& localEvt = evt;
-      sync_wait(localEvt.async_wait());
+      sync_wait(when_all(localEvt.async_wait(), just_from([&]() noexcept {
+                           localEvt.set();
```
AnujYamdagni, line 672: Is the `evt.set()` on line 681 still necessary?

AnujYamdagni (approved):

test/async_scope_test.cpp
janondrusek: Yes it is. It's there for the `set_value`, `set_error` and `set_done` when not inplace called.

PR #438, opened 2022-07-20
Check receiver earlier to avoid gcc11+ compiler error

### LINKED ISSUE #422: tag_invoke related compiler error with gcc >= 11
I was playing with the following example, and noticed it passes on GCC 10, clang 12,13,14, but fails to compile on GCC 11, 12.

```
    let_done(
        let_error(
                []() -> task<int> { co_return 5; }(),
            [](auto&&) noexcept { // lambda 1
                return just();
            }
        ) ,[]() noexcept { // lambda 2
            return just();
        }
    );
```

https://godbolt.org/z/fG6z49bM5. Note that in this example, I'm only constructing a sender, but not actually consuming (e.g. no `sync_wait`). I believe the error is coming from one of the `is_tag_invocable` evaluations, replicated below:

```
    auto l1 = [](auto&&) noexcept {return just(); }; // lambda 1
    auto l2 = []() noexcept { return just(); }; // lambda 2

    static_assert(
        !is_tag_invocable_v<
            _let_d::_cpo::_fn,
            _let_e::_sndr<_task::_task<int>::type, decltype(l1)>::type,
            decltype(l2)
        >);
```

I think the compiler sees`let_error` sender's [`tag_invoke`](https://github.com/facebookexperimental/libunifex/blob/main/include/unifex/let_error.hpp#L355), eventually checking whether `Receiver = unifex::_let_e::_rcvr<unifex::_task::_task<int>::type&&, main()::<lambda(auto:16&&)>, main()::<lambda()> >::type` is a receiver, which it is not since `main()::<lambda()>` (the second lambda in my example) is not itself a receiver.

The `static_assert` above passes compilation on GCC 10/clang, but fails on GCC 11/12 - but fails in the same way as the original example.

## Narrowing down to a smaller example

I simplified the overall code focusing on the `tag_invoke` portion of the code, agnostic of unifex: https://godbolt.org/z/Ee35h3eYq. This similarly passes compilation with GCC 10 and clang, but not GCC 11/12. GCC10 seems to be less eager and never attempts to instantiate `receiver_type<object>`, but GCC 11/12 more eagerly evaluates the template and fails the compilation while doing so. I can't tell which behavior is correct (compiler bug?).

ccotter:
gcc11 and above seems to be more eager in instantiating templates in
evaluating `is_tag_invocable`, and can fail compilation rather than
evaluating `is_tag_invocable` as false.

Test plan: The new test case fails to compile on gcc 11, 12, but passes
with gcc 10 and clang 13 and 14. The fix to let_error.hpp fixes the test
to compile and pass.

Fixes #422


ccotter:
I'm reasonably confident the implementation of [CWG2369 in GCC](https://github.com/gcc-mirror/gcc/commit/f1612b8ae8a60f62cf5456b3357a341550534a7e) is what triggered the change in GCC. I'm not well versed enough to understand if this is a compiler bug or other defect in evaluation of concept constraints. I don't believe clang has implemented CWG2369 yet, for what it's worth.

ispeters (changes requested):
Sorry about the delay in reviewing this, @ccotter; the changes look good, though, and I'd be happy to merge them after we get the merge conflict resolved.

ispeters:
Thanks for the update!  Let's see how the CI turns out.

ispeters (approved):

ispeters (changes requested):
Oh, shoot, there are new conflicts; can you get them resolved, please?  I _think_ that your CI should run automatically now that I've merged one of your PRs.

ispeters (approved):
Never mind that, I'm apparently just a GitHub maintainer newb.  If I choose "Squash & Merge" (which we want anyway), there will be no conflicts.

ccotter:
thanks @ispeters!

PR #439, opened 2022-07-20
Fix "include what you use" issues

ccotter:


ispeters (changes requested):
Thanks for doing this!  The changes look great but it looks like we've diverged since you submitted the PR (sorry for the delay).  If you can resolve the merge conflict, I'll happily merge this.

ccotter:
For reference, to re-run this check in the future, the following generates standalone translation units that help find include-what-you-use issues. This could be done in the cmake automatically as well, if that's of interest.

```
ls ../include/unifex/ | grep hpp$ | while read header; do printf "#include <unifex/$header>\nint main(){}" > ${header}_test.cpp; done
```

ispeters:
Thanks, this looks lovely.  Let's see how the CI turns out.

> This could be done in the cmake automatically as well, if that's of interest.

That _sounds_ interesting.  Can you elaborate?  If you're proposing a way to add "check IWYU" to our CI, I think that'd be really valuable.

ispeters (approved):

PR #443, opened 2022-07-21
connect_awaitable gcc 10.2 promise fix

ccotter:
The test I added crashed when built with gcc 10.2 due to a bug in this compiler version that passes pre-copy references to the coroutine function arguments rather than post-copy into the coroutine frame. Printing the addresses in gdb show that the references refer to the pre-copy memory addresses of the awaitable and receiver, rather than post-copy (in the coroutine frame).

gcc 10.2's level of coroutine support is definitely not as great as 10.3 or newer major versions, although I thought it was worth it nonetheless to highlight this bug with a potential fix. The fix I have is not all that great since it requires a memory allocation, so I'll definitely take suggestions.

### DIFF
--- a/include/unifex/connect_awaitable.hpp
+++ b/include/unifex/connect_awaitable.hpp
@@ -32,6 +32,15 @@
 
 #include <unifex/detail/prologue.hpp>
 
+#if defined(__GNUC__) && __GNUC__ == 10 && __GNUC_MINOR__ <= 2
+// GCC <= 10.2 appears to pass pre-copy references to the coroutine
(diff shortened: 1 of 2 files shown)


ispeters:
@ccotter, I hesitate to accept a patch like this—I'm inclined to think the gcc 10.2 bug means we have to raise our minimum-supported version for `task<>` to 10.3.  What do you think about that?  I could imagine updating the configure script to set `UNIFEX_NO_COROUTINES` to `1` (i.e. disable coroutine support) for gcc 10.2 and earlier.

ccotter:
Closing per above (better to document supported versions, which has an open PR in #466)

PR #447, opened 2022-08-05
fix `detach_on_cancel_test.cancellation_and_completion_race`

janondrusek:
* rewrite with sender algorithm composition
* allows for CI w/ TSAN enabled

### DIFF
--- a/test/detach_on_cancel_test.cpp
+++ b/test/detach_on_cancel_test.cpp
@@ -2,7 +2,6 @@
 #include <unifex/async_scope.hpp>
 #include <unifex/detach_on_cancel.hpp>
 #include <unifex/get_stop_token.hpp>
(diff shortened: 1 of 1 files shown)


doc/api_reference.md
```diff
@@ -236,11 +237,33 @@ caught and passed to the receiver's `set_error` with `std::current_exception()`.
...
+then(some_operation(),
+    [](auto& x) {
+      return func(x);
+    });
+```
+is roughly equivalent to the following coroutine code:
+```c++
+{
+  auto x = co_await some_operation();
+  func(x);
```
ispeters, line 263: I think we should include a `return` statement here.
```suggestion
  return func(x);
```

ispeters (approved):
LGTM

PR #460, opened 2022-09-13
move `unifex::async_scope` into the inline `v1` namespace

janondrusek:
* full name is `unifex::v1::async_scope`
* still can be referred to as `unifex::async_scope`


janondrusek:
This change is in preparation for introducing a new `unifex::v2::async_scope` with a different API.

Implementation is moved from `<unifex/async_scope.hpp>` to `<unifex/v1/async_scope.hpp>`. The old
header includes the new one for backwards compatibility.

ispeters (changes requested):
Looks like you missed the `inline namespace v1`.

ispeters (approved):
LGTM

PR #461, opened 2022-09-16
add `async_auto_reset_event`

janondrusek:
* backed by `async_manual_reset_event`
* offers `Stream` interface

### DIFF
--- a/include/unifex/async_auto_reset_event.hpp
+++ b/include/unifex/async_auto_reset_event.hpp
@@ -1,0 +1,175 @@
+// (c) Meta Platforms, Inc. and affiliates. Confidential and proprietary.
+
+#pragma once
+
+#include <unifex/async_manual_reset_event.hpp>
+#include <unifex/defer.hpp>
+#include <unifex/just_done.hpp>
+#include <unifex/just_void_or_done.hpp>
+#include <unifex/let_value.hpp>
+#include <unifex/let_value_with.hpp>
+#include <unifex/let_value_with_stop_token.hpp>
+#include <unifex/sender_concepts.hpp>
+
+#include <mutex>
+
+namespace unifex {
+
+namespace _aare {
+/**
+ * A *Stream* backed by an auto-reset event.
+ */
+struct async_auto_reset_event final {
+  /**
+   * Initializes an async_auto_reset_event in the unset state.
+   */
+  async_auto_reset_event() noexcept : async_auto_reset_event(false) {}
+
+  /**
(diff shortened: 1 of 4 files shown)


include/unifex/async_auto_reset_event.hpp
```diff
@@ -0,0 +1,150 @@
...
+   */
+  stream_view stream() noexcept;
+
+private:
+  enum class state { UNSET, SET, DONE };
+  /**
+   * Alternatives to using `mutex`:
+   *   1. "2 stage commit" with `atomic<state>`, requires re-try
+   *   2. drop `set_done()` support, undesirable
+   * @see discussion on D34156868 for reasoning.
```
ispeters, line 64: This isn't super useful upstream.  Assuming it can be made public, it would be better to inline the reasoning here.

include/unifex/async_auto_reset_event.hpp
```diff
@@ -0,0 +1,150 @@
...
+            auto stopCallback = [evt]() noexcept {
+              evt->set_done();
+            };
+
+            using stop_token_t = unifex::remove_cvref_t<decltype(stopToken)>;
+            using stop_callback_t =
+                typename stop_token_t::template callback_type<
+                    decltype(stopCallback)>;
+
+            return stop_callback_t{stopToken, stopCallback};
```
ispeters, line 117: I think we've learned this is a non-functional way to do stop callbacks, right?  There's probably a lifetime issue with synchronous cancellation here.

ispeters (changes requested):
Couple of comments below.

include/unifex/async_auto_reset_event.hpp
ispeters: We looked at this in detail together the other day and decided it's fine:
 - when cancelling on the same _Scheduler_ that's waiting on `async_wait()`:
    - ignoring inline _Schedulers_, the stop callback never synchronously wakes the event because `evt->set_done()` always schedules the wakeup
    - accounting for inline _Schedulers_: stop hitting yourself
 - when cancelling from a different _Scheduler_ than the one that's waiting on `async_wait()`:
    - there's technically a race between completing the stop callback and completing the `async_wait()` _but_ the `let_value_with` operation state will not complete the downstream _Receiver_ until the stop callback has been destroyed, which will synchronize with the completion of the callback; given this scenario is "cancel and complete on different _Schedulers_", the synchronization will block rather than no-op

ispeters (approved):
lgtm

PR #464, opened 2022-10-02
Use tag_t in most places

ccotter:
Most places use `tag_t` to obtain the type of CPOs. This commit changes the few spots where decltype is used on CPO types like `set_value` or `connect`.


ispeters (approved):
Nice!  Thanks!

PR #469, opened 2022-10-12
Support lvalue let_value_with senders

ispeters:
The existing operation state for `let_value_with` assumes that its *Sender* will always be an rvalue; this diff fixes the operation state's constructor to support lvalue *Senders*.


janondrusek:
Should we add a regression test?

ispeters:
> Should we add a regression test?

Yes, but I want to do that in a separate PR; there are no tests at all for `let_value_with` and I'm only pushing this PR to enable a particular test in the upcoming `spawn_detached` test suite.

janondrusek (approved):

PR #470, opened 2022-10-12
Introduce spawn_detached(sender, scope, allocator)

ispeters:
This diff introduces a new algorithm, `unifex::spawn_detached()`. `spawn_detached` takes a sender, an "async scope", and an optional allocator.  It nests the sender in the scope with `unifex::nest`, allocates an operation state using the allocator, and starts that operation.  The given async scope may be anything that `nest()` supports, which currently includes both `v1::async_scope` and `v2::async_scope`.

### DIFF
--- a/test/spawn_detached_test.cpp
+++ b/test/spawn_detached_test.cpp
@@ -63,7 +63,7 @@ TEST(spawn_detached_test, spawn_detached_increments_use_count) {
   bool lambdaHasExecuted{false};
 
   unifex::spawn_detached(
-      unifex::just_from([&, this]() noexcept {
+      unifex::just_from([&]() noexcept {
         EXPECT_EQ(1, scope.use_count());
         lambdaHasExecuted = true;
       }),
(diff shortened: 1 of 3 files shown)


test/spawn_detached_test.cpp
```diff
@@ -0,0 +1,188 @@
...
+        connected = true;
+        throw 42;
+      },
+      [&](auto&) noexcept {
+        // this successor factory doesn't run until start
+        started = true;
+        return unifex::just();
+      });
+
+  try {
```
janondrusek, line 159: We should figure out if `UNIFEX_NO_EXCEPTIONS` is supported. Maybe this test should be conditional?

test/spawn_detached_test.cpp
```diff
@@ -0,0 +1,188 @@
...
+  EXPECT_TRUE(didExecute);
+}
+
+TEST(spawn_detached_test, spawn_detached_increments_use_count) {
+  unifex::v2::async_scope scope;
+
+  bool lambdaHasExecuted{false};
+
+  unifex::spawn_detached(
+      unifex::just_from([&, this]() noexcept {
```
janondrusek:
```suggestion
      unifex::just_from([&]() noexcept {
```

this may fix the CI

janondrusek (approved):

PR #474, opened 2022-10-20
fix `variant_sender` blocking kind

janondrusek:


include/unifex/variant_sender.hpp
```diff
@@ -71,22 +71,27 @@ struct _op<Ops...>::type {
...
-        (First == blocking_kind::always &&
-        Second == blocking_kind::always_inline) ||
-        (Second == blocking_kind::always &&
-        First == blocking_kind::always_inline)) {
-      return max_blocking_kind<blocking_kind::always, Rest...>{}();
+        first == blocking_kind::always_inline &&
+        second == blocking_kind::always) {
+      return max_blocking_kind<Second, Rest...>{}();
     } else {
       return blocking_kind::maybe;
```
lewissbaker, line 96: Do you also want to handle the case where First and Second are both 'never' to map to max_blocking_kind<never, Rest...>?

include/unifex/variant_sender.hpp
```diff
@@ -139,7 +144,7 @@ class _sender<Senders...>::type {
   }
 
   friend constexpr auto tag_invoke(tag_t<blocking>, const type&) noexcept {
-    return _variant_sender::max_blocking_kind<cblocking<Senders>...>{}();
+    return _variant_sender::max_blocking_kind<Senders...>{}();
```
lewissbaker, line 147: Should this handle the case where the constexpr blocking_kind is 'maybe' to then go on to individually query the blocking kind of the selected variant type using `visit`?
e.g.
```c++
constexpr auto static_blocking = _variant_sender::max_blocking_kind<Senders...>{}();
if constexpr (static_blocking == blocking_kind::maybe) {
  return visit(blocking, sender.senderVariant_);
} else {
  return static_blocking;
}
```

PR #475, opened 2022-11-04
introduce `unifex::async_resource`

janondrusek:
- async_destroy CPO
- returns a ptr-like object


include/unifex/async_destroy.hpp
```diff
@@ -0,0 +1,50 @@
...
+inline constexpr struct _destroy_fn {
+private:
+  template <typename Resource>
+  static auto try_invoke_member(Resource& resource, int) noexcept
+      -> decltype(resource.destroy()) {
+    return resource.destroy();
+  }
+
+  template <typename Resource>
+  [[deprecated("noop async_destroy")]] static auto
```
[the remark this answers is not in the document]
ispeters: We had an offline discussion; you pointed out that this deprecation warning would fire if you _try and fail_ to customize `async_destroy()`, which seems like a plausibly-good thing, but then the error message should be more helpful, I think.
```suggestion
  [[deprecated("Can't async_destroy an object with no async_destroy customization; add a no-op if that's what's intended.")]] static auto
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+
+      while (*tail) {
+        tail = &(*tail)->next_;
+      }
+
+      *tail = adopted;
+    }
+  }
+
+  unifex::task<void> destroy_resource() noexcept {
```
ispeters, line 109: We should probably add comments to the methods in this type to keep their intentions straight.  I remember being confused as we debugged this what the differences were between `destroy_resource`, `destruct_container`, `destroy_this_resource`, etc.

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+
+  auto destruct_impl() noexcept {
+    return unifex::let_value(destructionEvt_.async_wait(), [this]() noexcept {
+      bool just = true;
+      switch (state()) {
+        case state_t::ALLOCATED:
+          // scheduled destruct moved the state from ALLOCATED
+          UNIFEX_ASSERT(false);
+          break;
+        case state_t::CONSTRUCTED: just = false; break;
```
ispeters, line 149: I don't like whatever clang-format rule is putting all of this on one line :(

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+    , schedule_destruct_(schedule_destruct)
+    , destroy_this_resource_(destroy_this_resource)
+    , deleter_(deleter)
+    , children_(nullptr)
+    , next_(nullptr)
+    , parent_(nullptr)
+    , state_(state_t::ALLOCATED) {}
+
+  state_t state() const noexcept {
+    // TODO optimize `memory_order_seq_cst`
```
ispeters, line 185: What do you think?  Now or later?

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+
+class async_scope_ref final {
+  unifex::v1::async_scope* scope_;
+  _container_base* container_;
+
+public:
+  /*implicit*/ async_scope_ref(unifex::v1::async_scope& scope) noexcept
+    : async_scope_ref(scope, nullptr) {}  // unmanaged scope
+
+  async_scope_ref(
```
ispeters, line 251: The style, inherited from the Committee's recommendations for the STL, is for this to be an `explicit` constructor.
```suggestion
  explicit async_scope_ref(
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+  friend bool operator!=(async_scope_ref lhs, async_scope_ref rhs) noexcept {
+    return !(lhs == rhs);
+  }
+};
+
+template <typename Resource, typename Scheduler>
+struct _container final : _container_base {
+  _container(Scheduler scheduler) noexcept
```
ispeters, line 279:
```suggestion
  explicit _container(Scheduler scheduler) noexcept(std::is_nothrow_move_constructible_v<Scheduler>)
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+        scheduler_,
+        unifex::just_from([factory = std::forward<ResourceFactory>(factory),
+                           this]() mutable noexcept {
+          // note for upstreaming: scope_ construction may succeed, but
+          // resource_ construction may fail
+          scope_.construct();
+          resource_.construct_with([&]() noexcept {
+            return factory(async_scope_ref{scope_.get(), this}, scheduler_);
+          });
+          return set_constructed();
```
ispeters, line 298: We're now upstreaming… We'll have to make `noexcept` conditional and handle exceptions.
```suggestion
                           this]() mutable noexcept(std::is_nothrow_invocable_v<ResourceFactory&, async_scope_ref, Scheduler&>) {
          scope_.construct();

          scope_guard cleanup = [&]() noexcept {
            scope_.destruct();

            // not actually sure what to do here... can we let
            // the exception bubble out?
          };

          resource_.construct_with([&]() {
            return factory(async_scope_ref{scope_.get(), this}, scheduler_);
          });
          
          cleanup.release();

          return set_constructed();
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+struct _container final : _container_base {
+  _container(Scheduler scheduler) noexcept
+      : _container_base{destruct_this_resource,  //
+                        schedule_destruct,       //
+                        destroy_this_resource,   //
+                        deleter},
+        scheduler_(scheduler) {}
+
+  template <typename ResourceFactory>
+  auto construct(ResourceFactory&& factory) noexcept {
```
ispeters, line 287: This `noexcept` is probably wrong.   We need at least `std::is_nothrow_constructible_v<remove_cvref_t<ResourceFactory>, ResourceFactory>`, but that does capture the copy of `scheduler_` and whatever `on()` does.  Maybe something like
```suggestion
  auto construct(ResourceFactory&& factory) noexcept(noexcept(unifex::on(scheduler_, unifex::just_from([factory = std::forward<ResourceFactory>(factory)]() {})))) {
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+};
+
+template <typename Resource, typename Scheduler>
+struct _container final : _container_base {
+  _container(Scheduler scheduler) noexcept
+      : _container_base{destruct_this_resource,  //
+                        schedule_destruct,       //
+                        destroy_this_resource,   //
+                        deleter},
+        scheduler_(scheduler) {}
```
ispeters, line 284:
```suggestion
        scheduler_(std::move(scheduler)) {}
```

include/unifex/async_destroy.hpp
```diff
@@ -0,0 +1,50 @@
...
+namespace unifex {
+namespace _async_destroy {
+
+inline constexpr struct _destroy_fn {
+private:
+  template <typename Resource>
+  static auto try_invoke_member(Resource& resource, int) noexcept
+      -> decltype(resource.destroy()) {
+    return resource.destroy();
+  }
```
ispeters, line 30: Should we permit throwing `destroy()` methods?

include/unifex/async_destroy.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+public:
+  template <typename Resource>
+  constexpr auto operator()(Resource& resource) const noexcept {
+    if constexpr (unifex::tag_invocable<_destroy_fn, Resource&>) {
+      return unifex::tag_invoke(_destroy_fn{}, resource);
+    } else {
+      return try_invoke_member(resource, 1);
+    }
+  }
```
ispeters, line 46: If we make `try_invoke_member` conditionally-`noexcept` then this probably needs a rewrite.

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+    DESTRUCT_SPAWNED,
+    CONSTRUCTED,
+    DESTROYING,
+  };
+
+  void (*destruct_this_resource_)(_container_base*) noexcept;
+  unifex::any_sender_of<> (*schedule_destruct_)(
+      _container_base*, _container_base*) noexcept;
+  unifex::any_sender_of<> (*destroy_this_resource_)(_container_base*) noexcept;
+  void (*deleter_)(_container_base*) noexcept;
```
ispeters, line 74: Can these all be `noexcept`?  I think `deleter_` and `destroy_this_resource_` (or `destruct_this_resource_`?) might be since they're basically type-erased destructors.  `schedule_destruct_` is probably a throwing function since `any_sender_of<>` allocates.

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+  _container_base* next_;
+  _container_base* parent_;
+  // means destruction should start
+  unifex::async_manual_reset_event destructionEvt_;
+  std::mutex mutex_;
+  std::atomic<state_t> state_;
+
+  unifex::task<void> close_child_scopes() noexcept {
+    auto* adopted = [&]() noexcept {
+      std::lock_guard lock{mutex_};
```
ispeters, line 88: Technically, locking a mutex can throw an exception.  The only cases I'm aware of would be bugs here—maybe we _should_ terminate if this throws.  What do you think?

include/unifex/async_resource.hpp
ispeters, line 109: Since coroutines allocate, this probably shouldn't be `noexcept`.
```suggestion
  unifex::task<void> destroy_resource() {
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+          if (!parent_) {
+            deleter_(this);
+          }
+          break;
+        case state_t::DESTROYING:
+          // parent tearing down
+          break;
+      }
+      return unifex::just_void_or_done(just) |
+          unifex::let_done([this]() noexcept {
```
ispeters, line 160: This `noexcept` should probably be conditional.
```suggestion
          unifex::let_done([this]() noexcept(noexcept(unifex::let_done(async_destroy(*this), unifex::just))) {
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+public:
+  /*implicit*/ async_scope_ref(unifex::v1::async_scope& scope) noexcept
+    : async_scope_ref(scope, nullptr) {}  // unmanaged scope
+
+  async_scope_ref(
+      unifex::v1::async_scope& scope, _container_base* container) noexcept
+    : scope_(&scope)
+    , container_(container) {}
+
+  void register_child(_container_base* child) noexcept {
```
ispeters, line 256: `spawn_detached` allocates and thus throws.
```suggestion
  void register_child(_container_base* child) {
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+          scope_.construct();
+          resource_.construct_with([&]() noexcept {
+            return factory(async_scope_ref{scope_.get(), this}, scheduler_);
+          });
+          return set_constructed();
+        }));
+  }
+
+  template <typename ResourceFactory>
+  auto construct_as_sender(ResourceFactory&& factory) noexcept {
```
ispeters, line 303: Same as above, this `noexcept` is wrong.

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+    return unifex::on(
+        scheduler_,
+        unifex::defer([factory = std::forward<ResourceFactory>(factory),
+                       this]() mutable noexcept {
+          // note for upstreaming: scope_ construction may succeed, but
+          // resource_ construction may fail
+          scope_.construct();
+          return unifex::then(
+              factory(async_scope_ref{scope_.get(), this}, scheduler_),
+              [&](auto&&... args) noexcept {
```
ispeters, line 313: Conditional `noexcept`….

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+    return unifex::sequence(
+        unifex::just_from([&self, parent]() noexcept {
+          [[maybe_unused]] auto old = self.state_.exchange(
+              state_t::DESTRUCT_SPAWNED, std::memory_order_seq_cst);
+          UNIFEX_ASSERT(old == state_t::ALLOCATED);
+          if (parent) {
+            parent->register_child(&self);
+          }
+        }),
+        unifex::unstoppable(unifex::on(self.scheduler_, self.destruct_impl())));
```
ispeters, line 357: This probably copies `scheduler_` so it might throw, plus `destruct_impl()`, maybe?

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+  static unifex::any_sender_of<>
+  destroy_this_resource(_container_base* base) noexcept {
```
ispeters, line 361: This probably throws given `any_sender_of<>` allocates…

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+template <typename T>
+struct _type_t {
+  using type = T;
+};
+
+template <
+    typename Resource = void,
+    typename Scheduler,
+    typename ResourceFactory>
+auto async_resource(
```
ispeters, line 386: Oh, you changed the name; wasn't this `make_async_resource`?  Why drop the `make`?  To me, it paralleled `make_unique` and `make_shared`, which produce `unique_ptr` and `shared_ptr`, respectively, so `make_async_resource -> async_resource_ptr` fits the pattern.

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+    Scheduler scheduler,
+    async_scope_ref outerScope,
+    ResourceFactory&& factory) noexcept {
```
ispeters, line 389: This is certainly not `noexcept.`
```suggestion
    ResourceFactory&& factory) {
```

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+          "Sender returning ResourceFactory must specify Resource");
+      return _type_t<Resource>{};
+    } else {
+      static_assert(
+          unifex::same_as<Resource, void>, "must not specify Resource");
+      return _type_t<factory_result_t>{};
+    }
+  };
+  using resource_t = typename decltype(result())::type;
+  auto* container = new _container<resource_t, Scheduler>{scheduler};
```
ispeters, line 405: Perhaps we should leave a `// TODO` to support allocators…

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+  auto construct = [&]() noexcept {
```
ispeters, line 406: Probably best to leave off the `noexcept`.
```suggestion
  auto construct = [&]() {
```

ispeters (commented):
More later…

include/unifex/async_resource.hpp
```diff
@@ -0,0 +1,423 @@
...
+      return std::exchange(children_, nullptr);
+    }();
+    {
+      for (auto* child = adopted; child != nullptr; child = child->next_) {
+        co_await child->close_scope();
+      }
+    }
+    {
+      std::lock_guard lock{mutex_};
+      // TODO repeat when children_ != nullptr
```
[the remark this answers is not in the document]
janondrusek: Stale, `TODO`.

include/unifex/async_resource.hpp
janondrusek: #485

include/unifex/async_resource.hpp
janondrusek: Offline sync: we keep `make_`.

PR #477, opened 2022-11-22
Disable continuation visitations by default for binary size savings

jesswong:
~45 kB binary size savings based off a sample app:

Compressed  : -7.0 KiB (-0.16%) => 4.37 MiB
Uncompressed: -44.5 KiB (-0.45%) => 9.67 MiB


jesswong:
async_trace is failing to build because the type is missing for continuation info. will fix the build errors.

jesswong:
Ran make && make test when UNIFEX_ENABLE_CONTINUATION_VISITATIONS is true and verified everything still builds and pass.

examples/async_trace.cpp
```diff
@@ -47,8 +47,12 @@ auto dump_async_trace(std::string tag = {}) {
       [tag = std::move(tag)](const std::vector<async_trace_entry> &entries) {
         std::cout << "Async Trace (" << tag << "):\n";
         for (auto &entry : entries) {
+        #if UNIFEX_ENABLE_CONTINUATION_VISITATIONS
           std::cout << " " << entry.depth << " [-> " << entry.parentIndex
                     << "]: " << entry.continuation.type().name() << " @ 0x";
+        #else
+          std::cout << "Continuation visitations must be enabled to see types in dump async trace\n";
+        #endif
```
ispeters, line 55: Could we cut a little less deeply?
```suggestion
          std::cout << " " << entry.depth << " [-> " << entry.parentIndex
                    << "]: "
        #if UNIFEX_ENABLE_CONTINUATION_VISITATIONS
                    << entry.continuation.type().name()
        #else
                    << "(continuation typenames disabled)"
        #endif
                    << " @ 0x";
```

ispeters (changes requested):
Looks good, but let's move the definition of `UNIFEX_ENABLE_CONTINUATION_VISITATIONS`.

ispeters (approved):
LGTM

PR #483, opened 2023-01-24
Fix the constraints in let_value_with_stop_token

ispeters:
When trying to nest a `let_value_with_stop_token()` *Sender* inside a `unifex::v2::async_scope`, I discovered there were problems checking the constraints on a `const Sender&` when `Sender` was a `let_value_with_stop_token` *Sender* constructed with a `mutable` lambda.

This diff fixes the above scenario and includes a regression test.  The fixes include:
 - fixing the type calculations on `SuccessFactory` (e.g. we invoke an lvalue reference not an rvalue reference)
 - correcting the `noexcept` calculations (e.g. although we *construct* our copy of the `SuccessorFactory` in a value category-sensitive way, we always *invoke* the factory as a mutable lvalue)
 - standardizing value category casts (i.e. using `std::move()` for moves and `static_cast<DestType&&>()` for forwards)
 - adding some `static_assert`s to confirm the validity of the above fixes
 - made the operation state immovable

### DIFF
--- a/include/unifex/let_value_with_stop_token.hpp
+++ b/include/unifex/let_value_with_stop_token.hpp
@@ -121,9 +121,13 @@ using stop_token_sender = typename _stop_token_sender<SuccessorFactory>::type;
 
 template <typename SuccessorFactory>
 class _stop_token_sender<SuccessorFactory>::type {
+  static_assert(
+      !std::is_reference_v<SuccessorFactory>,
+      "SuccessorFactory should be a value, not a reference");
+
 public:
   using inner_sender_t =
-      std::invoke_result_t<SuccessorFactory, inplace_stop_token>;
+      std::invoke_result_t<SuccessorFactory&, inplace_stop_token>;
 
   template <
       template <typename...>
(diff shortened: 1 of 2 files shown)


include/unifex/let_value_with_stop_token.hpp
```diff
@@ -194,13 +202,17 @@ struct _stop_token_operation<
...
         inplace_stop_token st,
-        Receiver&& r) noexcept(successor_is_nothrow&&
-                                   inner_receiver_nothrow_constructible&&
-                                       nothrow_connectable) {
-      return unifex::connect(func(st), receiver_t{this, st, (Receiver &&) r});
+        // we need to take r by reference to avoid problems at the call site
+        // related to unsequenced function argument evaluation
+        Receiver2&& r) noexcept(successor_is_nothrow&&
+                                    inner_receiver_nothrow_constructible&&
+                                        nothrow_connectable) {
```
janondrusek, line 213: Is `std::is_nothrow_constructible_v<Receiver, Receiver2>` not needed here?

janondrusek (approved):
q about `noexcept()`

include/unifex/let_value_with_stop_token.hpp
ispeters: `inner_receiver_nothrow_constructible` computes that for the old cold; I'll need to update it.

PR #484, opened 2023-01-27
Add an internal receiver to v2::async_scope's nest op

ispeters:
While writing `unifex::spawn_future()`, I discovered that waiting until the destructor of the `v2::async_scope`'s `nest()` operation to drop the scope reference is too late (it led to hangs).

This diff adds an internal receiver to the nest operation so we can detect when the operation is complete (which is likely before the operation state is destroyed) and drop the reference as soone as we reach that state.

### DIFF
--- a/include/unifex/v2/async_scope.hpp
+++ b/include/unifex/v2/async_scope.hpp
@@ -272,14 +272,19 @@ struct _nest_receiver<Sender, Receiver>::type final {
 
   template <typename Func>
   void complete(Func func) noexcept {
-    // keep a strong reference on the scope until this function returns
-    auto scope = std::move(op_->scope_);
+    // save this->op_ into a local because we're about to destroy the current
+    // object, invalidating the this pointer
+    auto op = op_;
 
-    // we're done with the inner operation
-    op_->op_.destruct();
+    // keep a strong reference on the scope until this function returns
+    auto scope = std::move(op->scope_);
+
+    // we're done with the inner operation; note: this call destroys the current
+    // object, which is why we've saved this->op_ into a local
+    op->op_.destruct();
 
     // from here the current object may be destroyed
-    func(std::move(op_->receiver_));
+    func(std::move(op->receiver_));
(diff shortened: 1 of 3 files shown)


include/unifex/v2/async_scope.hpp
```diff
@@ -191,55 +191,102 @@ struct _nest_op final {
...
+
+template <typename Sender, typename Receiver>
+struct _nest_receiver<Sender, Receiver>::type final {
+  nest_op<Sender, Receiver>* op_;
+
+  template <typename... T>
+  void set_value(T... values) noexcept {
+    auto op = op_;
+
+    op->cleanup();
```
ispeters, line 257: `op->cleanup()` destroys the nested operation state _and drops the scope reference_.

Perhaps the scope reference should be dropped later?  If one of `values...` is a pointer to a thing protected by the scope, the pointer could dangle by the time `op->receiver_` observes it.  Maybe `cleanup()` should return-by-move the operation's `scope_ref` so that we can drop the reference after  `set_value()` returns?

I might be over thinking this.

include/unifex/v2/async_scope.hpp
```diff
@@ -191,55 +191,102 @@ struct _nest_op final {
...
+    UNIFEX_CATCH(...) {
+      unifex::set_error(std::move(op->receiver_), std::current_exception());
+    }
+  }
+
+  template <typename E>
+  void set_error(E e) noexcept {
+    auto op = op_;
+
+    op->cleanup();
```
ispeters, line 271: Similar comment here as above; if `e` is a pointer to something protected by scope, it could dangle by the time it's published.

include/unifex/v2/async_scope.hpp
ispeters: After chatting with Ján, we've decided I should defer dropping the reference until after we're done invoking completion methods.

janondrusek (approved):

PR #486, opened 2023-02-10
Mark `ErrorValue` as l-value in `let_error`

janondrusek:
- aligns with the expectation of a final `Receiver`
- add repro test case, which otherwise does not compile due to type mismatch

### DIFF
--- a/include/unifex/let_error.hpp
+++ b/include/unifex/let_error.hpp
@@ -35,21 +35,21 @@ namespace unifex {
 
 namespace _let_e {
 template <typename Source, typename Func, typename Receiver>
-struct _op {
+struct _op final {
   class type;
 };
 template <typename Source, typename Func, typename Receiver>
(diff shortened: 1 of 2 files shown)


include/unifex/let_error.hpp
```diff
@@ -322,7 +322,7 @@ template<typename... Senders>
 using any_sends_done = std::disjunction<sends_done_impl<Senders>...>;
 
 template <typename Source, typename Func>
-class _sndr<Source, Func>::type {
+class _sndr<Source, Func>::type final {
 
   template <typename Error>
   using final_sender = callable_result_t<Func, remove_cvref_t<Error>>;
```
ispeters, line 328:
```suggestion
  using final_sender = callable_result_t<Func, remove_cvref_t<Error>&>;
```

ispeters (changes requested):
I wonder if resolving the comments I've left will address the memory leak.

ispeters (approved):

PR #482, opened 2023-01-24
Auto-format let_value_with_stop_token.hpp

ispeters:
This diff adds some `//` comments to the ends of some lines and then runs the result through `clang-format`.

PR #454, opened 2022-08-17
Upstream when_all_range algorithm

AnujYamdagni:
The when_all_range algorithm is similar to when_all but takes in a dynamic vector of senders.

PR #441, opened 2022-07-20
`CreateTest.AwaitTest` local `this` pointer

janondrusek:
* make ASAN happy
* allows CI w/ coroutines enabled

PR #491, opened 2023-03-22
merge `broken-stdlib` into `main` part 1

janondrusek:
add unifex::create for trivially wrapping a C-style async API

avoid warning about missing braces in initializer

avoid warning about implicit lambda capture of this

Remove the deprecated headers

Add let_value_with_stop_token algorithm (#345)

Adding a new sender algorithm to enable hooking cancellation callbacks onto the receiver's stop token directly.

Make take_until's next_sender a scheduler_provider

This diff changes `take_until`'s `next_sender` to respond to receiver
queries by forwarding them to the outer receiver.  In particular, this
makes the inner receiver a `scheduler_provider`, solving a problem we
have internally.

Add more receiver query customizations

Add detach on cancel (#361)

delete move operators from inplace_stop_token (#386)

- copy constructor makes a trivial copy of a pointer
- avoid leaving source object in an empty state

preserve operation state in let_value_with_stop_token (#388)

* SuccessorFactory should live past unifex::connect
* SuccessorFactory may be destroyed after unifex::start

extract fused_stop_source (#390)

* support manual lifetime of stop callbacks
* use in let_value_with_stop_*

Fix detach_on_cancel error_types propagation (#400)

* Fix detach_on_cancel error_types propagation

* Use unifex::variant instead of std::variant

Avoid tail calls when compiling with the Emscripten (WebAssembly) com… (#403)

* Avoid tail calls when compiling with the Emscripten (WebAssembly) compiler

add nothrow_connectable test for `let_value_with_stop_source()` (#393)

* missing regression test case, like `let_value_with_stop_token()`
* reduce code duplication in `let_value_with_stop_token_test`

Make `type_erased_stream` `Scheduler` aware (#404)

* expose `unifex::type_erased_stream` (result of `unifex::type_erase()`)
* add `get_scheduler` CPO into `type_erased_stream`

Make the when_all sender multiply-connectable (#409)

I discovered that `when_all` doesn't compose with `repeat_effect_until`
because the `when_all` *Sender* can only be connected as an rvalue.
This change introduces a regression test and a fix.

new algorithm unifex::unstoppable (#410)

fixing linter error (#414)

Fix asan error for let_value with coroutines (#416)

* Fix asan error for let_value with coroutines

* Add tests for set_done and set_error coroutine asan errors

* Do just(5) instead of passing in 5 directly to coroutine

* Fix MacOS Optimized test failure with allocate instead of coroutines

Fix let_value to query sender properties through sender_traits (#418)

make io_uring syscalls retrying on interruptions (#419)

Fix the error comparison

Remove extra call

fix cancellation race in `detach_on_cancel()` (#432)

* use ref counting to pass ownership of `detached_state`
* regression tests

add `detach_on_cancel()` to docs (#434)

add `then()` example to docs (#434)

`CreateTest.AwaitTest` local `this` pointer (#441)

* make ASAN happy
* allows CI w/ coroutines enabled

fix `detach_on_cancel_test.cancellation_and_completion_race` (#447)

* rewrite with sender algorithm composition
* allows for CI w/ TSAN enabled

Adding Variant Sender (#397)

* Adding Variant Sender

* Remove deprecated use of is_connectable

* Update noexcept clause and add test cases for different value type categories

* Add documentation for variant_sender

* Address nits

Upstream when_all_range algorithm (#454)

Upstream commit of when_all_range, Similar to when_all but takes in a vector of senders.

use `manual_lifetime` in `variant_sender` (#459)

* construct operation state in place
* regression test

add `async_auto_reset_event` (#461)

* backed by `async_manual_reset_event`
* offers `Stream` interface

Add missing include to when_all_range.hpp (#465)

PR #454 missed including `<unifex/continuations.hpp>` in
`when_all_range.hpp` so let's fix that.

Support lvalue let_value_with senders (#469)

The existing operation state for `let_value_with` assumes that its
*Sender* will always be an rvalue; this diff fixes the operation state's
constructor to support lvalue *Senders*.

Disable continuation visitations by default for binary size savings (#477)

* Disable continuation visitations by default for binary size savings

* Add unifex prefix to macro and fix async_trace

* Fix unused variable error

* Change ordering of field initializations

use stable release `os: ubuntu-20.04` (#481)

* gcc 9,10
* clang 10,11,12

Auto-format let_value_with_stop_token.hpp (#482)

This diff adds some `//` comments to the ends of some lines and then
runs the result through `clang-format`.

Fix the constraints in let_value_with_stop_token (#483)

When trying to nest a `let_value_with_stop_token()` *Sender* inside a
`unifex::v2::async_scope`, I discovered there were problems checking the
constraints on a `const Sender&` when `Sender` was a
`let_value_with_stop_token` *Sender* constructed with a `mutable`
lambda.

This diff fixes the above scenario and includes a regression test.  The
fixes include:
 - fixing the type calculations on `SuccessFactory` (e.g. we invoke an
   lvalue reference not an rvalue reference)
 - correcting the `noexcept` calculations (e.g. although we *construct*
   our copy of the `SuccessorFactory` in a value category-sensitive way,
   we always *invoke* the factory as a mutable lvalue)
 - standardizing value category casts (i.e. using `std::move()` for
   moves and `static_cast<DestType&&>()` for forwards)
 - adding some `static_assert`s to confirm the validity of the above
   fixes
 - made the operation state immovable

Mark `ErrorValue` as l-value in `let_error` (#486)

- aligns with the `let_value` algorithm, which keeps value (in case of
  `let_error` the `ErrorValue`) produced by a predecessor alive for
the duration of a successor operation
- add repro test case, which otherwise does not compile due to type
  mismatch


include/unifex/fused_stop_source.hpp
```diff
@@ -0,0 +1,84 @@
...
+                       stop_callback>)
+    : callback_(std::move(first), stop_callback{source})
+    , rest_(source, std::move(rest)...) {}
+
+  UNIFEX_NO_UNIQUE_ADDRESS First callback_;
+  UNIFEX_NO_UNIQUE_ADDRESS fused_stop_callback<Rest...> rest_;
+};
+
+template <typename... StopTokens>
+struct fused_stop_source : unifex::inplace_stop_source {
```
ispeters, line 65: We ended up deciding this doesn't work the way we hoped (there's some need for reference counting, or something).  I _think_ we're using it in some other algorithm nonetheless.  If this is unused, it's probably better not to port it to `main`, though.

include/unifex/let_value_with_stop_source.hpp
```diff
@@ -16,6 +16,7 @@
 #pragma once
 
 #include <unifex/config.hpp>
+#include <unifex/fused_stop_source.hpp>
```
ispeters, line 19: I guess we're using `fused_stop_source` here. :(

include/unifex/sender_concepts.hpp
```diff
@@ -285,19 +296,19 @@ using operation_t [[deprecated("Use connect_result_t instead of operation_t")]]
 template <typename Sender, typename Receiver>
 [[deprecated("Use sender_to instead of is_connectable_v")]]
 inline constexpr bool is_connectable_v =
-  is_callable_v<tag_t<connect>, Sender, Receiver>;
+  is_callable_v<decltype(connect), Sender, Receiver>;
```
ispeters, line 299: I wonder why we made this change.

include/unifex/sender_for.hpp
```diff
@@ -0,0 +1,116 @@
...
+          noexcept(
+              std::is_nothrow_constructible_v<remove_cvref_t<Sender>, Sender> &&
+              (std::is_nothrow_constructible_v<remove_cvref_t<Queries>, Queries> &&...)) {
+        return sender_for<CPO, remove_cvref_t<Sender>, _context<remove_cvref_t<Queries>...>>{
+            (Sender&&) snd, _context<remove_cvref_t<Queries>...>{{(Queries&&) queries}...}};
+      }
+    };
+  } // namespace _sf
+
+  using _sf::sender_for;
```
ispeters, line 99: This exists only in support of _Scheduler_ affinity; it might make sense to defer porting this file until you port that feature.

include/unifex/variant_sender.hpp
```diff
@@ -0,0 +1,125 @@
...
+    noexcept(std::conjunction_v<unifex::is_nothrow_connectable<match_reference_t<This, Senders>, Receiver>...>)
+  {
+    return std::visit(
+        [&r](auto&& sender) noexcept(
+            unifex::is_nothrow_connectable_v<decltype(sender), Receiver>) {
+          return operation<connect_result_t<Senders, Receiver>...>{
+             static_cast<decltype(sender)&&>(sender), static_cast<Receiver&&>(r)};
+        },
+        std::move(static_cast<decltype(that)>(that).senderVariant_));
+  }
```
ispeters, line 117: The bug blocking my `spawn_future` PR on MSVC in C++20 mode is in this section.  I suppose it's fine to port this to `main` right now and then port my fix on top later, but you could consider deferring porting `variant_sender` until after we get all the know bugs fixed.

ispeters (approved):
LGTM with some thoughts for you to consider.

include/unifex/sender_for.hpp
janondrusek: Removed.

include/unifex/sender_concepts.hpp
janondrusek: `main` was updated to use `tag_t`, I'll keep that

PR #493, opened 2023-03-24
fix on() tag_invoke overload

janondrusek:
- arguments in the concept were swapped
- add a regression test

### DIFF
--- a/test/on_test.cpp
+++ b/test/on_test.cpp
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


test/on_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <gtest/gtest.h>
+#include <thread>
+
+using namespace unifex;
+
+namespace {
+struct Customized {
+  std::thread::id id2 = std::this_thread::get_id();
+  template <
+
```
ispeters, line 31: Nit.
```suggestion
```

test/on_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+TEST(On, Smoke) {
+  auto id1 = std::this_thread::get_id();
+  auto id2 = std::this_thread::get_id();
+
+  single_thread_context thread;
+
+  auto result = sync_wait(on(thread.get_scheduler(), just_from([&id2] {
+                               id2 = std::this_thread::get_id();
+                             })));
+  EXPECT_NE(id1, id2);
```
ispeters, line 68: I _think_ `single_thread_context` has a method that returns its thread's ID; you might be able to assert not only that it's different from the "main" thread, but that it's exactly the `single_thread_context`'s thread.

ispeters (approved):
LGTM

PR #494, opened 2023-03-24
make `unstoppable()` a concrete type

janondrusek:
- allows for customizations via tag_invoke

### DIFF
--- a/include/unifex/unstoppable.hpp
+++ b/include/unifex/unstoppable.hpp
@@ -22,16 +22,64 @@
 
 namespace unifex {
 namespace _unstoppable {
-inline const struct _fn {
-  template <typename Sender>
-  constexpr auto operator()(Sender&& sender) const noexcept {
-    return with_query_value(
-        (Sender &&) sender, get_stop_token, unstoppable_token{});
+
+template <typename Sender>
+struct _sender final {
+  struct type;
+};
+
+template <typename Sender>
+using sender = typename _sender<remove_cvref_t<Sender>>::type;
+
+template <typename Sender>
+struct _sender<Sender>::type final {
+  UNIFEX_NO_UNIQUE_ADDRESS Sender sender_;
+
+  template <
+      template <typename...>
+      class Variant,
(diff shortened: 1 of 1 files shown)


include/unifex/unstoppable.hpp
```diff
@@ -22,16 +22,54 @@
...
+  using error_types = sender_error_types_t<Sender, Variant>;
+
+  static constexpr bool sends_done = sender_traits<Sender>::sends_done;
+
+  template <typename Receiver>
+  friend auto tag_invoke(tag_t<connect>, type self, Receiver&& r) noexcept(
+      is_nothrow_connectable_v<Sender, remove_cvref_t<Receiver>>) {
+    return connect(
+        with_query_value(
+            std::move(self).sender_, get_stop_token, unstoppable_token{}),
```
ispeters, line 55: I think this should probably be more constrained, and we should probably take `self` by forwarding reference.
```suggestion
  template(typename Self, typename Receiver) //
      (requires same_as<type, remove_cvref_t<Self>> AND sender_to<member_t<Self, Sender>, Receiver>) //
  friend auto tag_invoke(tag_t<connect>, Self&& self, Receiver&& r) noexcept(
      is_nothrow_connectable_v<member_t<Self, Sender>, remove_cvref_t<Receiver>>) {
    return connect(
        with_query_value(
            static_cast<Self&&>(self).sender_, get_stop_token, unstoppable_token{}),
```

include/unifex/unstoppable.hpp
```diff
@@ -22,16 +22,54 @@
...
+        static_cast<Receiver&&>(r));
+  }
+};
```
ispeters, line 58: You should probably add an implementation of `blocking`.
```suggestion
  friend auto tag_invoke(tag_t<blocking>, const type& s) noexcept {
    return blocking(s.sender_);
  }
};
```

ispeters (approved):
Couple of comments but otherwise LGTM.

PR #495, opened 2023-03-25
Make task<>'s thunk-on-resume unstoppable

ispeters:
When awaiting an async Sender that swallows done signals (such as `let_done(never_sender{}, just)`), the user-level code looks like it swallows done signals:
```
// never cancels
co_await let_done(never_sender{}, just);
```

However, `task<>`'s Scheduler affinity implementation transforms the above code into this:
```
co_await typed_via(let_done(never_sender{}, just), <current scheduler>);
```

The `schedule()` operation inside the injected `typed_via` can emit done if the current stop token has had stop requested, leading to very non-obvious cancellation behaviour that can't be worked around.

This diff introduces a pair of regression tests that capture the above scenario, and the analogous scenario of awaiting an async Awaitable that completes with done.  The next diff will fix these failing tests.

### DIFF
--- a/include/unifex/task.hpp
+++ b/include/unifex/task.hpp
@@ -15,24 +15,25 @@
  */
 #pragma once
 
+#include <unifex/any_scheduler.hpp>
 #include <unifex/async_trace.hpp>
 #include <unifex/await_transform.hpp>
+#include <unifex/blocking.hpp>
 #include <unifex/connect_awaitable.hpp>
-#include <unifex/inplace_stop_token.hpp>
-#include <unifex/inline_scheduler.hpp>
-#include <unifex/manual_lifetime.hpp>
+#include <unifex/continuations.hpp>
 #include <unifex/coroutine.hpp>
 #include <unifex/coroutine_concepts.hpp>
+#include <unifex/finally.hpp>
+#include <unifex/inline_scheduler.hpp>
+#include <unifex/inplace_stop_token.hpp>
+#include <unifex/invoke.hpp>
+#include <unifex/manual_lifetime.hpp>
+#include <unifex/scope_guard.hpp>
 #include <unifex/sender_concepts.hpp>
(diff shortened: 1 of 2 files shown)


lewissbaker:
I don't think that any via algorithm should be allowed to continue if the schedule op doesn't complete with set_value(). Do we should perhaps have the typed_via() algorithm wrap the schedule op in unstoppable()?

ispeters:
Can you elaborate, @lewissbaker? I'm unclear what you think I'm doing that shouldn't be done and I don't think I understand what you're proposing I do instead, either.

include/unifex/task.hpp
```diff
@@ -216,7 +217,7 @@ struct _promise {
       } else {
         return unifex::await_transform(
             *this,
-            typed_via(as_sender((Awaitable&&) awaitable), this->sched_));
+            finally(as_sender((Awaitable&&) awaitable), unstoppable(schedule(this->sched_))));
```
lewissbaker, line 220: Rather than make this change here, I wonder if it would be better to make this change in `typed_via()` so that it applies more widely.

`typed_via()` is trying to move execution onto some other context (like with `task`'s scheduler-affinity).
In many cases, it's a correctness issue if you don't end up continuing on that context.
e.g. maybe the context is a strand and you're relying on destructors running on that context to avoid certain data-races

So it may be worthwhile just saying that the schedule-op that typed_via() incorporates must complete with set_value(), and applying the unstoppable() to the schedule() for all usages of typed_via().

ispeters: Ah, I see.

I was being conservative, but yeah, I think you're right.

ispeters: This reliably deadlocks the `examples/linux/io_epoll_test` program so I'm going to revert to the original strategy.  Once the internal crashing bug is really put to bed, we can investigate whether the example program is just buggy.

PR #496, opened 2023-03-25
Fix signed integer overflow UB in reduce_with_trampoline example

robinlinden:
```
libunifex/examples/reduce_with_trampoline.cpp:40:51: runtime error:
signed integer overflow: 2146177440 + 7464960 cannot be represented in
type 'int'
```

I also ran into issues compiling w/ Clang 16 since `-Werror` is enabled, so there's a fix for that too. I guess Clang has gotten smarter about set-but-not-used variables.

### DIFF
--- a/examples/reduce_with_trampoline.cpp
+++ b/examples/reduce_with_trampoline.cpp
@@ -35,10 +35,10 @@ int main() {
               trampoline_scheduler{},
               transform_stream(
                   range_stream{0, 100'000},
-                  [](int value) { return value * value; })),
+                  [](unsigned value) { return value * value; })),
           0,
-          [](int state, int value) { return state + 10 * value; }),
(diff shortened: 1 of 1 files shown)


robinlinden:
If you want, I'll set up ubsan and clang-16 jobs in your CI as well so these kinds of issues are caught earlier in the future.

robinlinden:
There we go. I know apt.llvm.org can be a bit flaky if you're unlucky. :P

janondrusek:
Mind doing another rebase? Looks like I just merged a PR that touched the same line.

ispeters:
@robinlinden,

> If you want, I'll set up ubsan and clang-16 jobs in your CI as well so these kinds of issues are caught earlier in the future.

That would be really great!

PR #499, opened 2023-03-29
Check concept earlier in let_error

ccotter:
Re-apply #438 to fix compilation on GCC11/12 after merging a subset of the broken-stdlib changes.

### DIFF
--- a/include/unifex/let_error.hpp
+++ b/include/unifex/let_error.hpp
@@ -379,7 +379,8 @@ public:
     typename Receiver,
     typename...,
     typename SourceReceiver = receiver_type<member_t<Sender, Source>, Func, Receiver>)
-      (requires same_as<remove_cvref_t<Sender>, type> AND
+      (requires receiver<Receiver> AND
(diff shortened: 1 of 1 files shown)


ccotter:
@janondrusek - I think #491 accidentally regressed what #438 fixed. The test from #438 is still in place, but the concept checking was removed. GCC 11 and newer requires this due to GCC 11 implementing CWG2369 (clang does not implement it at all). If possible, could you enable a newer GCC 11 or 12 on the CI checks (for that matter, CI is checking only up to clang 12)? Thanks!

janondrusek:
Yep, totally my fault. We need to expedite adding latest compilers to our CI.

janondrusek (approved):

ccotter:
> Yep, totally my fault. We need to expedite adding latest compilers to our CI.

No problem - thanks for the speedy review! Excited to see the broken-stdlib changes merging into main so I don't have to maintain the when_all_range commits in my internal fork :)

PR #500, opened 2023-04-01
Fix stop_when's handling of stop requests

ispeters:
When trying to sync PR #495 into our internal repo, I discovered that there's a lifetime issue in `stop_when()`.  If the `stop_when()` operation receives a stop request from its Receiver and the last-to-finish child operation completes synchronously in response to the stop request then `stop_when()`'s stop callback will access the internal stop source after it's been destroyed.

This first diff just formats `test/stop_when_test.cpp` and `include/unifex/stop_when.hpp` with `clang-format` in preparation for fixing the above problem.

### DIFF
--- a/test/stop_when_test.cpp
+++ b/test/stop_when_test.cpp
@@ -15,11 +15,19 @@
  */
 #include <unifex/stop_when.hpp>
 
+#include <unifex/just_from.hpp>
+#include <unifex/let_value_with_stop_source.hpp>
+#include <unifex/never.hpp>
 #include <unifex/on.hpp>
 #include <unifex/scheduler_concepts.hpp>
 #include <unifex/sync_wait.hpp>
 #include <unifex/then.hpp>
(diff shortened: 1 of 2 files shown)


test/stop_when_test.cpp
```diff
@@ -127,3 +135,17 @@ TEST(StopWhen, Pipeable) {
   EXPECT_FALSE(sourceExecuted);
   EXPECT_TRUE(triggerExecuted);
 }
+
+#if !UNIFEX_NO_COROUTINES
+TEST(StopWhen, SynchronousCompletionFromReceiverStopRequestIsASANSafe) {
```
janondrusek, line 140: Would it repro with `allocate()` too?

janondrusek (approved):

test/stop_when_test.cpp
ispeters: Mysteriously, no.  It took me a long time to narrow this one down, and I can't get it any smaller.

PR #468, opened 2022-10-12
Introduce unifex::nest()

ispeters:
`unifex::nest()` is a CPO that delegates to either a `tag_invoke` customization taking a *Sender* and a "scope" reference, or to a member function on the given scope that takes a *Sender*.  This diff wires `unifex::nest()` to the `nest()` member function on `v2::async_scope` and to the `attach()` member function on `v1::async_scope`.

PR #425, opened 2022-06-13
fix cancellation race in `async_scope::attach*`

janondrusek:
* use refcount to pass ownership of `deliver_result`
* replace `fused_stop_source` with `inplace_stop_source`

PR #504, opened 2023-04-25
Make unifex::async_scope::spawn return a unifex::future<> (#372)

janondrusek:
* Make async_scope::spawn return a lazy<>

This diff changes `unifex::async_scope::spawn(Sender auto)` to return a new type, `unifex::lazy<...>`.  `lazy<>` is a _Sender_ that completes with the result of the _Sender_ given to `spawn`.

* Fix memory leak

* Rationalize the names of some internal bits

* Fuse the promise and the operation state

This diff merges the spawned operation's promise with its operation state so there's only one allocation.

One consequence of this change is that outstanding operations don't record themselves as complete within their corresponding scope until the associated `lazy<>` is either connected and started, or discarded.

* Fix build

Looks like our CI turns on exhaustive switch warnings, which I broke.

* Fix infinite hang in create_test.cpp

* Cleanup

* Last bit of cleanup before bed

* Try to fix ASAN failure

In the light of the morning, I think it's wrong to set the event before requesting stop on the stop source because setting the event could lead to destruction of the operation state, invalidating the stop source.

* Fix request_stop()

The callers of `request_stop()` are not owners so they may not call `decref()`, which means I can't invoke `set_done()` from `request_stop()`.

* Clean up, bug fixes, comments

Among other clean-up, this diff fixes a stack-use-after-return in `future<>`'s `connect()`.

* Add constraints and run clang-format

* Comments, tests, bug fixes, and clang-format

* Restore nothrow assertion in detached_spawn_call_on

* Round out the async_scope tests

add async_scope::attach (#392)

* returned `Sender` needs to be connected and started
* avoids paying penalty of `future<>`

Fix `record_done` ordering in `async_scope::attach` (#424)

* `record_done`, which decrements outstanding operation count, must be called after `set_*`
* add regression test

fix cancellation race in async_scope::attach* (#425)

* use refcount to pass ownership of `deliver_result`
* replace `fused_stop_source` with `inplace_stop_source`

make `async_scope::attached_sender` copyable (#428)

* copy constructor calls `async_scope::try_record_start` internally
* copy of attached Sender will increment outstanding number of operations on async_scope

`async_scope::attach` cleanup (#433)

* remove unused template argument
* add missing test case

update `async_scope` docs (#434)

* spawn() -> detached_spawn()
* spawn() returns a `future`
* add missing public methods

add `async_scope::attach` docs (#434)

fix `async_scope_test::attach_record_done` (#437)

fix `tag_invoke(CPO)` in `async_scope` (#462)

add `unifex::v2::async_scope` (#463)

* simpler than `unifex::v1::async_scope` (`nest()` and `join()`)
* does not support cancellation

Introduce unifex::nest() (#468)

`unifex::nest()` is a CPO that delegates to either a `tag_invoke` customization taking a *Sender* and a "scope" reference, or to a member function on the given scope that takes a *Sender*.  This diff wires `unifex::nest()` to the `nest()` member function on `v2::async_scope` and to the `attach()` member function on `v1::async_scope`.

Introduce spawn_detached(sender, scope, allocator) (#470)

This diff introduces a new algorithm, `unifex::spawn_detached()`. `spawn_detached` takes a sender, an "async scope", and an optional allocator.  It nests the sender in the scope with `unifex::nest`, allocates and operation state using the allocator, and starts that operation.  The given async scope may be anything that `nest()` supports, which currently includes both `v1::async_scope` and `v2::async_scope`.

Add an internal receiver to v2::async_scope's nest op (#484)

While writing `unifex::spawn_future()`, I discovered that waiting until the destructor of the `v2::async_scope`'s `nest()` operation to drop the scope reference is too late (it led to hangs).  This diff adds an internal receiver to the nest operation so we can detect when the operation is complete (which is likely before the operation state is destroyed) and drop the reference as soon as we reach that state.

Fix stop_when's handling of stop requests (#500)

* Fix stop_when's handling of stop requests

When trying to sync PR #495 into our internal repo, I discovered that there's a lifetime issue in `stop_when()`.  If the `stop_when()` operation receives a stop request from its Receiver and the last-to-finish child operation completes synchronously in response to the stop request then `stop_when()`'s stop callback will access the internal stop source after it's been destroyed.

This first diff just formats `test/stop_when_test.cpp` and `include/unifex/stop_when.hpp` with `clang-format` in preparation for fixing the above problem.

* Add a broken unit test

This diff adds a unit test to `test/stop_when_test.cpp` that crashes when ASAN is enabled because it dereferences a destroyed `inplace_stop_source`.

* Increment stop_when's refcount in its stop callback

This diff fixes the broken test from the previous diff by incrementing the `stop_when` operation's refcount while processing a stop request from the receiver.

Add spawn_future() (#489)

This diff adds `unifex::spawn_future(Sender, Scope)`.

Implement async_scope::spawn with spawn_future (#501)

This diff reimplements the `v1::async_scope::spawn()` method in terms of the newly-added `unifex::spawn_future()` algorithm.  I had to delete a few tests that exercise behaviour that's no longer supported.

Work around an MSVC bug in C++20 mode (#492)

* merge unit test that depends on `v2/async_scope`


include/unifex/stop_when.hpp
```diff
@@ -34,359 +34,372 @@
...
+              std::apply(
+                  [&](auto set_xxx, auto&&... args) {
+                    set_xxx(
+                        std::move(receiver_),
+                        static_cast<decltype(args)>(args)...);
+                  },
+                  static_cast<decltype(tuple)>(tuple));
+            } else {
+              // Should be unreachable
+              std::terminate();
```
ispeters, line 247: This kinda bugs me every time I see it; I feel like we should change the `if constexpr` to a `static_assert` and delete this branch.

include/unifex/v1/async_scope.hpp
```diff
@@ -0,0 +1,609 @@
...
+
+#include <atomic>
+#include <memory>
+#include <tuple>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+
+inline namespace v1 {
```
ispeters, line 45: In `main`, the `v0` namespace is `inline`, right?  If so, I think this has to be non-inline.
```suggestion
namespace v1 {
```

test/async_scope_v1_test.cpp
```diff
@@ -0,0 +1,661 @@
...
+    if (destroyed_) {
+      destroyed_->set();
+    }
+  }
+};
+
+}  // namespace
+
+struct async_scope_test : testing::Test {
+  async_scope scope;
```
ispeters, line 69: Perhaps this should be explicit about the version.
```suggestion
  v1::async_scope scope;
```

test/async_scope_v1_test.cpp
```diff
@@ -0,0 +1,661 @@
...
+  EXPECT_EQ(*id, thread.get_thread_id());
+  EXPECT_NE(*id, std::this_thread::get_id());
+}
+
+TEST_F(async_scope_test, lots_of_threads_works) {
+#if !defined(UNIFEX_TEST_LIMIT_ASYNC_SCOPE_THREADS)
+  constexpr int maxCount = 1'000;
+#else
+  constexpr int maxCount = UNIFEX_TEST_LIMIT_ASYNC_SCOPE_THREADS;
+#endif
```
ispeters, line 298: This conditional compilation was added to support an internal project that has since been killed.  Maybe we can simplify.

test/nest_test.cpp
```diff
@@ -0,0 +1,219 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
ispeters, line 2: This probably needs updating once we finish unifying the branches….

ispeters (approved):
I think the inline namespace in v1::async_scope is a blocker, but otherwise lgtm.

include/unifex/stop_when.hpp
janondrusek: #505

test/nest_test.cpp
janondrusek: noted in #485

include/unifex/stop_when.hpp
lewissbaker: This branch is there because one of the variant's alternative types is the empty `tuple<>` and `std::visit` is going to instantiate the lambda with that type as an argument.
This type doesn't have a `set_xxx` cpo type and so can't be passed to the `std::apply` call.

So there is a valid, reachable code-path according to the compiler, it's just that you should never be calling `deliver_result()` at runtime when the variant holds the type `std::tuple<>`.

PR #521, opened 2023-05-21
Replace deprecated std::aligned_union with array

vasama:
`std::aligned_union` and `std::aligned_union_t` are deprecated in C++23.


ispeters (approved):
Looks great.  Thanks!

PR #524, opened 2023-05-24
Revamp blocking_kind calculation

ispeters:
This diff promotes a *Sender's* "blocking kind" to a static property of the *Sender's* `sender_traits` named `blocking`, which replaces the `cblocking<Sender>()` function.

PR #525, opened 2023-05-24
Add is always scheduler affine

ispeters:
This diff introduces a new *Sender* trait called `is_always_scheduler_affine` that can be set by declaring a static `bool` member of the *Sender* class named `is_always_scheduler_affine`; the default is false.  This diff also updates all existing *Senders* in Unifex to declare their *Scheduler* affinity if the defaults are wrong.

A future diff will leverage this property to make `task<>`'s *Scheduler* affinity implementation more efficient.


ispeters:
This depends on #524.

janondrusek (approved):

PR #526, opened 2023-05-24
Format and resort task hpp

ispeters:
This diff reformats `unifex/task.hpp` and resorts its contents.

### DIFF
--- a/include/unifex/task.hpp
+++ b/include/unifex/task.hpp
@@ -36,7 +36,7 @@
 #include <unifex/unstoppable.hpp>
 
 #if UNIFEX_NO_COROUTINES
-# error "Coroutine support is required to use this header"
+#  error "Coroutine support is required to use this header"
 #endif
 
 #include <exception>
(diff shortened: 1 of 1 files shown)


ispeters:
This depends on #525.

janondrusek (approved):
Should we sprinkle `const` and `final` while at it?

ispeters:
> Should we sprinkle `const` and `final` while at it?

I've done some of that in the next PR; I think it'd be better to address missed `const` and `final` there.

include/unifex/task.hpp
```diff
@@ -183,71 +215,81 @@ struct _promise {
...
+                unstoppable(schedule(this->sched_))));
       }
     }
 
     template <typename Sender>
     decltype(auto) transform_sender_(Sender&& sndr) {
-      if constexpr (blocking_kind::always_inline == sender_traits<remove_cvref_t<Sender>>::blocking) {
-        return unifex::await_transform(*this, (Sender&&) sndr);
+      if constexpr (sender_traits<
+                        remove_cvref_t<Sender>>::is_always_scheduler_affine) {
```
ispeters, line 266: It was a mistake to merge this in this PR; *Scheduler* affinity is a weaker constraint than `always_inline` so there could be situations where we wake up from awaiting a *Sender* on the wrong *Scheduler*.  This will be addressed in #527.

PR #531, opened 2023-05-25
add `io_uring_context::open_listening_socket`

janondrusek:
* opens a socket on a specified port
* exposes a _Stream_ of `async_read_write_file` to handle accept(s)

### DIFF
--- a/include/unifex/linux/io_uring_context.hpp
+++ b/include/unifex/linux/io_uring_context.hpp
@@ -21,8 +21,13 @@
 #include <unifex/detail/atomic_intrusive_queue.hpp>
 #include <unifex/detail/intrusive_heap.hpp>
 #include <unifex/detail/intrusive_queue.hpp>
+#include <unifex/defer.hpp>
 #include <unifex/file_concepts.hpp>
 #include <unifex/filesystem.hpp>
+#include <unifex/io_concepts.hpp>
+#include <unifex/just_done.hpp>
+#include <unifex/let_value_with.hpp>
+#include <unifex/socket_concepts.hpp>
 #include <unifex/get_stop_token.hpp>
 #include <unifex/manual_lifetime.hpp>
 #include <unifex/receiver_concepts.hpp>
@@ -43,7 +48,10 @@
 
 #include UNIFEX_LIBURING_HEADER
 
+#include <netinet/in.h>
+#include <sys/socket.h>
 #include <sys/uio.h>
+#include <unistd.h>
 
 #include <unifex/detail/prologue.hpp>
 
@@ -62,6 +70,8 @@ class io_uring_context {
   class async_read_write_file;
   class async_write_only_file;
   class scheduler;
+  class accept_sender;
+  class accept_stream;
 
   io_uring_context();
 
@@ -925,6 +935,10 @@ class io_uring_context::scheduler {
       tag_t<open_file_write_only>,
       scheduler s,
       const filesystem::path& path);
+  friend accept_stream tag_invoke(
+      tag_t<open_listening_socket>,
+      scheduler s,
+      port_t port);
 
   friend bool operator==(scheduler a, scheduler b) noexcept {
     return a.context_ == b.context_;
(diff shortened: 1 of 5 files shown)


include/unifex/linux/io_uring_context.hpp
```diff
@@ -942,6 +956,171 @@ inline io_uring_context::scheduler io_uring_context::get_scheduler() noexcept {
...
+  };
+
+public:
+  // Produces number of bytes read.
+  template <
+      template <typename...>
+      class Variant,
+      template <typename...>
+      class Tuple>
+  using value_types = Variant<Tuple<ssize_t>>;
```
ispeters, line 1044:
```suggestion
  // Produces an open read-write file corresponding to the accepted connection.
  template <
      template <typename...>
      class Variant,
      template <typename...>
      class Tuple>
  using value_types = Variant<Tuple< async_read_write_file >>;
```

include/unifex/linux/io_uring_context.hpp
```diff
@@ -942,6 +956,171 @@ inline io_uring_context::scheduler io_uring_context::get_scheduler() noexcept {
...
+
+  // Note: Only case it might complete with exception_ptr is if the
+  // receiver's set_value() exits with an exception.
+  template <template <typename...> class Variant>
+  using error_types = Variant<std::error_code, std::exception_ptr>;
+
+  static constexpr bool sends_done = true;
+
```
ispeters, line 1052: Can you add
```suggestion
  // dunno—is this maybe? never?
  static constexpr blocking_kind blocking = ??;
  
  // I think you always complete on the io_uring context
  // regardless of where you start, right?
  static constexpr bool is_always_scheduler_affine = false;
```

include/unifex/linux/io_uring_context.hpp
```diff
@@ -942,6 +956,171 @@ inline io_uring_context::scheduler io_uring_context::get_scheduler() noexcept {
...
+    friend io_uring_context;
+
+  public:
+    template <typename Receiver2>
+    explicit operation(const accept_sender& sender, Receiver2&& r) noexcept(
+        std::is_nothrow_constructible_v<Receiver, Receiver2>)
+      : context_(sender.context_)
+      , fd_(sender.fd_)
+      , receiver_((Receiver2 &&) r) {}
+
```
ispeters, line 973:
```suggestion

  operation(operation&&) = delete;
```

include/unifex/linux/io_uring_context.hpp
```diff
@@ -942,6 +956,171 @@ inline io_uring_context::scheduler io_uring_context::get_scheduler() noexcept {
   return scheduler{*this};
 }
 
+class io_uring_context::accept_sender {
+  using offset_t = std::int64_t;
+
+  template <typename Receiver>
+  class operation : private completion_base {
```
ispeters, line 963: A little surprising that this is a class template that's used directly, rather than using a nested, non-template `type`.

include/unifex/linux/io_uring_context.hpp
```diff
@@ -942,6 +956,171 @@ inline io_uring_context::scheduler io_uring_context::get_scheduler() noexcept {
...
+  static constexpr bool sends_done = true;
+
+  explicit accept_sender(io_uring_context& context, int fd) noexcept
+    : context_(context)
+    , fd_(fd) {}
+
+  template <typename Receiver>
+  operation<remove_cvref_t<Receiver>> connect(Receiver&& r) && {
+    return operation<remove_cvref_t<Receiver>>{*this, (Receiver &&) r};
+  }
```
ispeters, line 1060: Is this really a move-only _Sender_?  I'd expect you could support connecting lvalues, especially considering that `operation`'s constructor takes the _Sender_ by `const&`.

include/unifex/linux/io_uring_context.hpp
```diff
@@ -942,6 +956,171 @@ inline io_uring_context::scheduler io_uring_context::get_scheduler() noexcept {
...
+private:
+  io_uring_context& context_;
+  int fd_;
+};
+
+class io_uring_context::accept_stream {
+public:
+  using offset_t = std::int64_t;
+
+  explicit accept_stream(io_uring_context& context, std::int32_t port) noexcept
```
ispeters, line 1071: Wikipedia says port numbers are 16-bits; should this be a `uint16_t`?

include/unifex/linux/io_uring_context.hpp
```diff
@@ -942,6 +956,171 @@ inline io_uring_context::scheduler io_uring_context::get_scheduler() noexcept {
...
+
+  auto next() noexcept {
+    return let_value_with(
+        [this]() noexcept {
+          if (!fd_.valid()) {
+            open_socket();
+          }
+          return fd_.get();
+        },
+        [this](auto fd) noexcept { return accept_sender{context_, fd}; });
```
ispeters, line 1083: Would this be cleaner if you did the "if not valid, open a socket" dance inside the `accept_sender` constructor?

include/unifex/linux/io_uring_context.hpp
ispeters: I think maybe so; `sin6_port` is an `in_port_t` according to https://man7.org/linux/man-pages/man7/ipv6.7.html, and `in_port_t` is a `uint16_t` according to https://pubs.opengroup.org/onlinepubs/009695399/basedefs/netinet/in.h.html.

include/unifex/socket_concepts.hpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+#pragma once
+
+#include <unifex/tag_invoke.hpp>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+namespace _socket {
+using port_t = std::uint16_t;
```
ispeters, line 24: Maybe instead of a raw `uint16_t` above, you should be using this type.

ispeters (changes requested):
Looks good, but I've left a whole bunch of comments.

ispeters (changes requested):
Looks good, but I've left a whole bunch of comments.

include/unifex/linux/io_uring_context.hpp
```diff
@@ -942,6 +956,171 @@ inline io_uring_context::scheduler io_uring_context::get_scheduler() noexcept {
...
+      , receiver_((Receiver2 &&) r) {}
+
+    void start() noexcept {
+      if (!context_.is_running_on_io_thread()) {
+        this->execute_ = &operation::on_schedule_complete;
+        context_.schedule_remote(this);
+      } else {
+        start_io();
+      }
+    }
```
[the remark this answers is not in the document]
janondrusek: #533

include/unifex/linux/io_uring_context.hpp
janondrusek: #533

include/unifex/linux/io_uring_context.hpp
janondrusek: added a TODO

include/unifex/linux/io_uring_context.hpp
janondrusek: #533

ispeters (approved):

PR #534, opened 2023-05-30
add cancellation support into `io_uring_context`

janondrusek:
Allows the `io_uring_context` _Senders_ to complete with a cancellation signal and removes corresponding submissions from the ring.

### DIFF
--- a/include/unifex/linux/io_uring_context.hpp
+++ b/include/unifex/linux/io_uring_context.hpp
@@ -441,7 +441,8 @@ class io_uring_context::read_sender {
 
     void start_io() noexcept {
       UNIFEX_ASSERT(context_.is_running_on_io_thread());
-
+      stopCallback_.construct(
+            get_stop_token(receiver_), cancel_callback{*this});
       auto populateSqe = [this](io_uring_sqe & sqe) noexcept {
         sqe.opcode = IORING_OP_READV;
         sqe.fd = fd_;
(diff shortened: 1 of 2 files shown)


include/unifex/linux/io_uring_context.hpp
```diff
@@ -460,8 +461,47 @@ class io_uring_context::read_sender {
...
+    static void on_schedule_stop_complete(operation_base* op) noexcept {
+      static_cast<operation*>(op)->request_stop_local();
+    }
+
+    void request_stop_local() noexcept {
+      UNIFEX_ASSERT(context_.is_running_on_io_thread());
+
+      stopCallback_.destruct();
+      execute_ = &operation::complete_with_done;
+      context_.schedule_local(this);
```
ispeters, line 485: Why do you need to schedule again?  Why not invoke `set_done` right here?

include/unifex/linux/io_uring_context.hpp
```diff
@@ -441,7 +441,8 @@ class io_uring_context::read_sender {
 
     void start_io() noexcept {
       UNIFEX_ASSERT(context_.is_running_on_io_thread());
-
+      stopCallback_.construct(
+            get_stop_token(receiver_), cancel_callback{*this});
```
ispeters, line 445: In the case that `start()` is invoked on a "remote" thread, you're deferring stop callback construction until after we've thunked to the IO thread.  Is this intentional?  Why not construct the stop callback in `start()`?

Also, what happens if the stop token has already received a stop request by the time you construct the stop callback?  For `inplace_stop_token`, the callback constructor will invoke the callback before returning—do you handle that scenario?

ispeters (changes requested):
It doesn't feel like it's safe to merge this without doing something about how the `io_uring_context` handles the CQE for a cancelled operation; I feel like the current iteration is going to lead to invoking `execute` on a dead operation state.

include/unifex/linux/io_uring_context.hpp
```diff
@@ -460,8 +461,47 @@ class io_uring_context::read_sender {
       }
     }
 
+    void request_stop() noexcept {
+      if (refCount_.fetch_add(1, std::memory_order_relaxed) == 0) {
```
ispeters, line 465: I wonder if it'd be cheaper to use a CAS here because the CPU might not need to write on failure.
```suggestion
      if (char expected = 1; refCount_.compare_exchange_strong(expected, 2, std::memory_order_relaxed)) {
```

test/io_uring_2_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+const char* fdPath = "/proc/self/fd/";
+
+struct IOUringTest : testing::Test {
+  void SetUp() override {
+    ASSERT_NE(pipe(pipes_), -1) << "unable to create pipe";
+  }
+
+  ~IOUringTest() {
+    close(pipes_[0]);
+    close(pipes_[1]);
```
ispeters, line 55: Is this safe to do if the assertion fails in `SetUp`?

ispeters (approved):
Nice!

test/io_uring_2_test.cpp
janondrusek: I am not 100% sure even after reading `man close`. Making this conditional.

PR #536, opened 2023-05-31
add `debug_async_scope`

janondrusek:
`v1::debug_async_scope` and `v2::debug_async_scope` wrap the
  corresponding `async_scope`s and can be used as a direct replacement
for debugging purposes. They keep track of nested operation states, allowing to identify the 'stuck' operations especially in case of a deadlock.

### DIFF
--- a/include/unifex/detail/debug_async_scope.hpp
+++ b/include/unifex/detail/debug_async_scope.hpp
@@ -1,0 +1,253 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <unifex/detail/intrusive_list.hpp>
+#if __has_include(<cxxabi.h>)
+#  define UNIFEX_NO_DEMANGLE 0
+#  include <cxxabi.h>
+#else
+// TODO
+// https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-undecoratesymbolname?redirectedfrom=MSDN
+#  define UNIFEX_NO_DEMANGLE 1
+#endif
+#include <mutex>
+#include <typeinfo>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex::detail {
+namespace _debug_async_scope {
+
+struct op_base {
+  explicit op_base(const std::type_info& concreteType) noexcept
+    : concreteType(concreteType) {
+    demangled = demangle(concreteType);
+  }
+
+  const std::type_info& concreteType;
+  const char* demangled;
+  bool shouldFree{false};
+
+  const char* demangle(const std::type_info& concreteType) noexcept {
+#if !UNIFEX_NO_DEMANGLE
+    int status = -1;
+    auto result = abi::__cxa_demangle(concreteType.name(), nullptr, 0, &status);
+    if (status == 0) {
+      shouldFree = true;
+      return result;
+    }
+    return concreteType.name();
+#else
+    return concreteType.name();
+#endif
+  }
+
+  op_base* next{nullptr};
+  op_base* prev{nullptr};
+
+  ~op_base() {
+    if (shouldFree) {
+      std::free((void*)demangled);
+    }
+  }
+};
+
+using debug_ops_t = intrusive_list<op_base, &op_base::next, &op_base::prev>;
+
+struct debug_op_list final {
+  std::mutex mutex_;
+  debug_ops_t ops_;
+
+  void register_debug_operation(op_base* op) noexcept {
+    std::lock_guard lock{mutex_};
+    ops_.push_back(op);
+  }
+
+  void deregister_debug_operation(op_base* op) noexcept {
+    std::lock_guard lock{mutex_};
+    ops_.remove(op);
+  }
+};
+
+template <typename Receiver>
+struct _operation final {
+  struct type;
+};
+
+template <typename Receiver>
+struct _operation<Receiver>::type : op_base {
+  template <typename Receiver2>
+  explicit type(
+      const std::type_info& t,
+      debug_op_list* ops,
+      Receiver2&& receiver) noexcept(std::
+                                         is_nothrow_constructible_v<
+                                             Receiver,
+                                             Receiver2>)
+    : op_base{t}
+    , ops_(ops)
+    , receiver_(static_cast<Receiver2&&>(receiver)) {}
+
+  type(type&&) = delete;
+
+  debug_op_list* ops_;
+  UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+
+  template <typename Func>
+  void complete(Func func) noexcept {
+    ops_->deregister_debug_operation(this);
+    func(std::move(receiver_));
+  }
+};
+
+template <typename Receiver>
+struct _receiver final {
+  struct type;
+};
+
+template <typename Receiver>
+struct _receiver<Receiver>::type final {
+  typename _operation<Receiver>::type* op_;
+
+  template <typename... Values>
+  void set_value(Values&&... values) noexcept {
+    op_->complete([&](Receiver&& receiver) noexcept {
+      UNIFEX_TRY {
+        unifex::set_value(
+            std::move(receiver), static_cast<Values&&>(values)...);
+      }
+      UNIFEX_CATCH(...) {
+        unifex::set_error(std::move(receiver), std::current_exception());
+      }
+    });
+  }
+
+  template <typename E>
+  void set_error(E&& e) noexcept {
+    op_->complete([&e](Receiver&& receiver) noexcept {
+      unifex::set_error(std::move(receiver), static_cast<E&&>(e));
+    });
+  }
+
+  void set_done() noexcept { op_->complete(unifex::set_done); }
+};
+
+template <typename Sender, typename Receiver>
+struct _operation_impl final {
+  struct type;
+};
+
+template <typename Sender, typename Receiver>
+using debug_op =
+    typename _operation_impl<Sender, remove_cvref_t<Receiver>>::type;
+
+template <typename Sender, typename Receiver>
+struct _operation_impl<Sender, Receiver>::type final
+  : _operation<Receiver>::type {
+  using base_op_t = typename _operation<Receiver>::type;
+  using receiver_t = typename _receiver<Receiver>::type;
+
+  template <typename Sender2, typename Receiver2>
+  explicit type(
+      debug_op_list* ops,
+      Sender2&& sender,
+      Receiver2&& receiver) noexcept(std::
+                                         is_nothrow_constructible_v<
+                                             base_op_t,
+                                             const std::type_info&,
+                                             debug_op_list*,
+                                             Receiver2>&&
+                                             is_nothrow_connectable_v<
+                                                 Sender2,
+                                                 receiver_t>)
+    : base_op_t(
+          typeid(connect_result_t<Sender, Receiver>),
+          ops,
+          static_cast<Receiver2&&>(receiver))
+    , op_(unifex::connect(static_cast<Sender2&&>(sender), receiver_t{this})) {}
+
+  type(type&&) = delete;
+
+  friend void tag_invoke(tag_t<start>, type& op) noexcept {
+    op.ops_->register_debug_operation(&op);
+    unifex::start(op.op_);
+  }
+
+private:
+  using op_t = connect_result_t<Sender, receiver_t>;
+  op_t op_;
+};
+
+template <typename Sender>
+struct _debug_scope_sender final {
+  struct type;
+};
+
+template <typename Sender>
+struct _debug_scope_sender<Sender>::type final {
+  template <
+      template <typename...>
+      typename Variant,
+      template <typename...>
+      typename Tuple>
+  using value_types = sender_value_types_t<Sender, Variant, Tuple>;
+
+  template <template <typename...> typename Variant>
+  using error_types = typename concat_type_lists_unique_t<
+      sender_error_types_t<Sender, type_list>,
+      type_list<std::exception_ptr>>::template apply<Variant>;
+
+  static constexpr bool sends_done = sender_traits<Sender>::sends_done;
+
+  template <typename Sender2>
+  explicit type(Sender2&& sender, debug_op_list* ops) noexcept(
+      std::is_nothrow_constructible_v<Sender, Sender2>)
+    : ops_(ops)
+    , sender_(static_cast<Sender2&&>(sender)) {}
+
+  template(typename Self, typename Receiver)          //
+      (requires same_as<type, remove_cvref_t<Self>>)  //
+      friend debug_op<Sender, Receiver> tag_invoke(
+          tag_t<connect>,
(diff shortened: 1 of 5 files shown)


include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,209 @@
...
+    , demangled(demangle(concreteType)) {}
+
+  const std::type_info& concreteType;
+  const char* demangled;
+
+  static const char* demangle(const std::type_info& concreteType) noexcept {
+    int status = -1;
+    auto result = abi::__cxa_demangle(concreteType.name(), nullptr, 0, &status);
+    UNIFEX_ASSERT(status == 0);
+    return result;
```
ispeters, line 41: I feel like this might be better if it checked `status` and did something like `return "";` or `return "<failed to demangle>";` or even `return concreteType.name();` on failure.

Also, where does `result` point when this succeeds?  Do you need to deallocate something?

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,209 @@
...
+#include <mutex>
+#include <typeinfo>
+
+#include <unifex/detail/prologue.hpp>
+
+namespace unifex::detail {
+namespace _debug_async_scope {
+
+struct op_base {
+  op_base(const std::type_info& concreteType) noexcept
```
ispeters, line 30: This should probably be explicit.
```suggestion
  explicit op_base(const std::type_info& concreteType) noexcept
```

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,209 @@
...
+  struct type;
+};
+
+template <typename Receiver>
+struct _operation<Receiver>::type : op_base {
+  template <typename Receiver2>
+  type(
+      const std::type_info& t,
+      debug_op_list* ops,
+      Receiver2&& receiver) noexcept  // TODO
```
ispeters, line 76: TODO

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,209 @@
...
+  : _operation<Receiver>::type {
+  using base_op_t = typename _operation<Receiver>::type;
+  using receiver_t = typename _receiver<Receiver>::type;
+
+  template <typename Sender2, typename Receiver2>
+  type(
+      debug_op_list* ops,
+      Sender2&& sender,
+      Receiver2&& receiver) noexcept  // TODO
+    : base_op_t(typeid(type), ops, static_cast<Receiver2&&>(receiver))
```
ispeters, line 144: `typeid(type)` is always going to include this wrapper operation state's type, which is noise, right?
```suggestion
    : base_op_t(typeid(op_t), ops, static_cast<Receiver2&&>(receiver))
```

ispeters: Or maybe even this, which would give you the name of the type that _would_ be the operation state if not for the debug wrapper.
```suggestion
    : base_op_t(typeid(connect_result_t<Sender, Receiver>), ops, static_cast<Receiver2&&>(receiver))
```

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,209 @@
...
+  template <
+      template <typename...>
+      typename Variant,
+      template <typename...>
+      typename Tuple>
+  using value_types = sender_value_types_t<Sender, Variant, Tuple>;
+
+  template <template <typename...> typename Variant>
+  using error_types =
+      sender_error_types_t<Sender, Variant>;  // TODO exception_ptr
```
ispeters, line 175: TODO

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,209 @@
...
+  template <typename Sender2>
+  type(Sender2&& sender, debug_op_list* ops) noexcept(
+      std::is_nothrow_constructible_v<Sender, Sender2>)
+    : ops_(ops)
+    , sender_(static_cast<Sender2&&>(sender)) {}
+
+  template(typename Self, typename Receiver)          //
+      (requires same_as<type, remove_cvref_t<Self>>)  //
+      friend auto tag_invoke(
+          tag_t<connect>, Self&& self, Receiver&& receiver) noexcept {  // TODO
```
ispeters, line 188: TODO

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,209 @@
...
+};
+
+}  // namespace _debug_async_scope
+
+using _debug_async_scope::debug_op_list;
+template <typename Sender>
+using debug_scope_sender =
+    typename _debug_async_scope::_debug_scope_sender<Sender>::type;
+
+}  // namespace unifex::detail
```
ispeters, line 207: We should probably merge as-is, but this _Sender_ wrapper is kinda-sorta general purpose so it might make sense to elevate it out of the `detail` namespace.

include/unifex/v1/debug_async_scope.hpp
```diff
@@ -0,0 +1,144 @@
...
+    detached_spawn(
+        on(static_cast<Scheduler&&>(scheduler), static_cast<Sender&&>(sender)));
+  }
+
+  template(typename Scheduler, typename Fun)             //
+      (requires scheduler<Scheduler> AND callable<Fun>)  //
+      void detached_spawn_call_on(Scheduler&& scheduler, Fun&& fun) {
+    detached_spawn_on(
+        static_cast<Scheduler&&>(scheduler),
+        just_from(static_cast<Fun&&>(fun)));
```
ispeters, line 71: This should probably `static_assert` that `fun` is no-throw callable; if `fun` throws, it'll invoke `terminate()` on completion.

include/unifex/v2/debug_async_scope.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+  bool joined() const noexcept { return scope_.joined(); }
+
+  bool join_started() const noexcept { return scope_.join_started(); }
+
+  std::size_t use_count() const noexcept { return scope_.use_count(); }
+
+  template(typename Sender)                                //
+      (requires sender<Sender>)                            //
+      [[nodiscard]] auto nest(Sender&& sender) noexcept {  // TODO
```
ispeters, line 38: TODO

test/debug_async_scope_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <cstdio>
+#include <thread>
+
+#include <gtest/gtest.h>
+
+using namespace unifex;
+using namespace std::chrono;
+using namespace std::literals;
+
+// guaranteed to deadlock
```
ispeters, line 37: Perhaps leave a comment explaining why the test exists at all?  Something like "useful for triggering a deadlock that can be inspected in the debugger", maybe.

ispeters (changes requested):
Basically LGTM but you left several TODOs behind.

ispeters (changes requested):
Basically LGTM but you left several TODOs behind.

include/unifex/detail/debug_async_scope.hpp
janondrusek: From https://gcc.gnu.org/onlinedocs/libstdc++/manual/ext_demangling.html

> The demangler interface is described in the source documentation linked to above. It is actually written in C, so you don't need to be writing C++ in order to demangle C++. (That also means we have to use crummy memory management facilities, so don't forget to free() the returned char array.)

Looks like we need to call `free()`.

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,247 @@
...
+    if (status == 0) {
+      shouldFree = true;
+    } else {
+      return concreteType.name();
+    }
+#else
+    auto result = concreteType.name();
+#endif
+    return result;
+  }
```
ispeters, line 57: The CI all passes so I assume this works, but I wonder what the spec says the order of operations is, here.

Naively, I'm a little worried that the `shouldFree{false}` might be evaluated _after_ the `demangled(demangle(...))` expression because fields are initialized in declared order.  _If_ that were to happen, the assignment of `shouldFree = true` in `demangle()` would be overwritten by the initialization.

Now, it's quite possible that `shouldFree{false}` is executed _before_ the `op_base` constructor.

The way to be absolutely certain is to leave `demangled` uninitialized and assign it the result of `demangle()` in the constructor body.

Also, I kinda hate the non-lexical scoping of your `#if/else` block.
```suggestion
  explicit op_base(const std::type_info& concreteType) noexcept
    : concreteType(concreteType) {
    demangled = demangle(concreteType);
  }

  const std::type_info& concreteType;
  const char* demangled;
  bool shouldFree{false};

  const char* demangle(const std::type_info& concreteType) noexcept {
#if !UNIFEX_NO_DEMANGLE
    int status = -1;
    auto result = abi::__cxa_demangle(concreteType.name(), nullptr, 0, &status);
    if (status == 0) {
      shouldFree = true;
      return result;
    }

    return concreteType.name();
#else
    return concreteType.name();
#endif
  }
```

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,247 @@
...
+
+  void register_debug_operation(op_base* op) noexcept {
+    std::lock_guard lock{mutex_};
+    ops_.push_back(op);
+  }
+
+  void deregister_debug_operation(op_base* op) noexcept {
+    std::lock_guard lock{mutex_};
+    ops_.remove(op);
+  }
```
ispeters, line 83: Technically, `mutex_.lock()` can throw; terminating if it does in this case is probably fine….

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,247 @@
...
+
+template <typename Receiver>
+struct _operation final {
+  struct type;
+};
+
+template <typename Receiver>
+struct _operation<Receiver>::type : op_base {
+  template <typename Receiver2>
+  type(
```
ispeters, line 94: Nit: most constructors in Unifex are `explicit`.
```suggestion
  explicit type(
```

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,247 @@
...
+    typename _operation_impl<Sender, remove_cvref_t<Receiver>>::type;
+
+template <typename Sender, typename Receiver>
+struct _operation_impl<Sender, Receiver>::type final
+  : _operation<Receiver>::type {
+  using base_op_t = typename _operation<Receiver>::type;
+  using receiver_t = typename _receiver<Receiver>::type;
+
+  template <typename Sender2, typename Receiver2>
+  type(debug_op_list* ops, Sender2&& sender, Receiver2&& receiver) noexcept(
```
ispeters, line 165: Nit
```suggestion
  explicit type(debug_op_list* ops, Sender2&& sender, Receiver2&& receiver) noexcept(
```

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,247 @@
...
+
+  template <template <typename...> typename Variant>
+  using error_types = typename concat_type_lists_unique_t<
+      sender_error_types_t<Sender, type_list>,
+      type_list<std::exception_ptr>>::template apply<Variant>;
+
+  static constexpr bool sends_done = sender_traits<Sender>::sends_done;
+
+  template <typename Sender2>
+  type(Sender2&& sender, debug_op_list* ops) noexcept(
```
ispeters, line 211: Nit
```suggestion
  explicit type(Sender2&& sender, debug_op_list* ops) noexcept(
```

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,247 @@
...
+      (requires same_as<type, remove_cvref_t<Self>>)  //
+      friend auto tag_invoke(
+          tag_t<connect>,
+          Self&& self,
+          Receiver&& receiver) noexcept(std::
+                                            is_nothrow_constructible_v<
+                                                debug_op<Sender, Receiver>,
+                                                debug_op_list*,
+                                                Sender,
+                                                Receiver&>) {
```
ispeters, line 226: I think you've got the value categories wrong here.

`member_t<A, B>` effectively returns `B` but with the same `const` and value category as `A`, which is what you're getting below when you do `static_cast<Self&&>(self).sender_`.

`Receiver` will either be deduced to be a class type, or an lvalue reference to a class type; using `Receiver&` here forces the lvalue reference version (ref of class is ref of class and ref of ref of class is ref of class).

```suggestion
                                                member_t<Self, Sender>,
                                                Receiver>) {
```

include/unifex/detail/debug_async_scope.hpp
```diff
@@ -0,0 +1,247 @@
...
+
+  template <typename Sender2>
+  type(Sender2&& sender, debug_op_list* ops) noexcept(
+      std::is_nothrow_constructible_v<Sender, Sender2>)
+    : ops_(ops)
+    , sender_(static_cast<Sender2&&>(sender)) {}
+
+  template(typename Self, typename Receiver)          //
+      (requires same_as<type, remove_cvref_t<Self>>)  //
+      friend auto tag_invoke(
```
ispeters, line 218: Might help compile time to say `debug_op<Sender, Receiver>` here.
```suggestion
      friend debug_op<Sender, Receiver> tag_invoke(
```

include/unifex/v1/debug_async_scope.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+    detached_spawn_on(
+        static_cast<Scheduler&&>(scheduler),
+        just_from(static_cast<Fun&&>(fun)));
+  }
+
+  template <typename Sender>
+  [[nodiscard]] auto
+  attach(Sender&& sender) noexcept(std::is_nothrow_constructible_v<
+                                   unifex::detail::debug_scope_sender<Sender>,
```
ispeters, line 81:
```suggestion
                                   unifex::detail::debug_scope_sender<remove_cvref_t<Sender>>,
```

include/unifex/v2/debug_async_scope.hpp
```diff
@@ -0,0 +1,64 @@
...
+  bool joined() const noexcept { return scope_.joined(); }
+
+  bool join_started() const noexcept { return scope_.join_started(); }
+
+  std::size_t use_count() const noexcept { return scope_.use_count(); }
+
+  template <typename Sender>
+  static constexpr bool sender_nothrow_constructible =
+      std::is_nothrow_constructible_v<
+          unifex::detail::debug_scope_sender<Sender>,
```
ispeters, line 39:
```suggestion
          unifex::detail::debug_scope_sender<remove_cvref_t<Sender>>,
```

include/unifex/v2/debug_async_scope.hpp
```diff
@@ -0,0 +1,64 @@
...
+          Sender,
+          unifex::detail::debug_op_list*>;
+  template <typename Sender>
+  static constexpr bool nest_nothrow_invocable = noexcept(
+      UNIFEX_DECLVAL(unifex::v2::async_scope)
+          .nest(UNIFEX_DECLVAL(unifex::detail::debug_scope_sender<Sender>)));
```
ispeters, line 45:
```suggestion
          .nest(UNIFEX_DECLVAL(unifex::detail::debug_scope_sender<remove_cvref_t<Sender>>)));
```

include/unifex/v2/debug_async_scope.hpp
```diff
@@ -0,0 +1,64 @@
...
+  template(typename Sender)      //
+      (requires sender<Sender>)  //
+      [[nodiscard]] auto nest(Sender&& sender) noexcept(
+          sender_nothrow_constructible<Sender>&&
+              nest_nothrow_invocable<Sender>) {
+    return scope_.nest(unifex::detail::debug_scope_sender<Sender>{
```
ispeters, line 51:
```suggestion
    return scope_.nest(unifex::detail::debug_scope_sender<remove_cvref_t<Sender>>{
```

ispeters (approved):
Some nits and some value category errors, but otherwise good.

ispeters (approved):
Some nits and some value category errors, but otherwise good.

PR #540, opened 2023-06-14
Create a nothrow task with tests

jesswong:
Modifies task.hpp to include nothrow_task. The behavior for nothrow_task and task.hpp is mostly the same, with the difference being nothrow_task calling std::terminate when an exception is thrown. Part of the motivation is to create a task that's smaller in binary size compared to task by minimizing the use of exceptions.

The other difference is that the scheduler affinity in nothrow_task and task is slightly different, as illustrated by the tests.

### DIFF
--- a/include/unifex/connect_awaitable.hpp
+++ b/include/unifex/connect_awaitable.hpp
@@ -76,12 +76,13 @@ public:
 
     template <typename Func>
     auto yield_value(Func&& func) noexcept {
+      static constexpr bool isNothrowInvocable = std::is_nothrow_invocable_v<Func>;
       struct awaiter {
         Func&& func_;
         bool await_ready() noexcept {
           return false;
         }
-        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(std::is_nothrow_invocable_v<Func>) {
+        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(isNothrowInvocable) {
           ((Func &&) func_)();
         }
         [[noreturn]] void await_resume() noexcept {
--- a/include/unifex/connect_awaitable.hpp
+++ b/include/unifex/connect_awaitable.hpp
@@ -81,7 +81,7 @@ public:
         bool await_ready() noexcept {
           return false;
         }
-        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(noexcept(((Func &&) func_)())) {
+        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(std::is_nothrow_invocable_v<Func>) {
           ((Func &&) func_)();
         }
         [[noreturn]] void await_resume() noexcept {
--- a/include/unifex/connect_awaitable.hpp
+++ b/include/unifex/connect_awaitable.hpp
@@ -75,21 +75,22 @@ public:
     }
 
     template <typename Func>
+    struct awaiter {
+      Func&& func_;
+      bool await_ready() noexcept {
+        return false;
+      }
+      void await_suspend(coro::coroutine_handle<promise_type>) noexcept(std::is_nothrow_invocable_v<Func>) {
+        ((Func &&) func_)();
+      }
+      [[noreturn]] void await_resume() noexcept {
+        std::terminate();
+      }
+    };
+
+    template <typename Func>
     auto yield_value(Func&& func) noexcept {
-      static constexpr bool isNothrowInvocable = std::is_nothrow_invocable_v<Func>;
-      struct awaiter {
-        Func&& func_;
-        bool await_ready() noexcept {
-          return false;
-        }
-        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(isNothrowInvocable) {
-          ((Func &&) func_)();
-        }
-        [[noreturn]] void await_resume() noexcept {
-          std::terminate();
-        }
-      };
-      return awaiter{(Func &&) func};
+      return awaiter<Func&&>{static_cast<Func&&>(func)};
     }
 
     template <typename Value>
--- a/include/unifex/connect_awaitable.hpp
+++ b/include/unifex/connect_awaitable.hpp
@@ -81,7 +81,7 @@ public:
         bool await_ready() noexcept {
           return false;
         }
-        void await_suspend(coro::coroutine_handle<promise_type>) {
+        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(noexcept(((Func &&) func_)())) {
           ((Func &&) func_)();
         }
         [[noreturn]] void await_resume() noexcept {
@@ -193,7 +193,7 @@ namespace _await_cpo {
       } catch (...) {
         ex = std::current_exception();
       }
-      co_yield[&] {
+      co_yield[&]() noexcept {
         unifex::set_error(std::move(receiver), std::move(ex));
       };
 #endif // !UNIFEX_NO_EXCEPTIONS
--- a/include/unifex/task.hpp
+++ b/include/unifex/task.hpp
@@ -227,21 +227,21 @@ struct _result_and_unhandled_exception final {
       if constexpr (nothrow) {
         std::terminate();
       } else {
-        expected_.reset_value();
+        result_.reset_value();
         unifex::activate_union_member(
-            expected_.exception_, std::current_exception());
-        expected_.state_ = _state::exception;
+            result_.exception_, std::current_exception());
+        result_.state_ = _state::exception;
       }
     }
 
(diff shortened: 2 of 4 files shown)


include/unifex/connect_awaitable.hpp
```diff
@@ -81,7 +81,7 @@ class _sender_task<Receiver>::type {
         bool await_ready() noexcept {
           return false;
         }
-        void await_suspend(coro::coroutine_handle<promise_type>) {
+        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(noexcept(((Func &&) func_)())) {
```
ispeters, line 84: I wonder if you could fix the Windows-only build error with this change.
```suggestion
        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(std::is_nothrow_invocable_v<Func>) {
```

include/unifex/task.hpp
```diff
@@ -222,54 +222,71 @@ struct _result_and_unhandled_exception final {
...
+        this->expected_.reset_value();
+        unifex::activate_union_member(
+            this->expected_.value_, static_cast<Args&&>(values)...);
+        this->expected_.state_ = _state::value;
       }
-      return std::move(expected_.value_).get();
     }
 
-    _expected<T> expected_;
+    std::conditional_t<nothrow, manual_lifetime<T>, _expected<T>> expected_;
```
ispeters, line 261: I think I'd rename this to `result_` now that it's not necessarily an `_expected<T>`.
```suggestion
    std::conditional_t<nothrow, manual_lifetime<T>, _expected<T>> result_;
```

include/unifex/task.hpp
```diff
@@ -314,8 +331,9 @@ struct _promise final {
     }
 
     template <typename Value>
+    // todo: consider if this should be nothrow or not
     decltype(auto) await_transform(Value&& value) {
-      if constexpr (is_sender_for_v<remove_cvref_t<Value>, schedule>) {
+      if constexpr (!nothrow && is_sender_for_v<remove_cvref_t<Value>, schedule>) {
```
ispeters, line 336: I think we should find a way to warn folks that no-throw tasks don't support a magic `co_await schedule(…)`.

Maybe, instead of adding `!nothrow` to the condition here, add a `static_assert(!nothrow, "Magic rescheduling isn't supported by no-throw tasks")`, or something, inside the `if constexpr` block.

What do you think?

include/unifex/task.hpp
```diff
@@ -455,7 +473,7 @@ struct _sr_thunk_promise final {
    */
   struct type final
     : _sr_thunk_promise_base
-    , _return_value_or_void<T>::type {
+    , _return_value_or_void<T, false>::type {
```
ispeters, line 476: Was this a TODO item?  I can't remember.  We can land like this, but we should probably follow up later with some analysis to determine whether `_sr_thunk_task` should have a no-throw version.

include/unifex/task.hpp
```diff
@@ -632,18 +650,18 @@ struct _sr_thunk_task<T>::type final : coro_holder {
  * Await the given sa_task<> in a context that will deliver stop requests from
  * the receiver on the expected scheduler.
  */
-template <typename T>
+template <typename T, bool nothrow>
 typename _sr_thunk_task<T>::type
```
ispeters, line 654: Maybe this should just return a `_sr_thunk_task<T, nothrow>::type` right now instead of being a TODO.

test/noexcept_task_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include <unifex/static_thread_pool.hpp>
+#include <unifex/when_all.hpp>
+#include <unifex/sync_wait.hpp>
+#include <unifex/task.hpp>
+#include <unifex/single_thread_context.hpp>
+#include <unifex/just_error.hpp>
+#include <unifex/stop_if_requested.hpp>
+
+
+#include <gtest/gtest.h>
```
ispeters, line 32: Could you `clang-format` this file?

test/noexcept_task_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+nothrow_task<int> bar() {
+  try {
+    co_await foo();
+    ADD_FAILURE();
+  }
+  catch (...) {
+    ADD_FAILURE();
+  }
+  co_return -1;
```
ispeters, line 77: This made me realize we should have a test case that proves that you can safely catch exceptions _within_ a `nothrow_task`; i.e. if `co_await foo()` throws an exception, you ought to be able to catch it here and handle it without the `nothrow_task` machinery somehow interrupting things and terminating the process.

test/noexcept_task_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+  co_await when_all(child(s, x), child(s, x));
+}
+
+nothrow_task<void> nothrowThrowsException() {
+  throw std::exception{};
+  co_return;
+}
+
+nothrow_task<void> nothrowJustError() {
+  co_await unifex::just_error(std::exception_ptr());
```
ispeters, line 56: I'm pretty sure it's Undefined Behaviour to try to throw a default-constructed `exception_ptr`.  This should be something like `std::make_exception_ptr(42)`.

ispeters (commented):
Looks pretty good!  I wonder if my suggestion in `connect_awaitable.hpp` addresses the MSVC build error.

include/unifex/task.hpp
jesswong: Won't this trigger an std::terminate if a nothrow task calls schedule()? I think schedule() in a nothrow task should still be valid behavior, even if it ends up behaving differently than schedule() in a regular task. 

i can add a comment block here calling this out for nothrow task, and add it explicitly to our documentation.

include/unifex/task.hpp
jesswong: I think we just missed this. I'll add a todo here. Most of the methods here are already noexcept, but this is dependent on _sr_thunk_promise_base which has an await_transform that we also were not sure if it should be noexcept yet.

include/unifex/task.hpp
ispeters: > Won't this trigger an std::terminate if a nothrow task calls schedule()?

No, it'd fail to compile.

> I think schedule() in a nothrow task should still be valid behavior, even if it ends up behaving differently than schedule() in a regular task. What do you think?

I'm torn. Long term, I agree it should be valid, but I'm worried about people switching from `task<T>` to `nothrow_task<T>` (especially us, internally) and accidentally changing the scheduling behaviour of their code.

If you want to change this coroutine from `task<int>` to `nothrow_task<int>` and you want to _keep_ the reschedule for some reason, you can just pass it through a `then` to hide its type:
```
task<int> expensiveInt() {
  co_await schedule(s);

  co_return 42;
}
```
can be changed to
```
nothrow_task<int> expensiveInt() {
  // still reschedules but doesn't change the current scheduler
  co_await then(schedule(s), []() noexcept {});

  co_return 42;
}
```

In the farther future, I expect to unwind the `sender_for` machinery that Eric added to be able to detect `co_await schedule(s);`. To do that, we'll have to invent a new way to change the current coroutine's current scheduler (something like `co_await reschedule_current_coroutine(s);`), which we can ensure supports both `task<>` and `nothrow_task<>`. Once we've done that, we can delete the magic in `task<>` and remove the `static_assert` in `nothrow_task<>`.

include/unifex/connect_awaitable.hpp
```diff
@@ -81,7 +81,7 @@ class _sender_task<Receiver>::type {
         bool await_ready() noexcept {
           return false;
         }
-        void await_suspend(coro::coroutine_handle<promise_type>) {
+        void await_suspend(coro::coroutine_handle<promise_type>) noexcept(std::is_nothrow_invocable_v<Func>) {
```
ispeters, line 84: MSVC is complaining that `Func` is "not a symbol", or something, which is bananas because it's clearly a symbol that's in scope.

There must be some bug with defining a function-local struct with traits that depend on the surrounding function's template parameters.  Some possible workarounds:
 - just before `struct awaiter {`, declare a `static constexpr bool isNothrowInvocable = std::is_nothrow_invocable_v<Func>;` and change this line to be `noexcept(isNothrowInvocable)`
 - move the definition of `awaiter` out of `yield_value`, which requires making it a class template like this:

```
template <typename Func>
struct awaiter final {
  // we want a forwarding reference here
  Func&& func_;

  // the rest as before
};

template <typename Func>
auto yield_value(Func&& func) noexcept {
  return awaiter<Func&&>{static_cast<Func&&>(func)};
}
```

ispeters (approved):

PR #541, opened 2023-06-15
Fix regression in task<>

ccotter:
Prior to #290, task<> could co_await awaitables that returned references in await_resume.

Test plan: added test case that passes at the commit before #290, then fails after #290, and passes again with the changes in this commit. I also added a test for task<> awaiting a sender that sends a reference, but this was not actually broken.

### DIFF
--- a/test/task_test.cpp
+++ b/test/task_test.cpp
@@ -29,12 +29,18 @@ using namespace unifex;
 
 int global;
 
+template <class Expected, class Actual>
+void check_is_type(Actual&&) {
+    static_assert(std::is_same_v<Expected, Actual>);
+}
+
 task<int&> await_reference_awaitable() {
   struct AwaitableGlobalRef {
     bool await_ready() noexcept { return true; }
     void await_suspend(coro::coroutine_handle<>) noexcept {};
     int& await_resume() noexcept { return global; }
   };
+  check_is_type<int&>(co_await AwaitableGlobalRef{});
   int& x = co_await AwaitableGlobalRef{};
   co_return x;
 }
--- a/test/task_test.cpp
+++ b/test/task_test.cpp
@@ -51,14 +51,23 @@ task<int&> await_reference_sender() {
 }
 
 TEST(Task, AwaitAwaitableReturningReference) {
-  global = 0;
+  static_assert(std::is_same_v<
+    sender_value_types_t<decltype(await_reference_awaitable()), std::variant, std::tuple>,
+    std::variant<std::tuple<int&>>
+  >);
+
   int& ref = sync_wait(await_reference_awaitable()).value();
+  static_assert(std::is_same_v<
+    std::optional<std::reference_wrapper<int>>,
+    decltype(sync_wait(await_reference_awaitable()))>);
   EXPECT_EQ(&ref, &global);
 }
 
 TEST(Task, AwaitSenderReturningReference) {
-  global = 0;
   int& ref = sync_wait(await_reference_sender()).value();
+  static_assert(std::is_same_v<
+    std::optional<std::reference_wrapper<int>>,
+    decltype(sync_wait(await_reference_sender()))>);
   EXPECT_EQ(&ref, &global);
 }
 
--- a/test/task_test.cpp
+++ b/test/task_test.cpp
@@ -18,12 +18,10 @@
 
 #if !UNIFEX_NO_COROUTINES
 
-#include <atomic>
 
-#include <unifex/just.hpp>
+#include <unifex/just_from.hpp>
 #include <unifex/sync_wait.hpp>
 #include <unifex/task.hpp>
-#include <unifex/then.hpp>
 
 #include <gtest/gtest.h>
 
@@ -42,7 +40,7 @@ task<int&> await_reference_awaitable() {
 }
 
 task<int&> await_reference_sender() {
-  int& x = co_await then(just(), []() -> int& { return global; });
+  int& x = co_await just_from([]() -> int& { return global; });
   co_return x;
 }
 
--- a/test/task_test.cpp
+++ b/test/task_test.cpp
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <unifex/coroutine.hpp>
+
+#if !UNIFEX_NO_COROUTINES
+
+#include <atomic>
+
+#include <unifex/just.hpp>
+#include <unifex/sync_wait.hpp>
+#include <unifex/task.hpp>
+#include <unifex/then.hpp>
+
+#include <gtest/gtest.h>
+
+using namespace unifex;
+
+int global;
+
+task<int&> await_reference_awaitable() {
+  struct AwaitableGlobalRef {
+    bool await_ready() noexcept { return true; }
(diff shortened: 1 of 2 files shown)


test/task_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <unifex/coroutine.hpp>
+
+#if !UNIFEX_NO_COROUTINES
+
+#include <atomic>
+
+#include <unifex/just.hpp>
+#include <unifex/sync_wait.hpp>
+#include <unifex/task.hpp>
+#include <unifex/then.hpp>
```
janondrusek, line 26:
```suggestion

#include <unifex/just_from.hpp>
#include <unifex/sync_wait.hpp>
#include <unifex/task.hpp>
```

test/task_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+    bool await_ready() noexcept { return true; }
+    void await_suspend(coro::coroutine_handle<>) noexcept {};
+    int& await_resume() noexcept { return global; }
+  };
+  int& x = co_await AwaitableGlobalRef{};
+  co_return x;
+}
+
+task<int&> await_reference_sender() {
+  int& x = co_await then(just(), []() -> int& { return global; });
```
janondrusek, line 45: nit:
```suggestion
  int& x = co_await just_from([]() -> int& { return global; });
```

janondrusek (commented):
I'll let @ispeters stamp the change. Left a nit suggestion.

ispeters:
Looks good. We have more comprehensive CI internally because of all the Unifex users we have so I want to run that before I sign off to make sure we're not breaking anything, but I should be able to get to this today or tomorrow.

I assume, though, that the result of the internal CI will be green because I bet you're only adding functionality we're not yet using.

Thanks!

test/task_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+    static_assert(std::is_same_v<Expected, Actual>);
+}
+
+task<int&> await_reference_awaitable() {
+  struct AwaitableGlobalRef {
+    bool await_ready() noexcept { return true; }
+    void await_suspend(coro::coroutine_handle<>) noexcept {};
+    int& await_resume() noexcept { return global; }
+  };
+  check_is_type<int&>(co_await AwaitableGlobalRef{});
```
ccotter, line 43: Before #290, this line passed. With #290, (with and without my changes in this PR), this line fails due to Actual being `std::reference_wrapper<int>`

The problem is in `with_scheduler_affinity`'s use of `finally`, where `finally` does not preserve references (finally decays the values of the input sender).

I *think* that we could
 1. fix `with_scheduler_affinity` to preserve any references that the input sender sends.
 2. remove the `wrap_reference_t` from connect_awaitable.hpp

Alternatively, I'm not sure why `finally` doesn't (or can't) preserve references. If that's a design bug that's fixable, then we could also update `finally` to fix this problem as well.

ispeters: We generally have trouble with senders that produce references. I'm not sure whether that's intentional or an accident.

I wonder if @ericniebler, @lewissbaker, @kirkshoop, or @LeeHowes knows.

It seems to me that, like functions, senders ought to be able to produce anything, including references.

ericniebler: Probably just bugs. A sender should be able to send references.

ccotter: Ok given that, so I think the two options are

 1. Update `finally` to handle references without `decay_t`-ing them (backwards incompatible change). Then, remove the `wrap_reference_t` stuff from connect_awaitable.hpp
 2. Introduce a `v2::finally` that preserves references, then update connect_awaitable to use it and remove `wrap_reference_t`.
 3. A third option is to not change `finally` at all and to update `connect_awaitable` / `with_scheduler_afinity` to work around `finally`'s limitation, although I don't think this is a good long term solution.

@ispeters @janondrusek which do you think is appropriate? Do we need to preserve backwards compatibility of `finally`?

ispeters: I prefer the simplicity of 1, but I'd like to explore the costs.

First, I wonder if it's really backwards incompatible. I'm not convinced that Unifex generally supports sending references today so, to the extent that we have users at all, I wouldn't expect anyone to be _trying_ to send references (otherwise we'd have complaints that it doesn't work). Why do you think that adding support for references to `finally` would change or break existing users?  I'm probably missing something, but it seems like it's likely to be a straight addition of functionality.

Second, if it really is a breaking change, I feel ignorant about the relevant norms. Some things on my mind:
 - We don't have a lot of users so there aren't that many people to break.
 - We _have_ tagged some commits with version numbers—I'd be happy to invent a new version number for "just before we fixed `finally`".
 - Given that we've tagged some commits with version numbers, do people expect `main` to be stable from commit to commit?

What do you think, @ccotter?

ccotter: I mean a breaking change since technically, updating `finally` to handle references would be a breaking change since behavior of certain code could change (or code fail to compile), since finally today handles input senders that send references by `decay_t`-ing the values sent on the output sender ([contrived example](https://godbolt.org/z/73Yn59zqf)). I myself don't have any code that relies on `finally` where the input sender sends references, so I'm completely fine with option 1. My main point with raising the compatibility concern was to have you speak to any compatibility with code you have internally.

 > Given that we've tagged some commits with version numbers, do people expect main to be stable from commit to commit?

Thanks by the way for recently starting to tag versions, as that is helpful to me. I think with the tagging in place, speaking for myself, I'd more happy for `main` to be less "stable." Given that the version numbers are 0.X anyway, I think we're OK making a relatively minor behavior change to `finally` and a new minor version.

For anyone that does rely on the `decay_t`-ing, it'd be straightforward to fix such code with a `then()` that does the decaying. If we cared about a smooth migration plan, and it doesn't sound like we do, we could even proactively detect existing code with a clang-query or other technique...

My assumption here is that your internal users are the most widespread users of unifex, so if nothing you are aware of relies on `finally` ` decay_t`-ing references today, then I think option 1 makes most sense and is also the simplest route.

ispeters: Ok, point taken regarding it being breaking.

I say we tag a new version (I should probably have done so before I started removing deprecated untyped sender things—we can do that retroactively) and just fix `finally`.

I don't know how we're going to find broken stuff internally without just breaking it and waiting for fallout, but we'll have to figure that out internally.

ccotter: https://github.com/facebookexperimental/libunifex/pull/545 fixes `finally` to preserve references in the value channel. If that's merged, I'll update this PR to fix the regression in `task`.

ispeters: Thanks!

test/task_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+  int& x = co_await just_from([]() -> int& { return global; });
+  co_return x;
+}
+
+TEST(Task, AwaitAwaitableReturningReference) {
+  global = 0;
+  int& ref = sync_wait(await_reference_awaitable()).value();
+  EXPECT_EQ(ref, 0);
+  global = 10;
+  EXPECT_EQ(ref, 10);
```
ispeters, line 58: Similar comment to one of your other PRs; I think we should check address equality here rather than validating that writing to `global` updates what's read from `ref`.
```suggestion
  EXPECT_EQ(&ref, &global);
```

test/task_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+TEST(Task, AwaitSenderReturningReference) {
+  global = 0;
+  int& ref = sync_wait(await_reference_sender()).value();
+  EXPECT_EQ(ref, 0);
+  global = 10;
+  EXPECT_EQ(ref, 10);
```
ispeters, line 66: Same.
```suggestion
  EXPECT_EQ(&ref, &global);
```

ispeters (approved):
Please update the tests.  I'll look into validating this internally before merging it.

ccotter:
```
..\test\task_test.cpp(56): note: see declaration of 'std::tuple<Result>'
        with
        [
            Result=int &
        ]
```

Not quite sure about this. I'm not familiar with MSVC. The best I could do was use godbolt to confirm that the latest MSVC accepts a basic [`is_same_v<T, variant<tuple<int&>>>`](https://godbolt.org/z/Kbf1ej797)

ispeters:
Dunno why we've been sitting on this; I checked, and merging this doesn't break any of our internal CI so I think it's good.

PR #542, opened 2023-06-25
Remove deprecated stuff

ispeters:
This builds on PRs #538 and #539 to remove the concept of "untyped" _Senders_ and thus stop calling any _Senders_ "typed" (they're all typed).
 - #539 made `typed_sender<S>` an alias for `sender<S>` so now I can replace all uses of `typed_sender<>` with `sender<>` and deprecate `typed_sender<>`.
 - In the same spirit as #539, there's no need to refer to bulk _Senders_ as "typed" so I've renamed `typed_bulk_sender<>` to `bulk_sender<>` and deprecated `typed_bulk_sender<>`.
 - There's no need to call it `typed_via()` so `via()` now means what `typed_via()` meant and we retain `typed_via()` as a deprecated alias for `via()`.
 - Similarly for `typed_via_stream()`, `via_stream()` now means what `typed_via_stream()` meant and we retain `typed_via_stream()` as a deprecated alias for `via_stream()`.
 - I'm also deleting three things from `sender_concepts.hpp` that have been deprecated for a long time.

### DIFF
--- a/include/unifex/sender_concepts.hpp
+++ b/include/unifex/sender_concepts.hpp
@@ -179,9 +179,9 @@ UNIFEX_CONCEPT //
     detail::_has_sender_types<sender_traits<remove_cvref_t<S>>>;
 
 template <typename S>
-UNIFEX_CONCEPT        //
-  typed_bulk_sender = //
-    sender<S> &&      //
+UNIFEX_CONCEPT    //
+  bulk_sender =   //
+    sender<S> &&  //
     detail::_has_bulk_sender_types<sender_traits<remove_cvref_t<S>>>;
 
 namespace _start_cpo {
--- a/include/unifex/sender_concepts.hpp
+++ b/include/unifex/sender_concepts.hpp
@@ -104,7 +104,7 @@ namespace detail {
       UNIFEX_FRAGMENT(detail::_has_bulk_sender_types_impl, S);
 
   template <typename S>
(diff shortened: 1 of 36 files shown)


include/unifex/sender_concepts.hpp
```diff
@@ -179,14 +179,9 @@ UNIFEX_CONCEPT //
     detail::_has_sender_types<sender_traits<remove_cvref_t<S>>>;
 
 template <typename S>
-UNIFEX_CONCEPT   //
-  typed_sender = //
```
ispeters, line 183: Assuming it's even possible, it might be a good idea to mark this deprecated instead of deleting it.

include/unifex/sender_concepts.hpp
```diff
@@ -179,14 +179,9 @@ UNIFEX_CONCEPT //
...
-    sender<S>;
-
-template <typename S>
-UNIFEX_CONCEPT        //
-  typed_bulk_sender = //
```
ispeters, line 188: As above with `typed_sender`, maybe this should be retained as a deprecated alias for `bulk_sender`.

include/unifex/sender_concepts.hpp
ericniebler: Concepts don't get attributes until C++23. But you can make it a Boolean variable template.

jesswong (approved):

PR #543, opened 2023-06-25
Handle many error predecessors in upon_error

ccotter:
upon_error did not handle predecessor senders that send 0 or more than 2 possible error types. This commit brings upon_error into conformance with the P2300 proposal's upon_error.


ccotter:
I noticed this writing some tests for a reference capable `finally()`, and I assumed `upon_error` not supporting !=1 error predecessors was not intentional.

ispeters (approved):

ispeters:
Thanks!

PR #546, opened 2023-06-26
Fix concat_type_lists_unique_t to be "more unique"

ccotter:
The current behavior of concat_type_lists_unique_t is a little confusing.

```
concat_type_lists_unique_t<type_list<int, int>, type_list<int>> ->
    type_list<int, int>
concat_type_lists_unique_t<type_list<int>, type_list<int, int>> ->
    type_list<int>
```

In other words, concat_type_lists_unique_t is not commutative.

The docs seemed to indicate concat_type_lists_unique_t expected the lists to already contain unique elements (the comment seemed incomplete though). I am changing the behavior of concat_type_lists_unique_t to remove duplicate elements from the input lists,  since it conveniently fixes any sender algos that use it to construct the value_types of the output sender to not contain duplicate types.

### DIFF
--- a/include/unifex/type_list.hpp
+++ b/include/unifex/type_list.hpp
@@ -102,20 +102,20 @@ namespace unifex
   template <typename... Ts>
   struct unique_type_list_elements;
 
-  template <typename... Ts>
-  using unique_type_list_elements_t = typename unique_type_list_elements<Ts...>::type;
+  template <typename List>
+  using unique_type_list_elements_t = typename unique_type_list_elements<List, type_list<>>::type;
 
-  template <>
-  struct unique_type_list_elements<type_list<>> {
+  template <typename... Ts>
+  struct unique_type_list_elements<type_list<>, type_list<Ts...>> {
     using type = type_list<>;
   };
 
-  template <typename T, typename... Ts>
-  struct unique_type_list_elements<type_list<T, Ts...>> {
+  template <typename T, typename... Ts, typename... Us>
+  struct unique_type_list_elements<type_list<T, Ts...>, type_list<Us...>> {
     using type = conditional_t<
-      is_one_of_v<T, Ts...>,
-      unique_type_list_elements_t<type_list<Ts...>>,
-      concat_type_lists_t<type_list<T>, unique_type_list_elements_t<type_list<Ts...>>>
+      is_one_of_v<T, Us...>,
+      typename unique_type_list_elements<type_list<Ts...>, type_list<Us..., T>>::type,
+      concat_type_lists_t<type_list<T>, typename unique_type_list_elements<type_list<Ts...>, type_list<Us..., T>>::type>
     >;
   };
 
--- a/include/unifex/type_list.hpp
+++ b/include/unifex/type_list.hpp
@@ -95,13 +95,40 @@ namespace unifex
   template <typename... UniqueLists>
   using concat_type_lists_t = typename concat_type_lists<UniqueLists...>::type;
 
+  // unique_type_list_elements<type_list<Ts...>>
+  //
+  // Result is produced via '::type' which will contain a type_list<Ts...> that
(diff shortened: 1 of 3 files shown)


test/upon_error_test.cpp
```diff
@@ -170,6 +170,16 @@ TEST(UponError, ManyErrorSender) {
   >);
 }
 
+TEST(UponError, ManyErrorSenderAllReturnInt) {
+  auto s = many_error_sender{} | upon_error([](auto) -> int {
+    return 0;
+  });
+  static_assert(std::is_same_v<
+    decltype(s)::value_types<std::variant, std::tuple>,
+    std::variant<std::tuple<double>, std::tuple<int>>
```
ccotter, line 179: My main goal of this PR was to fix the output sender from `upon_error` to not contain duplicate values in `value_types`. I chose to change the behavior of `concat_type_lists_unique_t`, but I'm happy to take another approach (such as directly changing upon_error and possibly other algos that might have a similar problem).

test/type_list_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+    unique_type_list_elements_t<type_list<int>>,
+    type_list<int>
+  >);
+  static_assert(std::is_same_v<
+    unique_type_list_elements_t<type_list<int, int>>,
+    type_list<int>
+  >);
+  static_assert(std::is_same_v<
+    unique_type_list_elements_t<type_list<int, double, int>>,
+    type_list<double, int>
```
ccotter, line 20: Note that my hasty implementation of `unique_type_list_elements_t` does not preserve order, or rather, it removes duplicate items from the front rather than end of list.

test/type_list_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+void verify_concat_type_lists_unique() {
+  static_assert(std::is_same_v<
+    concat_type_lists_unique_t<type_list<int, int>, type_list<int>>,
+    type_list<int>
+  >);
+  static_assert(std::is_same_v<
+    concat_type_lists_unique_t<type_list<int>, type_list<int, int>>,
+    type_list<int>
+  >);
+}
```
ispeters, line 45: I think you're claiming that order is preserved so maybe one more test?
```suggestion
  static_assert(std::is_same_v<
    concat_type_lists_unique_t<type_list<bool, int, double>, type_list<double, int, float>>,
    type_list<bool, int, double, float>
  >);
}
```

ispeters (changes requested):
Mostly looks good; just a couple of nits, really.

include/unifex/type_list.hpp
```diff
@@ -95,14 +95,41 @@ namespace unifex
...
+  // contains the unique elements from the input type_list.
+  template <typename... Ts>
+  struct unique_type_list_elements;
+
+  template <typename List>
+  using unique_type_list_elements_t = typename unique_type_list_elements<List, type_list<>>::type;
+
+  template <typename... Ts>
+  struct unique_type_list_elements<type_list<>, type_list<Ts...>> {
+    using type = type_list<>;
```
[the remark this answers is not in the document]
ccotter: Added a comment and gave a more clear name for the pack.

include/unifex/type_list.hpp
```diff
@@ -95,14 +95,41 @@ namespace unifex
...
+  };
+
+  template <typename T, typename... Ts, typename... SeenElements>
+  struct unique_type_list_elements<type_list<T, Ts...>, type_list<SeenElements...>> {
+    using type = conditional_t<
+      is_one_of_v<T, SeenElements...>,
+      typename unique_type_list_elements<type_list<Ts...>, type_list<SeenElements..., T>>::type,
```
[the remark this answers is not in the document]
ccotter: Indeed, good catch!

ispeters (approved):
Nice!

Issue #548, opened 2023-07-03
Look into using clang-tidy


ispeters:
@ccotter suggested on PR #545 that `clang-tidy` might be able to catch the bugs we've seen with incorrect forwards, perhaps at the expense of using `std::forward`.  Automated bug-catching is nice, and probably worth _some_ amount of build speed degradation so we should look into some `clang-tidy` CI.

ccotter:
I'd be happy to do a first pass at this and report back. One of my side hobbies it tinkering with clang-tidy and have some experience with the checks that verify correct uses of `move`, `forward`, and universal references.

lewissbaker:
Note that on modern versions of clang std::forward/move now get translated to intrinsics which shouldn’t have the same template instantiation overhead.

ccotter:
Ok, these two checks are of interest (both of which would appear in the next LLVM release version 17).
 - [`cppcoreguidelines-rvalue-reference-param-not-moved`](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/rvalue-reference-param-not-moved.html)
 - [`cppcoreguidelines`](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/missing-std-forward.html)

Both help identify all cases where functions do not use `std::move(rvalue_ref_param)` or `static_cast<T&&>(universal_ref_param)` (or `std::forward<T&&>(universal_ref_param)` if you end up preferring that coding style). E.g.,

```
/home/ccotter/git/libunifex/include/unifex/async_mutex.hpp:119:46: warning: rvalue reference parameter 's' is never moved from inside the function body [cppcoreguidelines-rvalue-reference-param-not-moved]
  119 |     tag_invoke(tag_t<connect>, lock_sender &&s, Receiver &&r) noexcept {
      |                                              ^
/home/ccotter/git/libunifex/include/unifex/async_mutex.hpp:119:60: warning: forwarding reference parameter 'r' is never forwarded inside the function body [cppcoreguidelines-missing-std-forward]
  119 |     tag_invoke(tag_t<connect>, lock_sender &&s, Receiver &&r) noexcept {
```

Both checks are expecting code to always use the "named cast" using `move` or `forward`. The checks won't emit any kind of fixits though. Below are the counts of each diagnostic for all header includes:

```
     36 [cppcoreguidelines-rvalue-reference-param-not-moved]
    635 [cppcoreguidelines-missing-std-forward]
```

[...]

PR #549, opened 2023-07-03
Replace incorrect std::forward usage

ccotter:
Each of these looks like they were meant tobe a move.

### DIFF
--- a/include/unifex/let_value_with_stop_source.hpp
+++ b/include/unifex/let_value_with_stop_source.hpp
@@ -51,7 +51,7 @@ template<typename Operation, typename Receiver>
 class _stop_source_receiver<Operation, Receiver>::type {
 public:
     explicit type(Operation& op, Receiver&& r) :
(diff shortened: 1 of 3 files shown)


ccotter:
I've got an unmerged clang-tidy check in my local build that finds places where `forward` is used on a non-forwarding reference. Ref: #548

include/unifex/let_value_with_stop_source.hpp
```diff
@@ -51,7 +51,7 @@ template<typename Operation, typename Receiver>
 class _stop_source_receiver<Operation, Receiver>::type {
 public:
     explicit type(Operation& op, Receiver&& r) :
-        op_(op), receiver_{std::forward<Receiver>(r)}
+        op_(op), receiver_{std::move(r)}
```
ispeters, line 54: This is better than it was, but this whole file looks like it needs to have the value categories of the various forwarded arguments reconsidered.

ispeters (approved):
Nice clean-up.

PR #551, opened 2023-07-05
Fix bugprone-move-forwarding-reference violations

ccotter:


include/unifex/v2/async_scope.hpp
```diff
@@ -264,18 +264,18 @@ struct _nest_receiver<Sender, Receiver>::type final {
...
+        unifex::set_error(std::forward<decltype(receiver)>(receiver), std::current_exception());
       }
     });
   }
 
   template <typename E>
   void set_error(E e) noexcept {
     complete([&](auto&& receiver) noexcept {
-      unifex::set_error(std::move(receiver), std::move(e));
+      unifex::set_error(std::forward<decltype(receiver)>(receiver), std::move(e));
```
ccotter, line 278: This is semi debatable. In the letter of the law (the law being [CppCoreGuidelines F.19](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-forward) checked by [bugprone-move-forwarding-reference](https://releases.llvm.org/11.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-move-forwarding-reference.html), this should use a `forward`. I think receivers usually or always are passed by rvalue reference within the `set_error` CPO, but if someone somehow ends up trying to pass an lvalue ref to a receiver that does or doesn't accept by lvalue ref on the receiver, this code is not taking an opinion on way or the other - the bug would be elsewhere.

ispeters (changes requested):
Comparing the build times on this PR's CI to those on #552, it looks like switching to `std::forward<>()` isn't very expensive; comparable runs vary by about +/- 30 seconds in both directions.  We may as well switch—`std::forward<>()` conveys intent better than some arbitrary `static_cast`.

Please just update the lambdas passed to `complete` not to take `auto&&` so we can stick with `std::move()` instead of silencing the linter with an over-general `std::forward<>()`.

ispeters (approved):

ispeters:
I can't figure out how to squash this nicely from my phone. I'll merge it once the kids are in bed.

PR #552, opened 2023-07-05
Replace C-style cast with move() in let* algos

ccotter:


include/unifex/let_value_with.hpp
```diff
@@ -100,20 +102,18 @@ template<typename StateFactory, typename SuccessorFactory, typename Receiver>
 struct _operation<StateFactory, SuccessorFactory, Receiver>::type {
   template <typename StateFactory2, typename SuccessorFactory2, typename Receiver2>
   type(StateFactory2&& stateFactory, SuccessorFactory2&& func, Receiver2&& r) :
-    stateFactory_(static_cast<StateFactory2&&>(stateFactory)),
     func_(static_cast<SuccessorFactory2&&>(func)),
-    state_(static_cast<StateFactory&&>(stateFactory_)()),
+    state_(std::forward<StateFactory2>(stateFactory)()),
```
ccotter, line 106: The `stateFactory_` member is not actually used after the constructor, so I removed it.

ispeters: It _is_ used in the sense that the `stateFactory_` might be stateful (e.g. a lambda with captures). You'll need to put this back.

ispeters (changes requested):
Looks great except that we need to keep `stateFactory_` in case it's a stateful callable.

include/unifex/let_value_with.hpp
ccotter: I should have realized the code was not accidentally written that way :P

include/unifex/let_done.hpp
```diff
@@ -227,7 +227,7 @@ class _op<Source, Done, Receiver>::type {
   , receiver_((Receiver2&&)dest)
   {
     unifex::activate_union_member_with(sourceOp_, [&] {
-        return unifex::connect((Source&&)source, source_receiver{this});
+        return unifex::connect(std::move(source), source_receiver{this});
```
ispeters, line 230: Somehow this has broken the MSVC build.  I wonder if `Source` is sometimes an lvalue reference.

ispeters: Yup; line 324 instantiates an operation with an lvalue ref `Source` if the _Sender_ argument is an lvalue ref.  This is a `std::forward`.

_However_, most of the operation state types that I've written myself have a templated constructor that takes a forwarding reference to the _Sender_, rather than templating the whole type on a forwarding reference.  It seems like there'd be less duplicated code in the resulting binary if we could follow that pattern.

test/let_value_test.cpp
```diff
@@ -281,3 +284,9 @@ TEST(Let, LetValueWithTraitlessPredecessor) {
   ASSERT_TRUE(ret);
   EXPECT_EQ(*ret, 42);
 }
+
+TEST(Let, LvalueConnectable) {
+  sync_wait(repeat_effect_until(
+      let_value(let_done(just(), [] { return just(); }), [] { return just(); }),
+    [n=0]() mutable noexcept { return n++ == 1000; }));
+}
```
ispeters, line 292: I think the Windows CI that failed is blowing the stack here somehow.  The SEH exception that's being reported seems to correspond to `STATUS_STACK_OVERFLOW`.

If I understand the intent of this test correctly, you don't need 1,000 iterations, just more than 1. Maybe lower the iteration count and see if the test passes?
```suggestion
TEST(Let, LvalueConnectable) {
  sync_wait(repeat_effect_until(
      let_value(let_done(just(), [] { return just(); }), [] { return just(); }),
    [n=0]() mutable noexcept { return n++ == 5; }));
}
```

include/unifex/let_error.hpp
```diff
@@ -256,15 +256,15 @@ class _op<Source, Func, Receiver>::type final {
   using final_receiver = final_receiver_type<Source, Func, Receiver, Error>;
 
 public:
-  template <typename Func2, typename Receiver2>
-  explicit type(Source&& source, Func2&& func, Receiver2&& dest) noexcept(
+  template <typename Source2, typename Func2, typename Receiver2>
+  explicit type(Source2&& source, Func2&& func, Receiver2&& dest) noexcept(
```
ispeters, line 260: By a similar argument to the above, in `let_done.hpp`, I think that this expresses more generality than is possible given the change below, with the same curiosity about `clang-tidy`'s attitude towards using `forward` with a class template parameter.

include/unifex/let_value.hpp
```diff
@@ -227,16 +228,16 @@ struct _op<Predecessor, SuccessorFactory, Receiver>::type {
   template <typename Operation, typename... Values>
   friend struct _successor_receiver;
 
-  template <typename SuccessorFactory2, typename Receiver2>
+  template <typename Predecessor2, typename SuccessorFactory2, typename Receiver2>
```
ispeters, line 231: Same as above in `let_done` and `let_error`.  I think lines 378-379 are hand-rolling the `member_t<Self, Predecessor>` type alias with this:
```
decltype((static_cast<Sender&&>(sender).pred_)
```

include/unifex/let_value_with_stop_source.hpp
```diff
@@ -165,7 +167,7 @@ struct _stop_source_operation<SuccessorFactory, Receiver>::type {
           nothrow_connectable) {
     return unifex::connect(
         static_cast<SuccessorFactory&&>(func)(stopSource_),
```
ispeters, line 169: This should probably be a `move`, too, which might mean it should be taken by rvalue reference.
```suggestion
        std::move(func)(stopSource_),
```

ispeters (commented):
Looks good, but I've got questions about what `clang-tidy` does and does not like.

include/unifex/let_done.hpp
```diff
@@ -218,16 +218,16 @@ class _op<Source, Done, Receiver>::type {
...
+  explicit type(Source2&& source, Done2&& done, Receiver2&& dest)
       noexcept(std::is_nothrow_move_constructible_v<Receiver> &&
                std::is_nothrow_move_constructible_v<Done> &&
                is_nothrow_connectable_v<Source, source_receiver>)
   : done_((Done2&&)done)
   , receiver_((Receiver2&&)dest)
   {
     unifex::activate_union_member_with(sourceOp_, [&] {
-        return unifex::connect((Source&&)source, source_receiver{this});
+        return unifex::connect(std::forward<Source2>(source), source_receiver{this});
```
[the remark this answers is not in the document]
ccotter: > Does clang-tidy complain about the following code?
 
To be up front, I implemented some of these checks in clang-tidy, including the `cppcoreguidelines-rvalue-reference-param-not-moved` check (which will be released in the next major version release). [This code](https://godbolt.org/z/5YrYKrKMc) by default complains, but there is an option `IgnoreNonDeducedTemplateTypes` that can be set to true which [silences this warning](https://godbolt.org/z/oe8zfG14e). Whether to warn on code such as this did come up in review, and it hinges on the meaning of the [CppCoreGuidelines F.18 enforcement](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f18-for-will-move-from-parameters-pass-by-x-and-stdmove-the-parameter):

 > Flag all X&& parameters (where X is not a template type parameter name) where the function body uses them without std::move.

[This is the discussion](https://reviews.llvm.org/D141569#inline-1395526) regarding this situation. Re-reading that convo and the CppCoreGuideline, it seems I interpreted the guideline to mean that non-deducing contexts are not considered in the *where X is not a template type parameter name* text. That doesn't quite seem right. Perhaps worth clarification from the guidelines authors, which I'll follow up on. At the very least, it seems that changing the default option `IgnoreNonDeducedTemplateTypes` to be true rather than defaulting false would make sense.

There is another related guideline [ES.56](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es56-write-stdmove-only-when-you-need-to-explicitly-move-an-object-to-another-scope) which has in its enforcement section

 > Flag when std::forward is applied to an rvalue reference (X&& where X is a non-template parameter type). Use std::move instead.
 > Flag when std::forward is applied to other than a forwarding reference. (More general case of the previous rule to cover the non-moving cases.)

I have always understood "a forwarding reference" to mean a reference deduced in a function call of the form `T&&` (which I think is confirmed in https://timsong-cpp.github.io/cppwp/n4861/temp.deduct.call#2.3. 

```
template <class T>
struct AStruct {
    template <class U>
    void something(T&& t, U&& u) {} // T&& is not a forwarding reference, but U&& is
};
```

So, `std::forward` is not valid here (in the opinion of the CppCoreGuidelines, and a potential clang-tidy check I have ready to submit for review).

## Summary

 > Does clang-tidy complain about the following code?
 
Yes, for the `cppcoreguidelines-rvalue-reference-param-not-moved` check, but it's configurable, and it also might be a bug in the tool not faithfully following the guideline, which I'll follow up on.

> I ask because that's what I'd intended, but maybe clang-tidy thinks it's an erroneous use of std::forward.
 
Right, the code appears to not follow another CppCoreGuideline (for which there is no clang-tidy check yet, but hopefully will include on in the future) that `forward` should only be used on "forwarding references," and `Source` is not considered a forwarding reference in this code.

## Next steps

Now, the CppCoreGuidelines are opinionated and intend to offer a set of "best practices" of sorts. Is it necessary to update let_done and others to have the operation constructor take a forwarding reference `Source2&&` and use `forward`, instead of how the code was originally written? I don't htink so, and I think we can leave the code as is (if that's the preferred idiom you like, as it seems to be in working order in the let* algos) and still have the code pass clang-tidy (with appropriate options, and/or updating default option values in the tool itself). We'd have to use `static_cast<Source&&>(source)` instead of `forward` though, but that's "OK" in that the code in this case is using a non-CppCoreGuideline (and C++ standards I think) definition of "forwarding reference."

Let me know if I've understood things here. I appreciate you reviewing these (it's helped me "field test" these relatively new clang-tidy checks on real code, which is valuable feedback for the tool).

include/unifex/let_value.hpp
ccotter: > lines 378-379 are hand-rolling the member_t<Self, Predecessor> type alias with this:
 
 Yep, I noticed this too, and added a test for "l value connectable"-ness though nothing had to be changed for this test to pass.

include/unifex/let_done.hpp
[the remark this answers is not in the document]
ispeters: > I implemented some of these checks in clang-tidy,

Oh, cool!

> At the very least, it seems that changing the default option `IgnoreNonDeducedTemplateTypes` to be true rather than defaulting false would make sense.

I wouldn't go that far.  I think Unifex is pretty weird C++.  The only other place I've seen a similar "saving for later" of a forwarding reference is a single example in some pretty obscure internal code.

> I have always understood "a forwarding reference" to mean a reference deduced in a function call of the form `T&&`

Yeah, same.

> Right, the code appears to not follow another CppCoreGuideline … that forward should only be used on "forwarding references," and Source is not considered a forwarding reference in this code.

Agreed, this isn't a traditional forwarding reference.

Regarding your next steps:
 - I think we're on the same page that, given that the operation state class templates are sometimes instantiated with lvalue reference types for the source _Sender_, the templated constructor that takes a forwarding reference is obscuring what's really going on.
 - I don't _really_ like any of the options you've presented, but I think the least-bad is the `static_cast`.

Is there a way to suppress a `clang-tidy` warning for a given line of code, a bit like using `#pragma` to suppress compiler warnings?  If there is, I think my preferred choice would be to use `std::forward` but suppress the "this isn't a forwarding reference" warning with an explanation.

As I alluded to above, I think of this case as a forwarding reference that's been "saved for later", almost like function currying, or a "lambda capture but for value category."  As far as I know, this is an unusual pattern so I wouldn't go changing tool defaults based on it—I wouldn't even configure the tool for this library to silence the warning in this case.  Absent an omniscient linter, I think this is a genuine case of not matching the recommended usage pattern and so a warning is appropriate.

I think this code deserves a comment; I had to reverse-engineer what it's doing to render you a decent code review.  There's a fair amount of Unifex that's like that; we've inherited a prototype that I shipped to prod so I now have different optimization goals than the original authors did.

# Summary
1. I wouldn't suggest you change your tool's defaults.
2. I would configure your tool to warn on this case for this library.
3. My preferred change is to use `std::forward` and disable the warning locally.
    - If local warning disablement is not possible, I'd use `static_cast`.
5. Regardless of the approach in 3, I think there should be a comment next to the cast that says something about how the not-really-a-forwarding-reference is effectively a forwarding reference that's been "saved for later" and now is later.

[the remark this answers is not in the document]
ccotter: Thanks for the detailed response.

We can go with `std::forward` with a local disablement. clang-tidy respects `// NOLINT(<optional check name>)` comments.

```
    return unifex::connect(
        std::forward<Source>(source),  // NOLINT(cppcoreguidelines-forward-non-forwarding-parameter)
        source_receiver{this});

// or, ignore all warnings

    return unifex::connect(std::forward<Source>(source), source_receiver{this}); // NOLINT
```

That said, `cppcoreguidelines-forward-non-forwarding-parameter` does not exist yet (hopefully that'll be an accepted check, possibly with a different name). As that check does not exist yet, I'll leave a comment, but that comment can later be changed (and/or kept) to a NOLINT for the specific warning.

include/unifex/let_value_with_stop_source.hpp
ccotter: `SuccessorFactory&&` also seems like a "saved for later forwarding reference," although it's used differently than in the other algos. Do we need `func_` to remain alive for the duration of the operation? In the operation constructor, we currently have (annotated with `// comment`)

```
template <typename SuccessorFactory2, typename Receiver2>
  type(SuccessorFactory2&& func, Receiver2&& r) noexcept(
      std::is_nothrow_constructible_v<SuccessorFactory, SuccessorFactory2>&&
          std::is_nothrow_constructible_v<Receiver, Receiver2>&& noexcept(
              connect_inner_op(func, (Receiver2 &&) r))) // borrow func, whether it's an lvalue ref or rvalue erf
    : func_{(SuccessorFactory2 &&) func} // forward func into func_
    , receiverToken_(get_stop_token(r))
    , innerOp_(connect_inner_op(func_, (Receiver2 &&) r)) {} // borrow func_, but don't move so that func_ remains alive
```

`SuccessorFactory&` in `connect_inner_op` seems correct so that it borrows but doesn't possibly move from.

PR #561, opened 2023-09-27
Fix exception assertion bug in let_value and add test cases

jesswong:
While we were attempting to downstream unifex, there was a bug where the assertion(op.cleanup_ != nullptr) would be trigged in let_value because we were not checking if the given function was throwing an exception or not. This fixes that issue by adding the exception checks and by adding test cases to catch issues like this in the future.

I validated that the test case no longer fails by running the new test on a revision where the fix was not present and that the test case failed.

### DIFF
--- a/test/let_value_test.cpp
+++ b/test/let_value_test.cpp
@@ -13,17 +13,20 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include <unifex/just.hpp>
 #include <unifex/let_value.hpp>
+
+#include <unifex/allocate.hpp>
+#include <unifex/just.hpp>
+#include <unifex/just_done.hpp>
+#include <unifex/just_error.hpp>
+#include <unifex/just_void_or_done.hpp>
+#include <unifex/let_error.hpp>
 #include <unifex/let_value_with.hpp>
 #include <unifex/scheduler_concepts.hpp>
 #include <unifex/sync_wait.hpp>
-#include <unifex/timed_single_thread_context.hpp>
 #include <unifex/then.hpp>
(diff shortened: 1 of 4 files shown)


test/let_value_test.cpp
```diff
@@ -264,6 +264,20 @@ TEST(Let, SimpleLetValueErrorWithAllocate) {
...
 }
 
+TEST(Let, LetValueSuccessorWithException) {
+  EXPECT_THROW(sync_wait(unifex::just() | unifex::let_value([]() {
+    throw std::runtime_error("Throwing error for testing purposes");
+    return unifex::just();
+  })), std::runtime_error);
+}
+
+TEST(Let, LetValueWithException) {
```
ispeters, line 274: I think I'd leave this test out; it's semantically the same as the above test, and I think we have other tests that confirm the piping semantics.

test/let_value_test.cpp
```diff
@@ -264,6 +264,20 @@ TEST(Let, SimpleLetValueErrorWithAllocate) {
   })), std::invalid_argument);
 }
 
+TEST(Let, LetValueSuccessorWithException) {
+  EXPECT_THROW(sync_wait(unifex::just() | unifex::let_value([]() {
+    throw std::runtime_error("Throwing error for testing purposes");
+    return unifex::just();
```
ispeters, line 270: I'm a little surprised that the compiler doesn't complain this is dead code….

ispeters (approved):
Nice!

test/let_value_test.cpp
jesswong: Yeah, it won't compile without it :/

Issue #547, opened 2023-06-29
Lifetime issues in `type_erased_stream`


janondrusek:
1. stop callbacks should be destroyed in `set_` functions before the result is delivered
2. cancellation requests may race with other completion signals, I observed an infrequent segfault

The fix is non-trivial and will likely follow the ref-counting mechanism introduced in `when_all`.

PR #565, opened 2023-10-04
fix cancellation lifetime issue in `type_erased_stream`

janondrusek:
fix cancellation lifetime issue in `type_erased_stream` (#547)
    
* operation state / stop callback race for completion
* ref counting mechanism similar to `when_all`

### DIFF
--- a/include/unifex/type_erased_stream.hpp
+++ b/include/unifex/type_erased_stream.hpp
@@ -28,6 +28,8 @@
 #include <unifex/bind_back.hpp>
 #include <unifex/exception.hpp>
 
+#include <atomic>
+
 #include <unifex/detail/prologue.hpp>
 
 namespace unifex {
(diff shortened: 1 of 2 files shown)


include/unifex/type_erased_stream.hpp
```diff
@@ -104,24 +104,36 @@ struct _stream<Values...>::type final {
     virtual void start_cleanup(cleanup_receiver_base& receiver) noexcept = 0;
   };
 
+  struct next_op_base {
+    virtual ~next_op_base() = default;
+    virtual bool complete() noexcept = 0;
```
ispeters, line 109: We don't need a virtual destructor because no one's going to delete through a pointer-to-base; you can enforce that with a non-public, non-virtual, defaulted destructor.
```suggestion
    virtual bool complete() noexcept = 0;
   protected:
    next_op_base() noexcept = default;
    next_op_nase(next_op_base&&) = delete;
    ~next_op_base() = default;
```

include/unifex/type_erased_stream.hpp
```diff
@@ -104,24 +104,36 @@ struct _stream<Values...>::type final {
...
   template <typename Receiver>
   struct _next_receiver final {
     struct type final : next_receiver_base {
       UNIFEX_NO_UNIQUE_ADDRESS Receiver receiver_;
+      next_op_base* op_;
 
-      explicit type(Receiver&& receiver)
-        : receiver_((Receiver &&) receiver) {}
+      explicit type(Receiver&& receiver, next_op_base* op)
+        : receiver_((Receiver &&) receiver), op_(op) {}
```
ispeters, line 119: It occurs to me that you could save on the virtual function overhead if you passed the address of the `atomic<char> refcount` instead of the whole op state.

ispeters (approved):

PR #570, opened 2023-10-18
add http server example with `io_uring`

janondrusek:
basic building blocks are coroutines, `async_scope` and _Stream_

* single threaded
* only supports `GET`
* serves static `<html>` on port `8080`, both IPv4 and IPv6
* shuts down after 1000ms by default
* stress tested via `ab -c 512 -n $((8*4096)) http://localhost:8080/`

image

### DIFF
--- a/examples/linux/http_server_io_uring_test.cpp
+++ b/examples/linux/http_server_io_uring_test.cpp
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


examples/linux/http_server_io_uring_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+      [&mainThread, &fixme](auto readWriteFile) {
+        spawn_detached(on(mainThread, handle(std::move(readWriteFile))), fixme);
+      });
+  co_await fixme.join();
+}
+
+task<void> quit(io_uring_context::scheduler sched) {
+  auto in = open_file_read_only(sched, "/dev/stdin");
+  std::array<char, buffer_size> buffer;
+  // TODO make async_read_some_at stoppable
```
ispeters, line 145: I think this is TODONE.

ispeters (approved):
I love it!

PR #573, opened 2023-10-20
add gcc-11...gcc-13

janondrusek:
* tweak noexcept constructors in tests due to a regression in GCC 11+
* tweak test build failures
* enable CI steps
* bump GTest to 1.14

PR #574, opened 2023-10-20
Fix dangling temporary warning in gcc 13

ispeters:
This diff inlines calls to `extract_this` to convince GCC 13 that there are no dangling temporaries.


ispeters:
@janondrusek included this change in #573

Issue #577, opened 2023-10-25
`task` resumes after being cancelled causing UB


glebov-andrey:
The error is reproducible by the `SynchronousCompletionFromReceiverStopRequestIsASANSafe` test if an assertion is added to `unifex::_task::_result_and_unhandled_exception::type::result()` before returning from the function https://github.com/facebookexperimental/libunifex/blob/cf735485832b69b7699b3812d7b799cb006a9222/include/unifex/task.hpp#L241-L244
as such:
```c++
UNIFEX_ASSERT(result_.state_ == _state::value);
return std::move(result_.value_).get();
```
<details><summary>Stack trace</summary>

```
unifex::_task::_result_and_unhandled_exception<void, false>::type::result task.hpp:244
unifex::_task::_awaiter<unifex::_task::_sr_thunk_promise<void>::type, unifex::_await::_sender_task<unifex::_when_all::_element_receiver<0, unifex::_let_v_w_stop_src::_stop_source_receiver<unifex::_let_v_w_stop_src::_stop_source_operation<StopWhen_SynchronousCompletionFromReceiverStopRequestIsASANSafe_Test::TestBody()::<lambda(auto:55&)>&&, unifex::_sync_wait::_receiver<std::tuple<std::variant<std::tuple<> >, std::variant<std::tuple<> > > >::type>::type, unifex::_sync_wait::_receiver<std::tuple<std::variant<std::tuple<> >, std::variant<std::tuple<> > > >::type>::type, unifex::_task::_task<void, false>::type&&, unifex::_then::_sender<unifex::_just::_sender<>::type, StopWhen_SynchronousCompletionFromReceiverStopRequestIsASANSafe_Test::TestBody()::<lambda(auto:55&)>::<lambda()> >::type&&>::type>::type::promise_type>::type::await_resume(void) task.hpp:600
[...]

PR #579, opened 2023-10-26
Fix UB in cancelled tasks again

ispeters:
As reported in #577, #578 didn't fix the UB in cancelled tasks. There were two bugs in that "fix":
 1. we were writing the coroutine handle to resume *after* synchronizing between the two racing operation states, leading to a data race when the deferred stop source completed second; and
 2. it's wrong to call `continuation_.done()` to store the coroutine handle to be resumed later because `done()` completes the operation with `set_done()` as a side effect.

This diff fixes both issues. Instead of storing a coroutine handle to be resumed, we store an enum value that describes which coroutine handle to resume (so we don't eagerly invoke `done()`), and we do the store before synchronizing to eliminate the data race.

I've added a unit test that fails with a TSAN-detected data race without the fix, and the fix silences the TSAN error.

I also clang-formatted `task.hpp` and the modified test file.

### DIFF
--- a/include/unifex/task.hpp
+++ b/include/unifex/task.hpp
@@ -222,8 +222,9 @@ struct _result_and_unhandled_exception final {
    * This is used to share the implementation of result handling between
    * type-specific promise types.
    */
-  struct type { //parametize on noexcept, noexcept terminates in body
-    void unhandled_exception() noexcept { // will be invoked in the catch block of the coroutine
+  struct type {  // parametize on noexcept, noexcept terminates in body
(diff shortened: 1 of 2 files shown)


ispeters:
@glebov-andrey, do you want to patch this in and see if it fixes your issue?

janondrusek (approved):

glebov-andrey:
Looks good, thanks!
All our tests pass and I haven't been able to cause any more crashes during cancellation with this patch and ASAN, UBSAN or TSAN.

Issue #485, opened 2023-02-03
update `.clang-format` and re-format everything


janondrusek:
1. do not put `case state_t::CONSTRUCTED: just = false; break;` on a single line
2. [avoid adding spaces](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#indentppdirectives) `#  include` for nested `#`
3. ...

PR #580, opened 2023-10-31
Clang format all but the Win32-specific files

ispeters:
This giant stack of changes is the result of running `clang-format` on everything but our Win32-specific files. For some reason, formatting `win32/windows_thread_pool.hpp` breaks the MSVC build. I'll tackle formatting those files in another PR.

This partly addresses #485. I haven't updated our `.clang-format` file, just applied the existing one. I figure it'll be a smaller, easier change to update the rules on a consistently-formatted repo, and having a consistently-formatted repo means I won't repeat the mistake of #579 where I accidentally included unrelated formatting changes in a behaviour-changing PR.

### DIFF
--- a/include/unifex/detail/any_heap_allocated_storage.hpp
+++ b/include/unifex/detail/any_heap_allocated_storage.hpp
@@ -24,141 +24,136 @@
 
 #include <unifex/detail/prologue.hpp>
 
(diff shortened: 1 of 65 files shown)


janondrusek (approved):
Huge!

PR #584, opened 2023-11-01
Format examples

ispeters:
This diff formats `examples/*.cpp`. Depends on #582.

### DIFF
--- a/examples/async_mutex.cpp
+++ b/examples/async_mutex.cpp
@@ -21,12 +21,12 @@
 
 #if !UNIFEX_NO_COROUTINES
 
-#include <unifex/scheduler_concepts.hpp>
-#include <unifex/task.hpp>
-#include <unifex/when_all.hpp>
-#include <unifex/single_thread_context.hpp>
+#  include <unifex/scheduler_concepts.hpp>
+#  include <unifex/single_thread_context.hpp>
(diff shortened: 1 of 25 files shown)


examples/async_mutex.cpp
```diff
@@ -21,12 +21,12 @@
...
-#include <unifex/task.hpp>
-#include <unifex/when_all.hpp>
-#include <unifex/single_thread_context.hpp>
+#  include <unifex/scheduler_concepts.hpp>
+#  include <unifex/single_thread_context.hpp>
+#  include <unifex/task.hpp>
+#  include <unifex/when_all.hpp>
 
-#include <cstdio>
+#  include <cstdio>
```
janondrusek, line 29: eh ...

janondrusek (approved):

examples/async_mutex.cpp
ispeters: Yeah. I'm figuring that it'll be easier to tweak the formatting rules and re-format again now that we know the auto-formatted code compiles and passes all the tests. If we want to change this formatting rule and fully resolve #485, we'll be able to do that with `clang-format -i $(find . -name \*.cpp -o -name \*.hpp)` and not worry about breaking anything.

PR #593, opened 2024-01-03
Allow copying/moving variant_sender

ccotter:
variant_sender's perfect forwarding constructor hid the move and copy constructors.


ispeters:
Thanks for the fix, and sorry about the huge delay on merging.

Issue #586, opened 2023-12-05
(Is there already) support for async stream


georgikoyrushki95:
Hi, I also commented on #507 as it seems related (at least to someone not too familiar with S/R yet).

 A little bit of context of what I want: I am looking for something that represents an async stream, to sketch out an example:

```C++
unifex::task<size_t> getFooSum(/* some params here */)
{
    auto request = /* build request from params */

    // This range of responses is filled from the underlying RPC framework, responses are arriving on
    // a separate thread (what we've termed "the callback thread" internally.
    auto rangeOfMultipartResponses = someRpcClient->sendRequest(request);
    auto fooValueSum = rangeOfMultipartResponses
        unifex::reduce_stream(0, [](auto resp){ resp.fooValue(); });
    co_return co_await fooValueSum;
}
```

Is this something that the generic `range_stream` would support OR we need to build our own stream adapter? And isn't this something generic enough to be present in `libunifex`? Basically something like `cppcoro::async_generator` that speaks the stream protocol? I have seen `cppcoro::async_generator` being mentioned in the docs, but wasn't able to find anything that can support that functionality here.

georgikoyrushki95:
Also if not, should it be built in a way that it interoperates with `cppcoro::async_generator`. Basically being able to go both ways `cppcoro::async_generator` <-> `libunifex::async_stream`?

ispeters:
Hi @georgikoyrushki95, I was able to wrap a `cppcoro::async_generator<int>` in a _Stream_ here: https://godbolt.org/z/Kh3rWTMao.

I don't think it makes sense to add that particular _Stream_ type to Unifex because we intend to be free of dependencies, but a type that's analogous to `cppcoro::async_generator<T>` would be a welcome addition.

georgikoyrushki95:
@ispeters thanks so much! This is exactly what I'm looking for!

> I don't think it makes sense to add that particular Stream type to Unifex because we intend to be free of dependencies, but a type that's analogous to cppcoro::async_generator<T> would be a welcome addition.

For this, do you mean a libunifex-specific async generator and adapting that to a stream? Basically identical to what you have done in your godbolt, the only difference being we add the equivalent of `cppcoro::async_generator<T>` in libunifex as a prerequisite?

Or do you want the stream to be backed by "pure" senders?

Both would work for us & it appears it'd be equivalent for the consumers of the stream values. The difference would be in how the values are produced, I guess.

[...]

PR #596, opened 2024-01-12
PoC: async_generator as a stream #586

georgikoyrushki95:
Hi, I reached out before Christmas to check if there are any async streams in `unifex` (#586). [In the discussion within the issue](https://github.com/facebookexperimental/libunifex/issues/586#issuecomment-1845934903), we established it'd be good to have a coroutine-based async stream, similar to `cppcoro::async_generator`. 

In this PR, I have added a small PoC that achieves part of the requirements, listed by @ispeters (list with checbkoxes below). I would appreciate some feedback & see if you guys are interested in seeing something like this in `unifex`.

A summary of the requirements for an `async_generator` + whether or not they are addressed in this PR:

- [X] Support `co_await` and `co_yield` inside the generator coroutine.
    ❗ This achieved already by `cppcoro::async_generator`

- [ ] Feel like a `unifex::task<>`
     - [X] Scheduler affinity
             ❗ In short: 
             (a) resuming the generator resumes it in the scheduler it was running before suspension;
             (b) resuming the consumer coroutine resumes it on the scheduler it was running before `co_await`-ing the generator.
     - [X] Support awaiting senders
             ❗ Achieved via `await_transform` customization in the generator promise, similar to `unifex::task<>`
     - [ ] Support unwind-on-cancellation semantics / stop tokens
            ❗ I did not look into that. Looks like it might be too much for 1 PR.
     - [ ] `noexcept` version of it
            ❗ Shouldn't be too much to add to this PR. Can do, if there is interest

- [ ] Ability to let callers opt-out of scheduler affinity. This is achieved in `unifex::task<>` by having this `struct _sa_task`.
       ❗ Have not looked into that

At a high-level, the generator supports `next()` and `cleanup()` (so it is a stream in itself). In addition, I did not expose `begin()`, `end()`. The reason for this: it forces clients to use `iterator`-based for loops, at least without something like `co_await for_each(...)` construct in the standard. To me, iterator-based for loops are quite archaic & error-prone, so I would prefer not to expose something like that. 

High-level skeleton might be summarized as:

```C++

template <typename T>
struct async_generator {
   // from the consumer's POV, expose it ONLY as a stream.
   /* ... */ next();
   /* ... */ cleanup();
private:
  // Private begin()/end(), used within next()
   /* ... */ begin();
   /* ... */ end();
};

### DIFF
--- a/include/unifex/async_generator.hpp
+++ b/include/unifex/async_generator.hpp
@@ -1,0 +1,546 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// A lot of stuff from cppcoro
+
+#pragma once
+
+#if !UNIFEX_NO_COROUTINES
+
+#  include <unifex/any_scheduler.hpp>
+#  include <unifex/async_scope.hpp>
+#  include <unifex/at_coroutine_exit.hpp>
+#  include <unifex/await_transform.hpp>
+#  include <unifex/defer.hpp>
+#  include <unifex/inline_scheduler.hpp>
+#  include <unifex/just_void_or_done.hpp>
+#  include <unifex/task.hpp>
+#  include <unifex/with_scheduler_affinity.hpp>
+
+#  include <optional>
+
+#  include <unifex/detail/prologue.hpp>
+
+namespace unifex {
+
+template <typename T>
+class async_generator;
+
+namespace detail {
+template <typename T>
+class async_generator_iterator;
+class async_generator_yield_operation;
+class async_generator_advance_operation;
+
+class async_generator_promise_base {
+public:
+  async_generator_promise_base() noexcept : m_exception(nullptr) {
+    // Other variables left intentionally uninitialised as they're
+    // only referenced in certain states by which time they should
+    // have been initialised.
+  }
+
+  async_generator_promise_base(const async_generator_promise_base& other) =
+      delete;
+  async_generator_promise_base&
+  operator=(const async_generator_promise_base& other) = delete;
+
+  std::suspend_always initial_suspend() const noexcept { return {}; }
+
+  async_generator_yield_operation final_suspend() noexcept;
+
+  void unhandled_exception() noexcept {
+    m_exception = std::current_exception();
+  }
+
+  void return_void() noexcept {}
+
+  /// Query if the generator has reached the end of the sequence.
+  ///
+  /// Only valid to call after resuming from an awaited advance operation.
+  /// i.e. Either a begin() or iterator::operator++() operation.
+  bool finished() const noexcept { return m_currentValue == nullptr; }
+
+  void rethrow_if_unhandled_exception() {
+    if (m_exception) {
+      std::rethrow_exception(std::move(m_exception));
+    }
+  }
+
+  // This is the CP needed for other things in the ecosystem to be able to
+  // "access" the current scheduler of our promise.
+  friend unifex::any_scheduler tag_invoke(
+      unifex::tag_t<unifex::get_scheduler>,
+      const async_generator_promise_base& p) noexcept {
+    return p.sched_;
+  }
+
+  // This is needed for at_coroutine_exit to do the async clean up
+  friend unifex::continuation_handle<> tag_invoke(
+      const unifex::tag_t<unifex::exchange_continuation>&,
+      async_generator_promise_base& p,
+      unifex::continuation_handle<> action) noexcept {
+    return std::exchange(p.continuation_, std::move(action));
+  }
+
+  unifex::coro::coroutine_handle<> unhandled_done() noexcept {
+    return continuation_.done();
+  }
+
+protected:
+  template <typename P>
+  friend struct async_gen_initial_suspend;
+
+  async_generator_yield_operation internal_yield_value() noexcept;
+
+  // Needed for jumping back on the generator's scheduler, in cases
+  // where the consumer coroutine is executing elsewhere.
+  unifex::async_scope scope_;
+
+  // Keep track of the consumer scheduler
+  unifex::any_scheduler consumerSched_{_default_scheduler};
+  // The scheduler we currently run on
+  unifex::any_scheduler sched_{_default_scheduler};
+  bool rescheduledBefore_{false};
+
+private:
+  friend class async_generator_yield_operation;
+  friend class async_generator_advance_operation;
+
+  inline static constexpr unifex::inline_scheduler _default_scheduler{};
+
+  std::exception_ptr m_exception;
+
+  // In this case, this keeps the consumer coroutine + a done() continuation.
+  // it's needed for at_coroutine exit for now, but also whenw e handle stop
+  // requests.
+  unifex::continuation_handle<> continuation_;
+
+protected:
+  void* m_currentValue;
+};
+
+class async_generator_yield_operation final {
+public:
+  async_generator_yield_operation(
+      unifex::continuation_handle<> continuation) noexcept
+    : continuation_(continuation) {}
+
+  bool await_ready() const noexcept { return false; }
+
+  template <typename Promise>
+  unifex::coro::coroutine_handle<>
+  await_suspend([[maybe_unused]] unifex::coro::coroutine_handle<Promise>
+                    producer) noexcept {
+    // simplest case => no need to reschedule at all, just resume the cosumer
+    // coroutine
+    if (producer.promise().sched_ == producer.promise().consumerSched_) {
+      return continuation_.handle();
+    }
+
+    // need to reschedule back onto the consumer coro; kick off an async event &
+    // return no-op
+    producer.promise().scope_.detached_spawn_call_on(
+        producer.promise().consumerSched_,
+        [consumerCoro = continuation_.handle()]() noexcept {
+          consumerCoro.resume();
+        });
+    return unifex::coro::noop_coroutine();
+  }
+
+  void await_resume() noexcept {}
+
+private:
+  unifex::continuation_handle<> continuation_;
+};
+
+// await_suspend when we yield from the generator
+inline async_generator_yield_operation
+async_generator_promise_base::final_suspend() noexcept {
+  // The same is done for unifex::task (check the cpp). This was confusing to
+  // read, but all we're doing is at the very last suspend, we want to clear up
+  // the async scope and schedule back onto the consumer's schedule.
+  auto cleanupTask = unifex::at_coroutine_exit([this]() -> unifex::task<void> {
+    co_await scope_.complete();
+
+    if (consumerSched_ != sched_) {
+      co_await unifex::schedule(consumerSched_);
+    }
+  });
+
+  cleanupTask.await_suspend_impl_(*this);
+  (void)cleanupTask.await_resume();
+
+  m_currentValue = nullptr;
+  return internal_yield_value();
+}
+
+inline async_generator_yield_operation
+async_generator_promise_base::internal_yield_value() noexcept {
+  return async_generator_yield_operation{continuation_};
+}
+
+class async_generator_advance_operation {
+protected:
+  async_generator_advance_operation(std::nullptr_t) noexcept
+    : m_promise(nullptr)
+    , m_producerCoroutine(nullptr) {}
+
+  async_generator_advance_operation(
+      async_generator_promise_base& promise,
+      unifex::coro::coroutine_handle<> producerCoroutine) noexcept
+    : m_promise(std::addressof(promise))
+    , m_producerCoroutine(producerCoroutine) {}
+
+public:
+  bool await_ready() const noexcept { return false; }
+
+  // await_suspend during co_await ++itr;
+  template <typename Promise>
+  unifex::coro::coroutine_handle<> await_suspend(
+      unifex::coro::coroutine_handle<Promise> consumerCoroutine) noexcept {
+    m_promise->continuation_ = consumerCoroutine;
+
+    auto consumerScheduler = unifex::get_scheduler(consumerCoroutine.promise());
+
+    // simplest case => no need to reschedule at all, just resume the producer
+    // coroutine
+    if (consumerScheduler == m_promise->sched_) {
+      m_promise->consumerSched_ = consumerScheduler;
+      return m_producerCoroutine;
+    }
+
+    // consumerScheduler != producerScheduler and the generator hasn't been
+    // rescheduled
+    // => continue executing on the consumer's schedule
+    if (!m_promise->rescheduledBefore_) {
+      m_promise->consumerSched_ = consumerScheduler;
+      m_promise->sched_ = consumerScheduler;
+      return m_producerCoroutine;
+    }
+
+    // consumerScheduler != producerScheduler and the generator has been
+    // rescheduled => we need to resume onto the generator's scheduler; return
+    // no-op + kick off an async event to hop onto the generator's scheduler
+    m_promise->scope_.detached_spawn_call_on(
+        m_promise->sched_,
+        [prodCoro = unifex::coro::coroutine_handle<
+             async_generator_promise_base>::from_promise(*m_promise),
+         consumerSched = std::move(consumerScheduler)]() noexcept {
+          // update consumerSched_ so we can re-hop onto the correct scheduler
+          prodCoro.promise().consumerSched_ = consumerSched;
+          // prodCoro.promise().sched_ already points to the correct scheduler
+          prodCoro.resume();
+        });
+    return unifex::coro::noop_coroutine();
+  }
+
+protected:
+  async_generator_promise_base* m_promise;
+  unifex::coro::coroutine_handle<> m_producerCoroutine;
+};
+
+template <typename T>
+class async_generator_promise final : public async_generator_promise_base {
+  using value_type = std::remove_reference_t<T>;
+  using Promise = async_generator_promise<T>;
+
+public:
+  async_generator_promise() noexcept = default;
+
+  async_generator<T> get_return_object() noexcept;
+
+  async_generator_yield_operation yield_value(value_type& value) noexcept {
+    m_currentValue = std::addressof(value);
+    return internal_yield_value();
+  }
+
+  async_generator_yield_operation yield_value(value_type&& value) noexcept {
+    return yield_value(value);
+  }
+
+  T& value() const noexcept { return *static_cast<T*>(m_currentValue); }
+
+  template <typename Value>
+  decltype(auto) await_transform(Value&& value) {
+    // If the sender we're awaiting for is unifex::schedule
+    if constexpr (unifex::is_sender_for_v<
+                      unifex::remove_cvref_t<Value>,
+                      unifex::schedule>) {
+      return transform_schedule_sender_(static_cast<Value&&>(value));
+    }
+    // If we already have a sender => just await it
+    else if constexpr (unifex::sender<Value>) {
+      return unifex::await_transform(
+          *this,
+          unifex::with_scheduler_affinity(
+              static_cast<Value&&>(value), this->sched_));
+    }
+    // Either await_transform has been customized or Value is an awaitable.
+    // Either way, we can dispatch to the await_transform CPO, then insert a
+    // transition back to the correct execution context if necessary.
+    else if constexpr (
+        unifex::tag_invocable<
+            unifex::tag_t<unifex::await_transform>,
+            decltype(*this),
+            Value> ||
+        unifex::detail::_awaitable<Value>) {
+      return unifex::with_scheduler_affinity(
+          *this,
+          unifex::await_transform(*this, static_cast<Value&&>(value)),
+          this->sched_);
+    } else {
+      // Otherwise, we don't know how to await this type. Just return it and
+      // let the compiler issue a diagnostic.
+      return (Value &&) value;
+    }
+  }
+
+  void transform_schedule_sender_impl_(unifex::any_scheduler newSched) {
+    // this->consumerSched_ points to the correct scheduler
+    this->rescheduledBefore_ = true;
+    this->sched_ = std::move(newSched);
+  }
+
+  template <typename ScheduleSender>
+  decltype(auto) transform_schedule_sender_(ScheduleSender&& snd) {
+    // This sender is a scheduler provider. Get the scheduler. This
+    // get_scheduler call returns a reference to the scheduler stored within
+    // snd, which is an object whose lifetime spans a suspend point. So it's
+    // ok to build an any_scheduler_ref from it:
+    transform_schedule_sender_impl_(unifex::get_scheduler(snd));
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: include/unifex/async_generator.hpp [hunk 1]
```diff
@@ -0,0 +1,546 @@
...
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// A lot of stuff from cppcoro
```
georgikoyrushki95, line 17: A lot of this stuff is copied from `cppcoro`. Not too sure how to reflect this here.

include/unifex/async_generator.hpp
```diff
@@ -0,0 +1,546 @@
...
+  async_generator& operator=(async_generator&& other) noexcept {
+    async_generator temp(std::move(other));
+    swap(temp);
+    return *this;
+  }
+
+  async_generator(const async_generator&) = delete;
+  async_generator& operator=(const async_generator&) = delete;
+
+  auto next() noexcept {
```
georgikoyrushki95, line 467: `next()` and `cleanup()` is as shared by @ispeters in https://godbolt.org/z/Kh3rWTMao

include/unifex/async_generator.hpp
ispeters: @lewissbaker, do you have any insight into how we should handle your copyrights here?

ispeters:
Sorry for the delay on this review; I spent some time looking at it with @janondrusek yesterday and we came up with a few high-level thoughts that I think you'll need to try to address before we can give useful detailed feedback.

First, we agree with your "general feeling" re: implementing the generator in terms of an iterator-based interface. If we had two missing pieces:
 - the `read()` algorithm from P2300, which returns a _Sender_ that produces the result of invoking a given CPO on the _Receiver_; and
 - an `any_receiver_ref<T&>` in the generator's `promise_type`;

then you could implement `next()` thusly:
```
  auto next() noexcept {
    // grab the receiver's scheduler; assume that the next-sender
    // is always started on the corresponding context
    return read(get_scheduler)
      | let_value([this](auto sched) {
        if (!scheduler_) {
          // capture the receiver's scheduler as the stream's
          // scheduler on the first run of the next-sender
          scheduler_ = sched;
        }

        // check to see if we're currently running on the saved scheduler
        return just_void_or_done(scheduler_ == get_scheduler(receiver))
            // get back on the desired scheduler when we're not already there
            | let_done([this]() { return schedule(scheduler_); })
            | let_value([this]() {
              // once we're on the right scheduler, use create<>()
              // to resume the generator's coroutine_handle<> after
              // saving the create-receiver in the promise so we can
              // complete create-sender from within the generator
              return create<T&>([this](auto& rec) {
                // receiver_ is an any_receiver_ref<T>
                coroutine_.promise().receiver_ = rec;
                coroutine_.resume();
              });
            });  
      });
  }
```
With the above implementation, you can invoke `promise().receiver_.set_value(v)` (or `….set_error(e)`) from a `co_yield` expression, and `promise().receiver_.set_done()` from the `final_awaiter`, which would map the generator's state transitions to the expected results of the next-sender pretty cleanly.

In the absence of `read()`, you'd need to write your own _Sender_ type so you can hook `connect` to get access to the _Receiver's_ _Scheduler_; if you want to implement `read()` first, I'd be happy to merge that PR.

It's not obvious to me whether `create<>()` is the best way to implement that part of the next-sender. I like that it's reusing existing code rather than adding yet another hand-rolled _Sender_, but I wonder if there'd be efficiencies in hand-rolling that part.

Second, I think the current `at_coroutine_exit` integration is probably doing something surprising; I suspect it runs the registered callbacks on the next `co_yield` or `co_return`, whichever comes first, but I'd want registered coroutines to only run when the generator is destroyed/completed.

`at_coroutine_exit` works by inserting the registered at-exit coroutine into the async stack so that, when the current coroutine goes looking for a continuation to resume it finds the at-exit coroutine instead of the actual caller.

Suppose you have this:
```
task<void> bar() {
  co_return;
}

task<void> foo() {
  co_await bar();
}
```
If you suspend the process just before `co_return` is about to end the `bar()` coroutine, you'd see that the promise associated with the `bar()` coroutine has a `continuation_` that points to the `foo()` coroutine. So you have this stack:
```
bar()
foo()
```
When `bar()` returns, it looks at the promise, finds the `foo()` handle, and resumes it.

If you change `bar()` like so:
```
task<void> bar() {
  co_await at_coroutine_exit([]() -> task<void> {
    co_return;
  });

  // breakpoint here
}
```
and then stop the debugger on the `breakpoint here` comment, the "stack" will look like this:
```
bar()
<lambda>
foo()
```
so, when `bar()` returns, it'll activate the registered at-exit coroutine, which will then activate `foo()` upon returning.

With your current implementation, the generator's continuation is whoever is waiting for the result of `next()` so inserting at-exit coroutines in the default way means running them before yielding the next value.

Instead, I think the generator should have an initially-empty `coroutine_handle<>` that represents the top of a stack of clean-up actions. Awaiting `at_coroutine_exit` from within the generator should push the given at-exit coroutine onto this stack. The cleanup-sender should then run whatever clean-up actions in finds in the stack when the _Stream_ is cleaned up.

Third, I regret the way we implemented the magic `co_await schedule(s)` semantics that reschedules the current coroutine on the new _Scheduler_ and updates the coroutine's concept of which _Scheduler_ it is affine to. We've left the current implementation in place because it's being used internally (and, presumably, externally), but we'd like to migrate to something that's more explicit and less magic. You might notice that we explicitly don't support `co_await schedule(s)` from a `nothrow_task<>`. I would prefer not to spread the regrettable implementation choice to this new coroutine type so, instead of copying the behaviour from `task<>`, could you please copy the `static_assert` behaviour from `nothrow_task<>`? It'll mean you'll need to express some of your scheduler-affinity checks differently, but I think it's the better choice in the long run.

Finally, we should be able to avoid the use of an `async_scope` within the generator. I _think_ that the above-suggested changes to how to implement `next()` means that the reschedule operation state will be optionally embedded in the next-sender's operation state, which moots this whole line of thought but, if not, we should be able to directly `connect` and `start` an operation state in some inline storage somewhere rather than relying on `spawn_detached` to heap-allocate one for you.

Regarding the questions in your checklist:

> - [x]  Support co_await and co_yield inside the generator coroutine.
>    ❗ This achieved already by cppcoro::async_generator

Agreed; looks good.

> - [ ] Feel like a unifex::task<>
>    - [x] Scheduler affinity
>        ❗ In short:
>            (a) resuming the generator resumes it in the scheduler it was running before suspension;
>            (b) resuming the consumer coroutine resumes it on the scheduler it was running before co_await-ing the generator.
>    - [x] Support awaiting senders
>        ❗ Achieved via await_transform customization in the generator promise, similar to unifex::task<>
>    - [ ] Support unwind-on-cancellation semantics / stop tokens
>        ❗ I did not look into that. Looks like it might be too much for 1 PR.
>    - [ ] noexcept version of it
>        ❗ Shouldn't be too much to add to this PR. Can do, if there is interest

Mostly agreed.

A `nothrow_generator` is probably useful for binary size-sensitive consumers (terminating on exception tends to save a lot of overhead), but it could wait for another PR.

I think unwind-on-cancellation might need to be implemented in the first draft. Without unwind-on-cancellation, I presume that awaiting a _Sender_ that completes with `set_done()` would either terminate or continue as if it succeeded and I think both outcomes are bad. Terminating on cancellation would avoid continuing execution in an undefined state, but would be a very disappointing surprise; continuing past `co_await foo()` when `foo()`'s sender cancels would mean that the post-conditions of `foo()` are not met yet we're continuing anyway—an even more disappointing surprise.

> - [ ] Ability to let callers opt-out of scheduler affinity. This is achieved in unifex::task<> by having this struct _sa_task.
>    ❗ Have not looked into that

`_sa_task` isn't really for opting out of scheduler affinity; an `_sa_task<>` is a `task<>` that assumes its caller enforces the scheduler-affinity invariants and so doesn't do any enforcement itself. I'm not sure whether the scheduler-affinity invariants need updating for generators because the flow of control is quite different from that of normal tasks. One thing that `task<>` does that we haven't yet contemplated in this generator design is thunk stop requests from the receiver onto the "right" scheduler so that any descendant _Senders_ that complete synchronously from a stop request do so on the right context.

georgikoyrushki95:
Putting it as draft as it will take some time :)

georgikoyrushki95:
@ispeters started looking into this again. I am having a little trouble instantiating an `any_receiver_ref<T&>` within `create<T&>(...)`. [This](https://github.com/facebookexperimental/libunifex/blob/main/include/unifex/any_sender_of.hpp#L86) seems to be the constructor for it. Would appreciate some help, a few of the things I tried:

```C++
return create<T&>([this](auto& rec) {
    any_receiver_ref<T&> r1{inplace_stop_token{}, &rec};
    any_receiver_ref<T&> r2{get_stop_token(rec), &rec};
    // below obviously doesn't work because the constructor expects 2 args
    any_receiver_ref<T&> r2{&rec};
}
```

georgikoyrushki95:
> Finally, we should be able to avoid the use of an async_scope within the generator. I think that the above-suggested changes to how to implement next() means that the reschedule operation state will be optionally embedded in the next-sender's operation state, which moots this whole line of thought but, if not, we should be able to directly connect and start an operation state in some inline storage somewhere rather than relying on spawn_detached to heap-allocate one for you.

This is true when we process a `next()` sender within the stream. My existing implementation also has the logic to return back to the waiting coroutine's scheduler [within the yield's await_suspend](https://github.com/facebookexperimental/libunifex/blob/46220cc0b04a193de6729cc1ab83e87f7871844c/include/unifex/async_generator.hpp#L146). 

To draw some comparison's with `unifex::task` - this isn't needed there since you only return to the caller once when the coroutine exits. So all you need there is the `at_coroutine_exit` bit. However, here we return the the caller on every yield (as well as during the coroutine exit). So do we need to take care of this within the yield operation?

georgikoyrushki95:
I did play around with the API of `any_receiver_ref` a little and I got the following compiler error:

```C++
/libunifex/include/unifex/async_generator.hpp:213:16: error: ‘struct unifex::_any::_rec_ref<unifex::type_list<>, int&>::type’ has no member named ‘set_value’
  213 |     receiver_->set_value(value);
      |     ~~~~~~~~~~~^~~~~~~~~
```

This is meant to be a type-erased handle on top of the underlying receiver, right? Are we sure this is properly implemented? I have only seen one usage of this within `any_scheduler` and nothing appears to be calling `set_value()`

ispeters:
> @ispeters started looking into this again. I am having a little trouble instantiating an `any_receiver_ref<T&>` within `create<T&>(...)`. [This](https://github.com/facebookexperimental/libunifex/blob/main/include/unifex/any_sender_of.hpp#L86) seems to be the constructor for it. Would appreciate some help, a few of the things I tried:
> 
> ```c++
> return create<T&>([this](auto& rec) {
>     any_receiver_ref<T&> r1{inplace_stop_token{}, &rec};
>     any_receiver_ref<T&> r2{get_stop_token(rec), &rec};
>     // below obviously doesn't work because the constructor expects 2 args
>     any_receiver_ref<T&> r2{&rec};
> }
> ```

What kind of errors are you getting? I managed to wrap the `create<>()` "receiver" in an `any_receiver_ref<int&>` [here](https://godbolt.org/z/vn7GGeK8r) and I can't tell what I'm doing differently from what you tried.

[This](https://godbolt.org/z/E9osvzvP3) would be better if you plan to actually use the caller's stop token, though.

georgikoyrushki95:
> > @ispeters started looking into this again. I am having a little trouble instantiating an `any_receiver_ref<T&>` within `create<T&>(...)`. [This](https://github.com/facebookexperimental/libunifex/blob/main/include/unifex/any_sender_of.hpp#L86) seems to be the constructor for it. Would appreciate some help, a few of the things I tried:
> > ```c++
> > return create<T&>([this](auto& rec) {
> >     any_receiver_ref<T&> r1{inplace_stop_token{}, &rec};
> >     any_receiver_ref<T&> r2{get_stop_token(rec), &rec};
> >     // below obviously doesn't work because the constructor expects 2 args
> >     any_receiver_ref<T&> r2{&rec};
> > }
> > ```
> 
> What kind of errors are you getting? I managed to wrap the `create<>()` "receiver" in an `any_receiver_ref<int&>` [here](https://godbolt.org/z/vn7GGeK8r) and I can't tell what I'm doing differently from what you tried.
> 
> [This](https://godbolt.org/z/E9osvzvP3) would be better if you plan to actually use the caller's stop token, though.

I got this to work, thanks for coming back. A thing which I am facing now is within the handler of a `yield` expression, how do we get back to the receiver's scheduler, without the usage of an `async_scope`. The sketch I got right now (w/o scheduler affinity) is something like this:

```C++
// within async_generator_promise<T> class
coro::suspend_always yield_value(value_type& value) noexcept {
    // How do we jump back on the correct schdeuler?
    set_value(std::move(*receiverOpt_), value);    
    return {};
}
```

My intuition is telling me that we need to return more than a mere `coro::suspend_always` - some sort of sender, which would allow us to jump back on the receiver's scheduler before calling to `set_value`. I thought that the below sketch could work, but alas:

```C++
// task should technically work since co_yield i; 
// translates to co_await promise.yield_value(i);
// and a task is awaitable
task<void> yield_value(value_type& value) noexcept {
   if (/* generator scheduler != receiverOpt_'s scheduler */)
       co_await schedule(get_scheduler(*receiverOpt));
    set_value(std::move(*receiverOpt_), value);
}
```

I do get the error:

```
libunifex/test/async_generator_test.cpp: In lambda function:
libunifex/test/async_generator_test.cpp:28:5: error: no member named ‘await_ready’ in ‘unifex::task<void>’ {aka ‘unifex::_task::_task<void, false>::type’}
```

georgikoyrushki95:
> I would prefer not to spread the regrettable implementation choice to this new coroutine type so, instead of copying the behaviour from task<>, could you please copy the static_assert behaviour from nothrow_task<>? It'll mean you'll need to express some of your scheduler-affinity checks differently, but I think it's the better choice in the long run.

@ispeters this was your comment about using `co_await schedule(xxx)` within the generator. It reads a little as if you don't want to support this for now. Is that the case or you want to express it differently from how it is in `task<>`?

georgikoyrushki95:
Just pushed [this commit](https://github.com/facebookexperimental/libunifex/pull/596/commits/39a0dc211e494b180c3d5e52ca7d48f53f87800c), addressing some of the comments. Gonna write up a quick progress update below:

- Implemented `next()` as suggested by @ispeters in [this comment](https://github.com/facebookexperimental/libunifex/pull/596#issuecomment-1909188298).
  - Instead of adding the generic [read()](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2300r7.html#design-sender-factory-read) from P2300, I just added a `read_scheduler()` algorithm, private to the generator. The idea was to progress with the rest of the stuff for now - the generic `read()` we can add towards the end.
- Removed usages of the `async_scope` and `at_coroutine_exit` that I used for the scheduler affinity invariants. Those, it seems, can fully be implemented via the sender / receiver protocol:
   - On resumption of the generator from the consumer (i.e. `co_await gen.next()`), we reschedule on the correct executor (if necessary).
   - On yielding of the generator - we have a custom receiver for `unifex::schedule` (`reschedule_receiver`), that "awakes" the consumer once back on the correct scheduler. This is also used in `final_suspend`.
- Unwind on cancellation as per @ispeters comment (**_I think unwind-on-cancellation might need to be implemented in the first draft. Without unwind-on-cancellation, I presume that awaiting a Sender that completes with set_done()..._**) is implemented and there is a test for this (`await_just_done_in_async_generator`).

**Things that still need to be done:**
- Drop support for `co_await schedule(...)` within the generator, as described in [this PR comment](https://github.com/facebookexperimental/libunifex/pull/596#issuecomment-1909188298) and the [issue](https://github.com/facebookexperimental/libunifex/issues/586#issuecomment-1875786980) as well. 

  I currently support `co_await schedule(...)` just like in `task<>`
- The equivalent of `_sa_task` for the generator - basically that assumes its caller enforces the scheduler-affinity invariants and so doesn't do any enforcement itself
- This (I haven't delved into it too much): _**One thing that task<> does that we haven't yet contemplated in this generator design is thunk stop requests from the receiver onto the "right" scheduler so that any descendant Senders that complete synchronously from a stop request do so on the right context.**_
- Generally cancellations as described in [P2300](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2300r7.html#design-cancellation).

include/unifex/async_generator.hpp
```diff
@@ -0,0 +1,374 @@
...
+                 // there
+                 | let_done([this]() {
+                      return schedule(*coroutine_.promise().sched_);
+                    }) |
+                 let_value([this]() {
+                      // once we're on the right scheduler, use create<>()
+                      // to resume the generator's coroutine_handle<> after
+                      // saving the create-receiver in the promise so we can
+                      // complete create-sender from within the generator
+                      return create<T>([this](auto& rec) {
```
ispeters, line 349: I think you should use `create<T&>` here. The values generated by this generator will all be passed to the next-receiver using code like this:
```
co_yield <expr>;
```
so the lifetime of `<expr>` will span the suspend point created by the `co_yield`. You can therefore efficiently and safely pass `<expr>` as an lvalue reference through to the next-receiver.
```suggestion
                      return create<T&>([this](auto& rec) {
```

include/unifex/async_generator.hpp
```diff
@@ -0,0 +1,374 @@
...
+    swap(temp);
+    return *this;
+  }
+
+  async_generator(const async_generator&) = delete;
+  async_generator& operator=(const async_generator&) = delete;
+
+  // Potential problem: not sure if get_scheduler(gen.next()) would return
+  // the right thing here. Perhaps we need a wrapper sender that also records
+  // each next sender's scheduler and customizes get_scheduler(...)
```
ispeters, line 325: I don't think this is a problem. P2300 has the concept of a "completion scheduler" that senders may report, but Unifex doesn't have that concept. You can't pass a sender to `unifex::get_scheduler()`, only a receiver.

include/unifex/async_generator.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+  ~async_generator() {
+    if (coroutine_) {
+      coroutine_.destroy();
+    }
+  }
+
+  async_generator& operator=(async_generator&& other) noexcept {
+    async_generator temp(std::move(other));
+    swap(temp);
```
ispeters, line 316: Eric Niebler taught me that you can do this in less code by receiving the argument by value.
```suggestion
  async_generator& operator=(async_generator other) noexcept {
    swap(other);
```

include/unifex/async_generator.hpp
```diff
@@ -0,0 +1,374 @@
...
+    } else {
+      unifex::set_done(std::move(*genCoro.promise().receiverOpt_));
+    }
+  }
+
+  void await_resume() noexcept {}
+
+private:
+  std::optional<value_type> value_;
+};
```
ispeters, line 151: Below, I argue that you can rely on the generator generating values with code like
```
co_yield <expr>;
```
which means that `<expr>` survives across a suspend point. I think this means you can just store the address of that value here.
```suggestion
template <typename T>
class async_generator_yield_operation {
public:
  using value_type = std::remove_reference_t<T>;

  async_generator_yield_operation() noexcept
    : value_{nullptr} {}

  explicit async_generator_yield_operation(value_type& value) noexcept
    : value_{&value} {}

  bool await_ready() const noexcept { return false; }

  template <typename Promise>
  void await_suspend([[maybe_unused]] unifex::coro::coroutine_handle<Promise>
                         genCoro) noexcept {
    const auto& consumerSched = genCoro.promise().consumerSched_;
    if (unifex::get_scheduler(genCoro.promise()) != consumerSched) {
      genCoro.promise().rescheduleOpSt_ = unifex::connect(
          unifex::schedule(consumerSched),
          reschedule_receiver<Promise>{value_, genCoro});
      unifex::start(*genCoro.promise().rescheduleOpSt_);
      return;
    }

    if (value_) {
      unifex::set_value(std::move(*genCoro.promise().receiverOpt_), *value_);
    } else {
      unifex::set_done(std::move(*genCoro.promise().receiverOpt_));
    }
  }

  void await_resume() noexcept {}

private:
  value_type* value_;
};
```

include/unifex/async_generator.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+struct _read_scheduler_t {
+  constexpr _read_scheduler_sender operator()() const noexcept { return {}; }
+};
+
+inline constexpr _read_scheduler_t read_scheduler{};
+
+template <typename Promise>
+struct reschedule_receiver {
+  std::optional<typename Promise::value_type> value_;
```
ispeters, line 90: See below for why I think this is safe.
```suggestion
  typename Promise::value_type* value_;
```

include/unifex/async_generator.hpp
```diff
@@ -0,0 +1,374 @@
...
+    return *p.sched_;
+  }
+
+  // This is needed for at_coroutine_exit to do the async clean up
+  friend unifex::continuation_handle<> tag_invoke(
+      const unifex::tag_t<unifex::exchange_continuation>&,
+      async_generator_promise& p,
+      unifex::continuation_handle<> action) noexcept {
+    return std::exchange(p.continuation_, std::move(action));
+  }
```
ispeters, line 195: I think this is the wrong implementation for `at_coroutine_exit`. Instead, the generator object should have an initially-null `coroutine_handle<>` that points to the top of a stack of continuations to be awaited in `cleanup()`. Invoking `at_coroutine_exit` inside a generator should push the new coroutine onto the stack. Then, `cleanup()` should run the stack of continuations (if it's non-empty) before completing with done.

Alternatively, perhaps the handle shouldn't be initially-null, but the result of invoking this coroutine:
```
nothrow_task<void> cleanup() {
  co_await just_done();
}
```
then you always have a non-empty stack.

ispeters (commented):
I *really* like where this is going! Thanks for persisting and sorry the review turn-around is so long.

include/unifex/async_generator.hpp
georgikoyrushki95: @ispeters this is actually a leftover from when I was using `at_coroutine_exit`. There were a few things, for which I needed it:

1) awaiting `join()` on the async scope
2) rescheduling back onto the consumer's scheduler (basically the equivalent of what is done in `unifex::task<>` on coroutine exit.

Right now 1) is no longer needed.

2) I believe is handled within the `async_generator_yield_operation<>`. This one has the scheduler affinity logic within its `await_suspend` - if we aren't on the correct scheduler, we kick off a `unifex::reschedule` to go back to it.

Finally, we return this awaiter both in `yield_value()` and `final_suspend`. So this should automatically handle the rescheduling during the generator's coroutine exit. 

Let me know what you think, but it seems we don't need `at_coroutine_exit` at all here.

georgikoyrushki95: I see what you mean and also your very original comment explains the same thing (some stuff starts making sense after a while with this code base... 😄).

ispeters: I'm not 100% certain this is going to work, but here's how I _think_ it'll work.

[`at_coroutine_exit`](https://github.com/facebookexperimental/libunifex/blob/main/include/unifex/at_coroutine_exit.hpp#L317-L320) returns a [`_cleanup_task<>`](https://github.com/facebookexperimental/libunifex/blob/main/include/unifex/at_coroutine_exit.hpp#L266). `_cleanup_task<>`'s implementation of [`await_suspend`](https://github.com/facebookexperimental/libunifex/blob/main/include/unifex/at_coroutine_exit.hpp#L279-L290) [invokes `exchange_continuation`](https://fburl.com/jwpk5nds) on the calling coroutine's promise. So, I _think_ you can implement `exchange_continuation` in such a way that each call pushes another clean-up task onto the stack of things to resume from `cleanup()`.

PR #598, opened 2024-01-23
Introduce `stop_on_request` algorithm

tmoi29:
`stop_on_request` is a sender algorithm that allows external stop sources to cancel the composed sender. It takes in `N` stop tokens and if any one of them or the receiver receives a stop signal, the resulting sender will be cancelled.

### DIFF
--- a/include/unifex/stop_on_request.hpp
+++ b/include/unifex/stop_on_request.hpp
@@ -154,6 +154,8 @@ struct _op<Receiver, StopTokens...>::type {
       destructOnError.release();
     }
     UNIFEX_CATCH(...) {
+      // all the stop callbacks have been destroyed by the time we get here
+      // so it's not possible for a callback to race with this load
       if (callbackState_.load(std::memory_order_acquire) ==
           _callback_state::AT_LEAST_ONE_CALLED) {
         // we received a stop request before we experienced an error
@@ -168,12 +170,14 @@ struct _op<Receiver, StopTokens...>::type {
     auto expected = _callback_state::INIT;
     // Update state to mark that all callbacks have been constructed only if the
     // previous state was INIT. If the previous state was AT_LEAST_ONE_CALLED,
-    // don't change the state
+    // don't change the state, but invoke complete() on behalf of the callback
+    // that was invoked
     if (!callbackState_.compare_exchange_strong(
             expected,
             _callback_state::ALL_CONSTRUCTED_NOT_CALLED,
             std::memory_order_acq_rel)) {
       // complete if we transitioned from AT_LEAST_ONE_CALLED
+      UNIFEX_ASSERT(expected == _callback_state::AT_LEAST_ONE_CALLED);
       complete();
     }
     // otherwise let the other callbacks handle the completion
--- a/include/unifex/stop_on_request.hpp
+++ b/include/unifex/stop_on_request.hpp
@@ -165,17 +165,18 @@ struct _op<Receiver, StopTokens...>::type {
       return;
     }
 
-    // update state to mark that all callbacks have been constructed
-    auto oldState = callbackState_.exchange(
-        _callback_state::ALL_CONSTRUCTED_NOT_CALLED, std::memory_order_acq_rel);
-    if (oldState == _callback_state::AT_LEAST_ONE_CALLED) {
-      // if a stop callback was already called, then we can handle completion
+    auto expected = _callback_state::INIT;
+    // Update state to mark that all callbacks have been constructed only if the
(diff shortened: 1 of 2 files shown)


test/stop_on_request_test.cpp
```diff
@@ -0,0 +1,401 @@
...
+                                  stopSources[i * numSources + 4].get_token()) |
+        unifex::let_done([&]() {
+                                wasCancelled = true;
+                                return unifex::just();
+                              });
+    unifex::sync_wait(unifex::when_all(
+        unifex::when_all_range(std::move(tasks)), cancellationSender));
+    EXPECT_TRUE(wasCancelled);
+    wasCancelled = false;
+  }
```
ispeters, line 72: I think it's unfortunate that this test is being hidden from some CI jobs just because it depends on coroutines. Maybe we could rewrite it to work everywhere?
```suggestion
  constexpr size_t iterations = 10;
  constexpr size_t numSources = 5;

  for (size_t i = 0; i < iterations; i++) {
    std::array<unifex::single_thread_context, numSources> threads{};
    std::array<unifex::inplace_stop_source, numSources> stopSources{};

    auto makeTask = [&](size_t index) noexcept {
      auto& thread = threads[index];
      auto* stopSource = &stopSources[index];

      return unifex::on(thread, unifex::just_from([stopSource]() noexcept {
        stopSource->request_stop();
      });
    };

    std::vector<decltype(makeTask(0))> tasks;

    for (size_t j = 0; j < numSources; j++) {
      tasks.push_back(makeTask(j));
    }

    bool wasCancelled = false;
    auto cancellationSender = std::apply([&](auto&... stopSources) noexcept {
      return unifex::stop_on_request(stopSources.get_token()...)
          | unifex::let_done([&]() {
            wasCancelled = true;
            return unifex::just();
          });
    });

    unifex::sync_wait(unifex::when_all(
        unifex::when_all_range(std::move(tasks)), std::move(cancellationSender)));

    EXPECT_TRUE(wasCancelled);
  }
```

ispeters (changes requested):
Can't say I really understand the Windows-only build errors you're getting, but they reference `task<>` and `any_sender_of<>`, which go away if you take the suggestion I've made below. Maybe the build errors will resolve, too?

include/unifex/stop_on_request.hpp
```diff
@@ -0,0 +1,252 @@
...
+    auto expected = _callback_state::INIT;
+    // Update state to mark that all callbacks have been constructed only if the
+    // previous state was INIT. If the previous state was AT_LEAST_ONE_CALLED,
+    // don't change the state
+    if (!callbackState_.compare_exchange_strong(
+            expected,
+            _callback_state::ALL_CONSTRUCTED_NOT_CALLED,
+            std::memory_order_acq_rel)) {
+      // complete if we transitioned from AT_LEAST_ONE_CALLED
+      complete();
```
ispeters, line 177: Let's add an assert here.
```suggestion
      UNIFEX_ASSERT(expected == _callback_state::AT_LEAST_ONE_CALLED);
      complete();
```

include/unifex/stop_on_request.hpp
```diff
@@ -0,0 +1,252 @@
...
+        unifex::set_error(std::move(receiver_), std::current_exception());
+      }
+
+      return;
+    }
+
+    auto expected = _callback_state::INIT;
+    // Update state to mark that all callbacks have been constructed only if the
+    // previous state was INIT. If the previous state was AT_LEAST_ONE_CALLED,
+    // don't change the state
```
ispeters, line 171:
```suggestion
    // don't change the state, but invoke complete() on behalf of the callback that was invoked
```

include/unifex/stop_on_request.hpp
```diff
@@ -0,0 +1,252 @@
...
+
+      // construct a callback for each stop token
+      // Note: if any of these tokens are already stopped, then its
+      // corresponding callback will be immediately invoked
+      constructCallbacks();
+
+      destructOnError.release();
+    }
+    UNIFEX_CATCH(...) {
+      if (callbackState_.load(std::memory_order_acquire) ==
```
ispeters, line 157:
```suggestion
      // all the stop callbacks have been destroyed by the time we get here
      // so it's not possible for a callback to race with this load
      if (callbackState_.load(std::memory_order_acquire) ==
```

ispeters (changes requested):
Looks good! Let's just add an assert and some more documentation that explains our hard-won understanding of the race condition we almost committed and then we can merge.

ispeters (approved):

PR #606, opened 2024-04-15
Remove the cmake configuration machinery for config.hpp

### LINKED ISSUE #510: Get rid of `config.hpp.in`
Static `config.hpp` that just handles all the configurable differences in the preprocessor.

Already done for detecting coroutine support, which used to use cmake.

denizevrenci:
We can rely on the preprocessor for it instead since we only support C++17 and above.
Closes #510

### DIFF
--- a/include/unifex/config.hpp
+++ b/include/unifex/config.hpp
@@ -24,13 +24,13 @@
 
 #include <cassert>
 
-// the configured options and settings for unifex
-#define UNIFEX_VERSION_MAJOR @libunifex_VERSION_MAJOR@
-#define UNIFEX_VERSION_MINOR @libunifex_VERSION_MINOR@
-
-#cmakedefine01 UNIFEX_NO_MEMORY_RESOURCE
-#cmakedefine UNIFEX_MEMORY_RESOURCE_HEADER <@UNIFEX_MEMORY_RESOURCE_HEADER@>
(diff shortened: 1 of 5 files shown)


ispeters:
Looks like this addresses #510.

include/unifex/config.hpp
```diff
@@ -60,8 +60,7 @@
 #  if __has_include(<coroutine>) && defined(__cpp_lib_coroutine)
 #    define UNIFEX_COROUTINES_HEADER <coroutine>
 #    define UNIFEX_COROUTINES_NAMESPACE std
-#  elif __has_include(<experimental/coroutine>) && \
-        !(defined(_MSC_VER) && defined(__clang__)) // clang-cl.exe
+#  elif __has_include(<experimental/coroutine>)
```
ispeters, line 63: Why this change? I don't remember the history here, but we've had a whole bunch of problems with MSVC-specific stuff so I'm hesitant to enable coroutine support where it's presently disabled.
```suggestion
#  elif __has_include(<experimental/coroutine>) && \
        !(defined(_MSC_VER) && defined(__clang__)) // clang-cl.exe
```

ispeters (changes requested):
Looks awesome, but I think we need to bring back the experimental coroutine exclusion.

ispeters (approved):

PR #607, opened 2024-04-24
Make scope_guard nodiscard

denizevrenci:
We don't want to mistakenly trigger the destructor before the end of the scope.


ispeters (approved):
Thanks!

PR #619, opened 2024-07-11
Make any_sender_of<> play nicer with MSVC

ispeters:
`unifex::any_sender_of<>` needs a type-erased operation state, which it creates with `unifex::any_unique<concrete-op>`. The tricky bit is that operation states are expected to be immovable so constructing that `any_unique` requires care. The old approach was to use a type called `_rvo<>` that implicitly converted to the desired type, but it was brittle. This change invents a new opstate type with two jobs:
 1. wrap some other opstate, and
 2. construct that other opstate by invoking a callable.


ispeters:
I think this is breaking an internal build.

janondrusek (approved):

PR #623, opened 2024-08-05
Fix Wformat compiler warnings

ccotter:
```
examples/linux/http_server_io_uring_test.cpp: In function 'unifex::task<{anonymous}::Request> {anonymous}::parse_request(unifex::linuxos::io_uring_context::async_read_write_file&)':
examples/linux/http_server_io_uring_test.cpp:99:34: error: format '%ld' expects argument of type 'long int', but argument 2 has type 'std::basic_string<char>::size_type' {aka 'unsigned int'} [-Werror=format=]
   99 |       std::printf("req too big=%ld\n", req.size());
      |                                ~~^     ~~~~~~~~~~
      |                                  |             |
      |                                  long int      std::basic_string<char>::size_type {aka unsigned int}
      |                                %d
examples/linux/http_server_io_uring_test.cpp: In lambda function:
examples/linux/http_server_io_uring_test.cpp:160:54: error: format '%ld' expects argument of type 'long int', but argument 2 has type 'std::chrono::duration<long long int, std::ratio<1, 1000> >::rep' {aka 'long long int'} [-Werror=format=]
  160 |             then([ms] { std::printf("Timeout after %ldms\n", ms.count()); }),
      |                                                    ~~^       ~~~~~~~~~~
      |                                                      |               |
      |                                                      long int        std::chrono::duration<long long int, std::ratio<1, 1000> >::rep {aka long long int}
      |                                                    %lld
```

### DIFF
--- a/examples/linux/http_server_io_uring_test.cpp
+++ b/examples/linux/http_server_io_uring_test.cpp
@@ -96,7 +96,7 @@ parse_request(io_uring_context::async_read_write_file& readWriteFile) {
     }
     // protect from infite request
     if (req.size() > 8 * buffer.size()) {
-      std::printf("req too big=%ld\n", req.size());
+      std::printf("req too big=%zu\n", req.size());
       request.method = Method::OTHER;
       break;
     }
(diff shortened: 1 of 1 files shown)


ispeters (changes requested):
Thanks for the PR! I like the intent and want to merge it, but it seems there are some broken CI jobs that'll need fixing first.

PR #626, opened 2024-08-24
Give spawn_detached's receiver an async stack frame

ispeters:
This change adds an `AsyncStackFrame` (when support is enabled) to the capstone receiver connected to any sender passed to `unifex::spawn_detached()`. Forcing `unifex::v1::async_scope`'s `detached_spawn*` methods to be always-inline means that capturing the return address of `unifex:spawn_detached()` captures the call site of the v1 scope's methods.


janondrusek (approved):
Wanna add a test for the async stack enabled?

PR #627, opened 2024-08-24
Simplify await_transform's type-switching logic

ispeters:
This change simplifies the implementation-selection logic in `unifex::await_transform()`. Instead of a multi-level `if constexpr`, we now choose how to transform an awaited expression with mutually-exclusive `requires` clauses. The new order of preference is:

1. rely on the awaited type's customization of `await_transform`
2. return naturally-awaitable expressions directly
3. map senders to awaitables with the existing `_as_awaitable` logic

This preference ordering matches the old behaviour.

### DIFF
--- a/include/unifex/await_transform.hpp
+++ b/include/unifex/await_transform.hpp
@@ -200,29 +200,27 @@ struct _fn {
     return unifex::tag_invoke(_fn{}, promise, (Value&&)value);
   }
 
-  // Default implementation.
-  template(typename Promise, typename Value)            //
-      (requires(!tag_invocable<_fn, Promise&, Value>))  //
+  // Default implementation for naturally awaitable types
+  template(typename Promise, typename Value)  //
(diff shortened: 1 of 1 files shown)


include/unifex/await_transform.hpp
```diff
@@ -200,29 +200,27 @@ struct _fn {
...
-        return _as_awaitable<Promise, Value>{(Value&&)value, h};
-      } else {
-        static_assert(
-            unifex::sender_to<Value, _receiver_t<Promise, Value>>,
-            "This sender is not awaitable in this coroutine type.");
-        return (Value&&)value;
-      }
-    } else {
-      return (Value&&)value;
-    }
```
janondrusek, line 225: Is this branch now missing?

ispeters: It is. I'll restore it.

jesswong (approved):

PR #629, opened 2024-08-26
Customize get_return_address for as_sender

ispeters:
This change customizes `unifex::get_return_address()` for the `as_sender` adapter that makes awaitables connectable.

```
void foo(auto someAwaitable) {
  auto s = unifex::as_sender(std::move(someAwaitable));

  // now returns the call-site of as_sender, which is probably the call
  // to as_sender in foo (but might be the call to foo if foo is
  // inlined), rather than the result of
  // unifex::_get_return_address::default_return_address()
  auto rtnAdr = unifex::get_return_address(s);
}
```

### DIFF
--- a/include/unifex/connect_awaitable.hpp
+++ b/include/unifex/connect_awaitable.hpp
@@ -216,7 +216,9 @@ struct _sndr {
 
     static constexpr bool sends_done = true;
 
-    type(Awaitable awaitable) : awaitable_((Awaitable&&)awaitable) {}
+    explicit type(Awaitable awaitable, instruction_ptr returnAddress)
+      : awaitable_((Awaitable&&)awaitable)
+      , returnAddress_(returnAddress) {}
 
     template(typename Receiver)                   //
         (requires receiver_of<Receiver, Result>)  //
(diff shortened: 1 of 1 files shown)


include/unifex/connect_awaitable.hpp
```diff
@@ -216,7 +216,9 @@ struct _sndr {
 
     static constexpr bool sends_done = true;
 
-    type(Awaitable awaitable) : awaitable_((Awaitable&&)awaitable) {}
+    explicit type(Awaitable awaitable, instruction_ptr returnAddress)
+      : awaitable_((Awaitable&&)awaitable)
```
janondrusek, line 220: Do we want to clean up the C-style cast?

include/unifex/connect_awaitable.hpp
```diff
@@ -279,7 +294,8 @@ struct _fn {
       (requires detail::_awaitable<Awaitable>)  //
       _sender<remove_cvref_t<Awaitable>>
       operator()(Awaitable&& awaitable) const {
-    return _sender<remove_cvref_t<Awaitable>>{(Awaitable&&)awaitable};
+    return _sender<remove_cvref_t<Awaitable>>{
+        (Awaitable&&)awaitable, instruction_ptr::read_return_address()};
```
janondrusek, line 298: same ^^

janondrusek (approved):

PR #616, opened 2024-07-05
Add initial support for async stack traces to Unifex

ispeters:
This PR copies the core of Folly's [async stack trace support](https://github.com/facebook/folly/tree/main/folly/tracing) into `include/unifex/tracing` and builds on it to add support for generalized *Senders*.

When `UNIFEX_NO_ASYNC_STACKS` is falsey, `unifex::connect` returns a wrapped operation state that injects async stack tracing into the operation tree.
 - The wrapper operation:
    - stores an `AsyncStackFrame` for the wrapped operation; and
    - wraps the receiver.
 - In the wrapper operation's customization of `unifex::start` we:
    - create an `AsyncStackRoot` on the stack;
    - push the wrapper operation's `AsyncStackFrame` onto the current async stack;
    - activate the wrapper operation's `AsyncStackFrame` on the current `AsyncStackRoot`; and
    - start the wrapped operation.
 - In the wrapper receiver's completion methods we:
    - create an `AsyncStackRoot` on the stack;
    - copy the *parent* operation's `AsyncStackFrame` to the stack;
    - activate the parent `AsyncStackFrame` on the current `AsyncStackRoot`; and
    - invoke the parent operation's receiver.

The effect is that we build up a linked list (technically a DAG) of `AsyncStackFrame`s pointing "up" toward the start of the operation as `unifex::start` recurses into the nested operation state and then unwind it on the way back out as the receiver completion methods are invoked. At any given time, the current thread's `AsyncStackRoot` is sitting on the most recently-activated "normal" stack frame that is participating in async stack management, allowing Folly's `co_bt.py` debugger extension to figure out when it should stop walking normal stack frames and start walking async stack frames.

As alluded to above, the behaviour of the async stack tracing machinery is controlled by the `UNIFEX_NO_ASYNC_STACKS` preprocessor macro. If it's truthy, async stacks are not traced; if it's falsey, they are traced. The default in `unifex/config.hpp` is to enable async stack tracing in non-Windows debug builds.
 - Why not Windows builds?
    - Because there's something weird about how `any_sender_of<>` builds on Windows (both Clang and MSVC); the resolution is to land PR #619, but that PR breaks an internal Meta build so I'll have to come back to it.
 - Why only debug builds?
    - The additional work done to track async stacks adds non-trivial binary size to the output so I figure it should default to off for release builds. You can turn it on by defining `UNIFEX_NO_ASYNC_STACKS=0` in your release build script if the extra debuggability is worth the extra binary size in production.

[...]

PR #632, opened 2024-08-29
Add async stack support to coroutines

ispeters:
This change extends the work in #616 to support async stack frames in
`task<>` coroutines, including those that invoke `at_coroutine_exit()`.

In `task<>`, when `UNIFEX_NO_ASYNC_STACKS` is falsey, the awaiter returned from
`task<>`'s customization of `unifex::await_transform` stores an
`AsyncStackFrame`. The awaiter pushes its frame onto the current async stack in
`await_suspend()` and pops it again in `await_resume()`; since
`await_resume()` is only invoked for value and error completions, this
arrangement leaves it up to the waiting task to pop the awaiter's frame
when the awaited task completes with done. This can be expressed as a
new rule:

- when a coroutine completes with a value or an error, it is responsible
  for popping its own `AsyncStackFrame`; but
- when a coroutine completes with done, the *caller* is responsible for
  popping the callee's `AsyncStackFrame` as a part of the caller's
  `unhandled_done()` coroutine.

To support this new requirement of `unhandled_done()` (that it is
responsible for popping the callee's stack frame), this change
introduces `popAsyncStackFrameFromCaller`, which takes the caller's
stack frame by reference so that it can assert that, after popping the
current async frame (whatever it is), the new top frame is the caller's
frame.

A `task<>` promise has an `AsyncStackFrame*` that, when it's not
`nullptr`, points to the `AsyncStackFrame` in the awaiter waiting for
the task. This pointer exists even when `UNIFEX_NO_ASYNC_STACKS` is
truthy to help mitigate against ODR violations; linking together two TUs
with `UNIFEX_NO_ASYNC_STACKS` set differently is not explicitly
supported but, by ensuring this pointer always exists, some ODR problems
are avoided. When a `task<>` is awaited from a TU with async stack
support enabled, the awaited task's awaiter sets the promise's
`AsyncStackFrame*` to point to the awaiter's frame; when a `task<>` is
awaited from a TU with async stack support disabled, this assignment
never happens and the promise's pointer remains null.

The above description of `task<>`'s async stack maintenance only covers
the recursive case of on coroutine awaiting another. The base case is
handled in `connect_awaitable()`, where an `AsyncStackRoot` is set up
before starting the connected awaitable.

`stop_if_requested` used to model both `sender` and `awaitable` so that
`co_await stop_if_requested();` could take advantage of symmetric
transfer. The `stop_if_requested` sender now customizes
`await_transform` to express its participation in async stack
management. This means of expressing async stack awareness is
unsatisfying but I don't have any better ideas right now.

Lastly, `unifex::await_transform()` now wraps naturally-awaitable
arguments in an `awaiter_wrapper` that ensures the `coroutine_handle<>`
passed to the wrapped awaitable is one that establishes an active
`AsyncStackRoot` before resuming the real waiting coroutine.

### DIFF
--- a/include/unifex/connect_awaitable.hpp
+++ b/include/unifex/connect_awaitable.hpp
@@ -249,10 +249,19 @@ public:
       bool WithAsyncStackSupport = !UNIFEX_NO_ASYNC_STACKS>
   auto operator()(Awaitable&& awaitable, Receiver&& receiver) const
       -> _await::sender_task<remove_cvref_t<Receiver>, WithAsyncStackSupport> {
+    auto returnAddress = instruction_ptr::read_return_address();
+
+    if constexpr (is_tag_invocable_v<tag_t<get_return_address>, Awaitable>) {
+      // the awaitable has customized get_return_address so let's use its value
+      // rather than synthesizing one inside the connect machinery we're
+      // implementing
+      returnAddress = get_return_address(awaitable);
+    }
+
     return connect_impl<WithAsyncStackSupport>(
         std::forward<Awaitable>(awaitable),
         std::forward<Receiver>(receiver),
-        instruction_ptr::read_return_address());
+        returnAddress);
   }
 } connect_awaitable{};
 }  // namespace _await_cpo
--- a/include/unifex/connect_awaitable.hpp
+++ b/include/unifex/connect_awaitable.hpp
@@ -21,6 +21,8 @@
 #include <unifex/coroutine_concepts.hpp>
 #include <unifex/receiver_concepts.hpp>
 #include <unifex/sender_concepts.hpp>
+#include <unifex/tracing/async_stack.hpp>
+#include <unifex/tracing/get_async_stack_frame.hpp>
 #include <unifex/type_traits.hpp>
 #include <unifex/unhandled_done.hpp>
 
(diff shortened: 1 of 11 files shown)


include/unifex/connect_awaitable.hpp
```diff
@@ -99,12 +114,27 @@ class _sender_task<Receiver>::type {
         friend auto tag_invoke(CPO cpo, const promise_type& p) noexcept(
             std::is_nothrow_invocable_v<CPO, const Receiver&>)
             -> std::invoke_result_t<CPO, const Receiver&> {
-      return cpo(std::as_const(p.receiver_));
+      if constexpr (
+          WithAsyncStackSupport && same_as<CPO, tag_t<get_async_stack_frame>>) {
+        return &p.frame_;
+      } else {
+        return std::move(cpo)(std::as_const(p.receiver_));
+      }
```
jesswong, line 122: just a nit and not a big deal. this is a weird check. is it possible to separately define a get_async_stack_frame tag_invoke that will be called instead of this generic CPO function?

include/unifex/connect_awaitable.hpp
```diff
@@ -297,6 +365,7 @@ struct _fn {
       (requires detail::_awaitable<Awaitable>)  //
       _sender<remove_cvref_t<Awaitable>>
       operator()(Awaitable&& awaitable) const {
+    // TODO: this is going to generate an unfortunate return address
```
jesswong, line 368: why's that?

jesswong (approved):
LGTM?

include/unifex/with_scheduler_affinity.hpp
```diff
@@ -144,7 +157,7 @@ struct _fn final {
     if constexpr (
         !same_as<blocking_kind, blocking_t> &&
         (blocking_kind::always_inline == blocking_t{})) {
-      return Awaitable{(Awaitable &&) awaitable};
+      return Awaitable{(Awaitable&&)awaitable};
```
janondrusek, line 160: nit: wanna rid of the C-style cast?

include/unifex/tracing/async_stack-inl.hpp
```diff
@@ -67,6 +67,16 @@ popAsyncStackFrameCallee(unifex::AsyncStackFrame& calleeFrame) noexcept {
   calleeFrame.stackRoot = nullptr;
 }
 
+inline void popAsyncStackFrameFromCaller(
+    [[maybe_unused]] unifex::AsyncStackFrame& callerFrame) noexcept {
+  auto root = tryGetCurrentAsyncStackRoot();
+  assert(root != nullptr);
```
janondrusek, line 73: why call it `try` if it cannot be `nullptr`?

include/unifex/connect_awaitable.hpp
ispeters: It is weird, but I ran into trouble (that I don't remember in detail now) expressing the conditions correctly on a separate `tag_invoke` overload.

include/unifex/connect_awaitable.hpp
ispeters: Because it's inside the bowels of Unifex here; I think a better return address might be `get_return_address(awaitable)`. We might have actually changed this in one of the branches that improves return address reporting for `task<>`.

include/unifex/tracing/async_stack-inl.hpp
ispeters: The current async stack root is allowed to be `nullptr` in general, but the stack is in a bad state if it's null in this case. The alternative way to read the current stack root is `getCurrentAsyncStackRoot()`, but that returns a reference; if the stack is corrupt at this point, we can't meaningfully assert that a reference is valid, but we _can_ assert that a nullable pointer isn't null.

include/unifex/with_scheduler_affinity.hpp
ispeters: I do want to, but the clock has run out.

PR #640, opened 2024-10-29
Do not read __builtin_return_address when async stacks is disabled

jesswong:
wasm is broken when not using async stacks because there is no default return address. this diff fixes the issue by not reading the return address when async stacks is disabled


ispeters (approved):
LGTM

PR #642, opened 2024-11-07
Modify async stacks to keep track of a vector of ASRs

jesswong:
This diff updates async stack to keep track of the ASRs via a std::vector instead of a TLS key. This will allow us to keep track of the different ASRs for all threads, instead of just the crashing thread.

We also store the thread id to reduce ambiguity.

### DIFF
--- a/source/async_stack.cpp
+++ b/source/async_stack.cpp
@@ -20,16 +20,16 @@
 #include <cassert>
 #include <mutex>
 
-#if !defined(ASYNC_STACK_ROOT_USE_PTHREAD)
+#if !defined(UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD)
 #if defined(__linux__)
-#  define ASYNC_STACK_ROOT_USE_PTHREAD 1
+#  define UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD 1
 #else
 // defaults to using vector to store AsyncStackRoots instead of a pthread key
-#  define ASYNC_STACK_ROOT_USE_PTHREAD 0
+#  define UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD 0
 #endif
 #endif
 
-#if ASYNC_STACK_ROOT_USE_PTHREAD
+#if UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD
 
 #  include <pthread.h>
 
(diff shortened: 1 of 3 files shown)


source/async_stack.cpp
```diff
@@ -39,10 +42,47 @@ extern "C" {
 inline pthread_key_t folly_async_stack_root_tls_key = 0xFFFF'FFFFu;
 }
 
-#endif  // FOLLY_ASYNC_STACK_ROOT_USE_PTHREAD
+#else
+
+#include <vector>
+
+// TODO: expose array of AsyncStackRootHolder
```
jesswong, line 49: This is a todo because we need to access this out of process. In which case, we may need a tls key after all

source/async_stack.cpp
```diff
@@ -39,10 +42,47 @@ extern "C" {
...
+static std::unique_ptr<std::mutex> asyncStackRootsMutex = std::make_unique<std::mutex>();
+static std::vector<void*> asyncStackRoots;
+
+static void addAsyncStackRootHolder(void* holder) noexcept {
+  std::lock_guard<std::mutex> lock(*asyncStackRootsMutex);
+  asyncStackRoots.push_back(holder);
+}
+
+static void removeAsyncStackRootHolder(void* holder) noexcept {
+  if (asyncStackRootsMutex == nullptr) {
```
jesswong, line 65: The main thread that will create the vector and the mutex can end up destroying the vector and mutex, before the thread_local ASR is destroyed. In this case, we won't even have a vector of ASRs to fallback to.

An example of this happening is in https://github.com/facebookexperimental/libunifex/blob/main/examples/synchronous_sleep.cpp

source/async_stack.cpp
```diff
@@ -20,13 +20,16 @@
 #include <cassert>
 #include <mutex>
 
+#if !defined(ASYNC_STACK_ROOT_USE_PTHREAD)
```
ispeters, line 23: We should prefix our macros with `UNIFEX`.
```suggestion
#if !defined(UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD)
```

source/async_stack.cpp
```diff
@@ -39,10 +42,47 @@ extern "C" {
...
+
+// TODO: expose array of AsyncStackRootHolder
+// static std::vector<void*> getAsyncStackRoots() noexcept {
+//   if (!asyncStackRootsMutex.try_lock()) {
+//     // assume we crashed holding the lock and give up
+//     return {};
+//   }
+//   std::lock_guard<std::mutex> lock(asyncStackRootsMutex, std::adopt_lock);
+//   return asyncStackRoots;
+// }
```
ispeters, line 83: Maybe you should expose this as a type. Something like:

```cpp
struct AsyncStackRootHolderList {
  std::vector<void*> holders_;
  std::mutex mutex_;

  void add(void*) { … }

  void remove(void*) { … }
};

extern "C" auto* kUnifexAsyncStackRootHolderList = new AsyncStackRootHolderList();
```

If you leak it like that, the main thread won't destroy it. You can also find the start of the vector from out of process by looking up the symbol. Not sure how you make a copy of the vector from out of process…

source/async_stack.cpp
```diff
@@ -39,10 +42,47 @@ extern "C" {
...
+}
+
+static void removeAsyncStackRootHolder(void* holder) noexcept {
+  if (asyncStackRootsMutex == nullptr) {
+    return;
+  }
+  std::lock_guard<std::mutex> lock(*asyncStackRootsMutex);
+  auto it = std::find(asyncStackRoots.begin(), asyncStackRoots.end(), holder);
+  if (it != asyncStackRoots.end()) {
+    asyncStackRoots.erase(it);
```
ispeters, line 71: You can do this more efficiently than `.erase()` because you don't care about order.
```suggestion
    std::swap(*it, asyncStackRoots.back());
    asyncStackRoots.pop_back();
```

source/async_stack.cpp
```diff
@@ -38,11 +41,48 @@ extern "C" {
...
+std::vector<void*> getAsyncStackRoots() noexcept {
+  if (!asyncStackRootsMutex.try_lock()) {
+    // assume we crashed holding the lock and give up
+    return {};
+  }
+  std::lock_guard<std::mutex> lock(mutex_, std::adopt_lock);
+  return asyncStackRootHolders_;
+}
+
+extern "C" auto* kUnifexAsyncStackRootHolderList = new AsyncStackRootHolderList();
```
jesswong, line 81: don't we need to explicitly delete this to avoid leaks?

jesswong:
Latest changes were tested internally

source/async_stack.cpp
ispeters: We would, but leaking on purpose allows you to avoid worrying about background threads using this after the main thread has run after-main() global destructors.

source/async_stack.cpp
```diff
@@ -38,11 +41,46 @@ extern "C" {
...
+}
+
+std::vector<void*> getAsyncStackRoots() noexcept {
+  if (!mutex_.try_lock()) {
+    // assume we crashed holding the lock and give up
+    return {};
+  }
+  std::lock_guard<std::mutex> lock(mutex_, std::adopt_lock);
+  return asyncStackRootHolders_;
+}
```
ispeters, line 76: The formatting on this looks wonky; can you run this file through `clang-format`?
```suggestion
  void add(void* holder) noexcept {
    std::lock_guard<std::mutex> lock(mutex_);
    asyncStackRootHolders_.push_back(holder);
  }

  void remove(void* holder) noexcept {
    std::lock_guard<std::mutex> lock(mutex_);
    auto it = std::find(asyncStackRootHolders_.begin(), asyncStackRootHolders_.end(), holder);
    if (it != asyncStackRootHolders_.end()) {
      std::swap(*it, asyncStackRootHolders_.back());
      asyncStackRootHolders_.pop_back();
    }
  }

  std::vector<void*> getAsyncStackRoots() noexcept {
    if (!mutex_.try_lock()) {
      // assume we crashed holding the lock and give up
      return {};
    }
    std::lock_guard<std::mutex> lock(mutex_, std::adopt_lock);
    return asyncStackRootHolders_;
  }
```

source/async_stack.cpp
```diff
@@ -38,11 +41,46 @@ extern "C" {
...
+void remove(void* holder) noexcept {
+  std::lock_guard<std::mutex> lock(mutex_);
+  auto it = std::find(asyncStackRootHolders_.begin(), asyncStackRootHolders_.end(), holder);
+  if (it != asyncStackRootHolders_.end()) {
+    std::swap(*it, asyncStackRootHolders_.back());
+    asyncStackRootHolders_.pop_back();
+  }
+}
+
+std::vector<void*> getAsyncStackRoots() noexcept {
```
ispeters, line 69: None of these are `noexcept`; locking a mutex is allowed to throw, and adding to or copying the vector will (probably) allocate, which could throw.
```suggestion
void add(void* holder) {
  std::lock_guard<std::mutex> lock(mutex_);
  asyncStackRootHolders_.push_back(holder);
}

void remove(void* holder) {
  std::lock_guard<std::mutex> lock(mutex_);
  auto it = std::find(asyncStackRootHolders_.begin(), asyncStackRootHolders_.end(), holder);
  if (it != asyncStackRootHolders_.end()) {
    std::swap(*it, asyncStackRootHolders_.back());
    asyncStackRootHolders_.pop_back();
  }
}

std::vector<void*> getAsyncStackRoots() {
```

source/async_stack.cpp
```diff
@@ -38,11 +41,46 @@ extern "C" {
...
-#endif  // FOLLY_ASYNC_STACK_ROOT_USE_PTHREAD
+#endif  //UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD
 
 namespace unifex {
 
+#if UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD == 0
+#include <vector>
+
+struct AsyncStackRootHolderList {
+  std::vector<void*> asyncStackRootHolders_;
```
rileyberton, line 52: Might want to have a reserve here of some reasonable number to avoid allocations when pushing new roots.

source/async_stack.cpp
```diff
@@ -38,11 +41,52 @@ extern "C" {
...
 
+#if UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD == 0
+
+struct AsyncStackRootHolderList {
+  std::vector<void*> asyncStackRootHolders_;
+  std::mutex mutex_;
+
+  AsyncStackRootHolderList() {
+    // reserve some space to avoid reallocations
+    asyncStackRootHolders_.reserve(300);
```
ispeters, line 58: This seems like a lot; the vector should only have O(#threads) elements in it…

ispeters (approved):
LGTM!

source/async_stack.cpp
jesswong: Yeah it's definitely an overestimate. From some quick searching, I got an estimate that some apps peak between 200-500 threads

rileyberton: I think it's reasonable to set this to something like 16 or 32.  At 32 this will eliminate the first 6 or so allocations for resizing the vector, to get to 500 would only require 4 more allocations and I think this is OK.

PR #643, opened 2024-11-26
Remove .gitmodules and doc/std

jesswong:
This diff removes .gitmodules since libunifex is no longer associated with wg21 and does not require the wg21 submodule or the draft papers. This will make syncing libunifex easier


ispeters (approved):
Are there any references to the now-deleted submodule that need cleaning up, too? Maybe a Makefile that generates papers from source, or something? If so, those should go, too, but otherwise this looks good.

jesswong:
> Are there any references to the now-deleted submodule that need cleaning up, too? Maybe a Makefile that generates papers from source, or something? If so, those should go, too, but otherwise this looks good.

In doc/std there was a makefile that references the submodule. I just removed the whole dir since it also included some other draft papers that Eric and Lewis were working on.

PR #644, opened 2024-11-27
Add threadid to AsyncStackRootHolder

jesswong:
This diff adds the threadid to the ASR in order to unambiguously correlate the ASR's symbol to the thread.

This is loosely based off folly's implementation of getOSThreadID: https://github.com/facebook/folly/blob/main/folly/system/ThreadId.cpp#L41

The main difference is that folly supports EMSCRIPTEN and BSD, which I do not include.

We expect to only need this for ios, android, and linux applications.

### DIFF
--- a/source/async_stack.cpp
+++ b/source/async_stack.cpp
@@ -91,9 +91,7 @@ struct AsyncStackRootHolderList {
 extern "C" {
 auto* kUnifexAsyncStackRootHolderList = new AsyncStackRootHolderList();
 }
-#endif  // UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD == 0
 
-#if !UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD
 namespace utils {
 static std::uint64_t get_os_thread_id() {
 #  if defined(__APPLE__)
@@ -107,7 +105,7 @@ static std::uint64_t get_os_thread_id() {
 #  endif
 }
 }  // namespace utils
-#endif  // !UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD
+#endif  // UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD == 0
 
 namespace {
 
@@ -146,6 +144,9 @@ struct AsyncStackRootHolder {
   ~AsyncStackRootHolder() noexcept {
     kUnifexAsyncStackRootHolderList->remove(this);
   }
+
+  std::uint64_t get_thread_id() const noexcept { return threadId; }
+
 #endif
 
   AsyncStackRoot* get() const noexcept {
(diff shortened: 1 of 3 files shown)


rileyberton (approved):
LGTM

rileyberton (commented):
Looks like the windows build failed in CI

source/async_stack.cpp
```diff
@@ -145,6 +149,20 @@ static thread_local AsyncStackRootHolder currentThreadAsyncStackRoot;
 
 }  // namespace
 
+namespace utils {
+  uint64_t getOSThreadID() {
+   #if defined(__APPLE__)
+     uint64_t tid;
+     pthread_threadid_np(nullptr, &tid);
+     return tid;
```
ispeters, line 157: I think this Apple-specific implementation implies there's no need for an Apple-specific definition of `UNIFEX_SYS_gettid`, right?

source/async_stack.cpp
```diff
@@ -145,6 +149,20 @@ static thread_local AsyncStackRootHolder currentThreadAsyncStackRoot;
...
+   #elif defined(_WIN32)
+     return uint64_t(GetCurrentThreadId());
+   #else
+    return uint64_t(syscall(UNIFEX_SYS_gettid));
```
ispeters, line 161: Can we avoid defining `UNIFEX_SYS_gettid` by just switching here on whether `SYS_gettid` is defined?

Also, I'm a bit anxious about the cost of a syscall in each `AsyncStackRoot` constructor since we construct so many of them. Is this a cheap syscall implemented in user space? If so then it doesn't matter, but if it's a typical syscall that transits into kernel space then we should probably revisit capturing this value in the `AsyncStackRootHolder` since we'll have one of those per thread and the per-thread value of this is constant.

ispeters (changes requested):
Mostly nits but there's one possibly non-nit question about whether the syscall in the Linux implementation is expensive.

source/async_stack.cpp
jesswong: So I updated this to use the glibc wrapper of gettid because it implements caching ([D49174906](https://www.internalfb.com/diff/D49174906))

I also spoke to David and his response is that it's difficult to say because it depends on the implementation and we need to profile the code in order to know. This will be used a lot on android and his response is below:

> I see. The overhead of the syscall is mostly dependent on what speculative execution mitigations are turned on e.g. is the kernel hardened. I don't know how we configure our Android kernels.

The more mitigations the costlier the overhead of a user to kernel transition

The only real way to know is to profile the code. But as a hunch thread creation should be off the hot path so I don't imagine adding a gettid in there will be problematic

jesswong: Synced with @ispeters offline and decided to move threadid to async stack root holder for getting the threadid per thread only

source/async_stack.cpp
```diff
@@ -58,34 +63,49 @@ struct AsyncStackRootHolderList {
...
+  pthread_threadid_np(nullptr, &tid);
+  return tid;
+#  elif defined(_WIN32)
+  return std::uint64_t(GetCurrentThreadId());
+#  else
+  return std::uint64_t(gettid());
+#  endif
 }
-#endif // UNIFEX_ASYNC_STACK_ROOT_USE_PTHREAD == 0
+}  // namespace utils
```
ispeters, line 107: Since `get_os_thread_id` is static and inside the `unifex` namespace, I don't think you don't need the `utils` namespace.

ispeters (approved):
LGTM

PR #646, opened 2024-12-05
Add threadid macros for emscripten

jesswong:
It turns out there's some usage of unifex in emscripten. We don't plan on adding support for this work on emscripten so just return 0 is fine

### DIFF
--- a/source/async_stack.cpp
+++ b/source/async_stack.cpp
@@ -99,13 +99,12 @@ static std::uint64_t get_os_thread_id() {
   return tid;
 #  elif defined(_WIN32)
   return std::uint64_t(GetCurrentThreadId());
-#  elif defined(__FreeBSD__)
-  long tid;
-  thr_self(&tid);
 #  elif defined(__EMSCRIPTEN__)
   return 0;
-#  else
+#  elif defined(__linux__)
   return std::uint64_t(gettid());
+#  else
+#    error "Unsupported platform in get_os_thread_id"
 #  endif
 }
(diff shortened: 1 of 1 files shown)


source/async_stack.cpp
```diff
@@ -99,6 +99,11 @@ static std::uint64_t get_os_thread_id() {
   return tid;
 #  elif defined(_WIN32)
   return std::uint64_t(GetCurrentThreadId());
+#  elif defined(__FreeBSD__)
+  long tid;
+  thr_self(&tid);
```
ispeters, line 104: Can't decide if it's worth adding BSD support here; you're not returning `tid` so this code is broken but we won't catch that since we don't have any known usage on BSD. Probably best to delete this.
```suggestion
```

source/async_stack.cpp
```diff
@@ -99,6 +99,11 @@ static std::uint64_t get_os_thread_id() {
...
+#  elif defined(__EMSCRIPTEN__)
+  return 0;
 #  else
   return std::uint64_t(gettid());
```
ispeters, line 108: Do you know whether you could express this `else` as an `elsif` that specifically targets the expected platform so the `else` branch could be an error?
```suggestion
#  elif defined(__linux__) // or whatever is actually correct
  return std::uint64_t(gettid());
#  else
#    error Unsupported platform in get_os_thread_id
```

ispeters (approved):
You should probably just delete the broken BSD branch, but LGTM otherwise.

PR #647, opened 2024-12-05
Add missing licensing headers

jesswong:
Unifex is not currently adhering to Meta's open source standards. This diff makes unifex compliant by adding the missing license headers

### DIFF
--- a/examples/asynchronous_sleep.cpp
+++ b/examples/asynchronous_sleep.cpp
@@ -1,4 +1,18 @@
+/*
+Copyright (c) Meta Platforms, Inc. and affiliates.
 
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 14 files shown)


examples/asynchronous_sleep.cpp
```diff
@@ -1,4 +1,18 @@
+/*
+Copyright (c) Meta Platforms, Inc. and affiliates.
```
ispeters, line 2: The comment blocks in this diff don't match the existing ones (e.g. [in `then.hpp`](https://github.com/facebookexperimental/libunifex/blob/main/include/unifex/then.hpp).

ispeters (approved):
Looks good, but it'd be nice to match the formatting of the existing licensing blocks.

PR #650, opened 2025-01-06
Use intrusive_heap in timed_single_thread_context

### LINKED ISSUE #648: unifex task/timed_single_thread_context cannot run 1 million concurrent tasks
https://github.com/Xavorim/cpp_tasks_bench?tab=readme-ov-file#results-on-my-system benchmarks libunifex against a few other concurrency libraries. The benchmark attempts to launch 1 million concurrent tasks. Unfortunately, the benchmark states libunifex hangs at just 100K tasks. I am able to replicate this myself, with the benchmark program at 100% for a minute until I killed the program.

I reimplemented the benchmark with `async_manual_reset_event` instead of `timed_single_thread_context`, and I am able to launch 1 million tasks in about 4 seconds: https://github.com/Xavorim/cpp_tasks_bench/pull/1

I'm entering this issue to see if the timer scheduler can be improved. I'll also compare against stdexec (as the benchmark repo doesn't include it currently).

ccotter:
Use a heap to avoid possibly linear enqueue operation time.

Fixes #648

Note: this doesn't quite accomplish what I wanted, since the `intrusive_heap` is not a real heap (yet).

### DIFF
--- a/include/unifex/detail/intrusive_heap.hpp
+++ b/include/unifex/detail/intrusive_heap.hpp
@@ -89,7 +89,12 @@ public:
     }
   }
 
-  void remove(T* item) noexcept {
+  // Returns true if the item was in the heap
+  bool remove(T* item) noexcept {
+    if (item->*Prev == nullptr && item != head_) {
+        return false;
+    }
+
     auto* prev = item->*Prev;
     auto* next = item->*Next;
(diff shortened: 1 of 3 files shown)


ispeters:
Thanks for the PR! Looking forward to approving it. :)

PR #665, opened 2025-11-28
New primitives replacing/extending create()

MaxMotovilov:
A higher level, batteries-included framework for building senders that would allow relative newbies to produce production quality implementations without descending into the minutiae of synchronization, lifetime management and unifex internals. It also aims to significantly reduce the size of boilerplate compared to writing a custom unifex sender from scratch.


MaxMotovilov:
Reopening with squashed commits (after push --force)

PR #670, opened 2026-01-08
[fix] potential deadlocks during cancellation of async_pass senders

MaxMotovilov:
Stop callbacks should not be destroyed while under lock - there is a spinlock in the implementation of `remove_callback()` that may never end if normal completion and cancellation request race.

### DIFF
--- a/include/unifex/async_pass.hpp
+++ b/include/unifex/async_pass.hpp
@@ -45,8 +45,11 @@ inline constexpr bool nothrow_constructible_v =
 struct accept_op_base_noargs {
   void rethrow(std::exception_ptr ex) { (*set_error_)(this, std::move(ex)); }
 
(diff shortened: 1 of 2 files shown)


lewissbaker:
The internal spin lock of a stop callback should never participate in a deadlock as the lock is never held while calling into user code that might try to lock something else.

Attempts to destroy a stop callback from within the callback invocation should similarly not block.

The main case that a stop callback might participate in a deadlock is if you hold a lock while destroying the stop callback and the callback registered by the stop callback also tries to unconditionally acquire the lock. In this case a background thread could request stop and invoke the callback in between the first thread acquiring the lock and destroying the stop callback, causing the first thread to block on completion of the callback invocation which is blocked on acquiring the lock.

PR #694, opened 2026-05-22
io_uring_context with non-TU local c++23 htons implementation

calgray:
The single low lying issue with exporting `io_uring_context` to module is the use of the C API function `htons` which results in module compilation error:

```
error: ‘void unifex::linuxos::io_uring_context::accept_stream::open_socket()’ exposes TU-local entity ‘__uint16_t __bswap_16(__uint16_t)’
 1371 |   void open_socket() noexcept {
      |        ^~~~~~~~~~~
In file included from /usr/include/endian.h:35,
                 from /usr/include/sys/types.h:176,
                 from /usr/include/stdlib.h:518,
                 from /usr/include/c++/15.2.1/cstdlib:83,
                 from /usr/include/c++/15.2.1/ext/string_conversions.h:45,
                 from /usr/include/c++/15.2.1/bits/basic_string.h:4444,
                 from /usr/include/c++/15.2.1/string:56,
                 from /usr/include/c++/15.2.1/bits/locale_classes.h:42,
                 from /usr/include/c++/15.2.1/bits/ios_base.h:43,
                 from /usr/include/c++/15.2.1/ios:46,
                 from /usr/include/c++/15.2.1/bits/ostream.h:43,
                 from /usr/include/c++/15.2.1/ostream:42,
                 from /usr/include/c++/15.2.1/iostream:43,
                 from /code/modules/unifex/src/unifex.cppm:3:
/usr/include/bits/byteswap.h:34:1: note: ‘__uint16_t __bswap_16(__uint16_t)’ declared with internal linkage
   34 | __bswap_16 (__uint16_t __bsx)
      | ^~~~~~~~~~
ninja: build stopped: subcommand failed.
```

On linux this is a static function as follows:

```c
// /usr/include/bits/byteswap.h
static __inline __uint16_t
__bswap_16 (__uint16_t __bsx)
{
#if __GNUC_PREREQ (4, 8)
  return __builtin_bswap16 (__bsx);
#else
  return __bswap_constant_16 (__bsx);
#endif
}
```

static functions (excluding static methods) are not exportable through modules, however c++23 introduces new module-friendly byteswapping via `std::byteswap`. 

tested with gcc 15 using:

```cpp
export namspace unifex {
    using ::unifex::linuxos::io_uring_context;
}
```

### DIFF
--- a/.github/workflows/libunifex-ci.yml
+++ b/.github/workflows/libunifex-ci.yml
@@ -3,15 +3,13 @@ name: libunifex CI
 # Trigger on pushes to all branches and for all pull-requests
 on: [push, pull_request]
 
-env:
-  CMAKE_VERSION: 3.16.2
-  NINJA_VERSION: 1.9.0
-
-
 jobs:
   build:
     name: ${{ matrix.config.name }}
     runs-on: ${{ matrix.config.os }}
+    env:
+      CMAKE_VERSION: ${{ matrix.config.cmake_version || '3.16.2' }}
+      NINJA_VERSION: ${{ matrix.config.ninja_version || '1.9.0' }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


MaxMotovilov:
Try checking for [https://en.cppreference.com/cpp/feature_test#cpp_lib_byteswap](`__cpp_lib_byteswap`) - although I have seen inconsistent definitions of feature macros in the past. Worst case, you can just check for C++23.

Note that you would need to add C++23 CI configuration(s) as well since presently we only test against C++17 (now with some exclusions) and C++20. In fact, I would invite you to submit a PR adding C++23 CI runs prior to this one because, obviously, we shouldn't be merging code with no CI coverage.

calgray:
Some notes from adding C++23 CI tests:
* `CMAKE_CXX_STANDARD=23` support was added in cmake 3.20.0. Rather than update the minimum cmake of all pipelines, I've allowed the job matrix to override the cmake and ninja versions
* CMake 3.20.0 changed the casing of the linux [installation artifact file name](https://github.com/Kitware/CMake/releases/tag/v3.20.0), hence some branching logic there
* Clang 22 with C++23 linux job was attempted, but failed due to google test containing a `char8_t` to `char32_t` conversion that clang picks up as warning and build error (see https://github.com/calgray/libunifex/actions/runs/26685996238/job/78654203848)

MaxMotovilov:
> * Clang 22 with C++23 linux job was attempted, but failed due to google test containing a `char8_t` to `char32_t` conversion that clang picks up as warning and build error (see https://github.com/calgray/libunifex/actions/runs/26685996238/job/78654203848)

Could be https://github.com/google/googletest/issues/4762? We can wait till they release a fix then try adding clang. Thank you for starting the process anyway, it's 2026 already - well past time to start maintaining C++23 compliance!

